PR #6, opened 2020-01-07
Added genesis functions

sunnya97:

- [ ] Linked to github-issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Wrote tests

______

For Admin Use:
- Reviewers Assigned


x/wasm/alias.go
```diff
@@ -54,6 +55,11 @@ var (
 	NewContract               = types.NewContract
 	CosmosResult              = types.CosmosResult
 
+	// genesis aliases
+	ValidateGenesis = types.ValidateGenesis
```
ethanfrey: Is this the new style to place commands?

I feel "best practices" on the cosmos sdk changes every couple months and I cannot keep up

x/wasm/internal/types/genesis.go
```diff
@@ -0,0 +1,29 @@
+package types
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// GenesisState is the struct representation of the export genesis
+type GenesisState struct {
```
ethanfrey: This in general seems very hard to validate. The fact that the various components for any contract are scattered. Using the same information, how about:

```go
type GenesisState []ContractState
// (or type GenesisState map[string]ContractState if that makes more sense)

type ContractState struct {
        CodeInfo         CodeInfo       `json:"code_info"`
	CodesBytes        []byte         `json:"code_bytes"`
	ContractAddress sdk.AccAddress `json:"contract_address"`
	ContractInfo     Contract       `json:"contract_info"`
	ContractState    []Model        `json:"contract_state"`
}
```

ethanfrey: I thought about this before and have some issues with such a huge code-dump in genesis. For example, the CodeBytes is 60-100kb, maybe this could live in a binary file next to the genesis file, and just keep a reference here. Then again, maybe that violates the idea that the genesis file is everything you need. I just see that a few contracts could easily add a MB to the genesis size. And if there are 100s, this will make the genesis file almost unusable.

x/wasm/internal/types/types.go
```diff
@@ -6,6 +6,12 @@ import (
 	auth "github.com/cosmos/cosmos-sdk/x/auth/exported"
 )
 
+// Model is a struct that holds a KV pair
+type Model struct {
```
ethanfrey: This is how I defined it. I will make them both `[]byte` in the next upgrade (very soon), as there was strong desire for binary keys internally by the first contract dev.

x/wasm/internal/keeper/genesis.go
```diff
@@ -0,0 +1,74 @@
...
+//
+// CONTRACT: all types of accounts must have been already initialized/created
+func InitGenesis(ctx sdk.Context, keeper Keeper, data types.GenesisState) {
+	for id, info := range data.CodeInfos {
+		bytecode := data.CodesBytes[id]
+		newId, err := keeper.Create(ctx, info.Creator, bytecode)
+		if err != nil {
+			panic(err)
+		}
+		newInfo := keeper.GetCodeInfo(ctx, newId)
```
ethanfrey: What is this check for?

x/wasm/internal/keeper/genesis.go
```diff
@@ -0,0 +1,74 @@
...
+		if !info.Creator.Equals(newInfo.Creator) {
+			panic("code creators not same")
+		}
+		if bytes.Compare(info.CodeHash, newInfo.CodeHash) != 0 {
```
ethanfrey: This definitely is super important to validate :+1:

I would just recommend `! bytes.Equal(.., ..)` instead for readability

x/wasm/internal/keeper/keeper.go
```diff
@@ -328,6 +341,16 @@ func (k Keeper) generateContractAddress(ctx sdk.Context, codeID uint64) sdk.AccA
 	return addrFromUint64(contractID)
 }
 
+func (k Keeper) GetLastID(ctx sdk.Context, lastIDKey []byte) uint64 {
```
ethanfrey: I would remove the second arg and just hardcode it to `types.KeyLastCodeID`, as that is all it is used for.

Also, this should be called `GetNextID`, as there is no contract yet for this ID (I believe)

x/wasm/internal/keeper/genesis.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+}
+
+// ExportGenesis returns a GenesisState for a given context and keeper.
+func ExportGenesis(ctx sdk.Context, keeper Keeper) types.GenesisState {
+	var genState types.GenesisState
+
+	maxCodeID := keeper.GetLastID(ctx, types.KeyLastCodeID)
+	for i := uint64(0); i < maxCodeID; i++ {
```
ethanfrey: This should start at 1

x/wasm/internal/types/genesis.go
ethanfrey: After reading through the pr more, I realize there are two separate groups - Code and Contract - which need their own indexes. But I would still group all info on one object together. Maybe something like:

```go
type GenesisState struct {
  Code []CodeState
  Contracts []ContractState
}

type CodeState struct {
        CodeInfo         CodeInfo       `json:"code_info"`
	CodesBytes        []byte         `json:"code_bytes"`
}

type ContractState struct {
	ContractAddress sdk.AccAddress `json:"contract_address"`
	ContractInfo     Contract       `json:"contract_info"`
	ContractState    []Model        `json:"contract_state"`
}
```

x/wasm/alias.go
sunnya97: No, I just wanted `InitGenesis` to be in keeper folder, so it can use unexposed functions `setContractInfo` and `setContractInfo`.  Is there a better way to access them from the outer package without exposing those functions?

ethanfrey: Good reason. And I am happy to have those functions unexported

x/wasm/internal/keeper/genesis.go
sunnya97: Just a sanity check to make sure it registered the Code as having the same creator it did last time.

x/wasm/internal/keeper/keeper.go
sunnya97: Was trying to keep API similarity to `autoIncrementID`.  I feel `autoIncrementID` should be split into two specialized functions that only take in `ctx`.  Thoughts?

x/wasm/internal/keeper/genesis.go
sunnya97: Is there a reason for not 0 indexing this?

sunnya97:
Do you mind if I rename the current `Contract` struct to `ContractInfo`?  That way it's more in line with `CodeInfo`.

sunnya97:
> The cosmoshub2 genesis export was 3.2 MB https://github.com/cosmos/launch/blob/master/genesis.cosmoshub-2.json and I remember hearing that cause problems for a few people to initialize from

If I remember correctly, I think someone thought that it was causing issues for them, but it wasn't, and their issue was something else.  A couple megabytes in memory to read a JSON should be fine.

But that being said, you're right, as the state increases into the gigabytes, to the level of the number of contracts something like Ethereum has, then it's definitely going to have major issues.

I agree it is nice to have all the genesis info in one file, but I guess in the future it's okay if that is a `.zip` instead of a `.json` per say.  I think in that case, we should have a mechanism where the genesis json refers to a relative file directory, and includes a hash of the file (in this case, we happen to have the code hash already, but for other use cases it might not).   I think should be figured out as a common standard across SDK projects.

For now, I think keeping it all in genesis json is fine, but we should start thinking about an ADR for doing this generally in the SDK.

x/wasm/internal/types/types.go
sunnya97: I agree, binary keys sounds right to me.  It can be casted to string as needed.

x/wasm/internal/keeper/keeper.go
ethanfrey: Yeah, that sounds good. I saw your fix and I like it

x/wasm/internal/keeper/genesis.go
ethanfrey: Honestly, I don't remember. I discussed this with @shanev and he implemented this, at least the start of the wasmd project (originally in cosmwasm/modules, but then I moved it over). Maybe I requested this, or maybe he did. I think it could be either way, or start at 50 for all I care. But it should be well documented, so we are consistent. Maybe we make a constant `firstCodeID`, so this is obvious?

ethanfrey:
> Do you mind if I rename the current Contract struct to ContractInfo? That way it's more in line with CodeInfo.

Please do. The code works, but could definitely use some stylistic cleanup. And if you find any bugs, please report/fix them.

ethanfrey:
> For now, I think keeping it all in genesis json is fine, but we should start thinking about an ADR for doing this generally in the SDK.

This is a good point. Let's go the most straight-forward approach now. There are some months til we need to have this scalability, and best if it works with some system supported by the sdk rather than bikeshedding something else.

x/wasm/internal/keeper/genesis.go
sunnya97: Hmm, the starting from `1` assumption is hard-coded into a bunch of tests.  Opened as #10 to constrain scope of this PR

sunnya97:
I think all comments have been addressed 👍

sunnya97:
I think all comments have been addressed 👍

x/wasm/internal/types/genesis.go
```diff
@@ -0,0 +1,13 @@
+package types
+
+// GenesisState is the struct representation of the export genesis
+type GenesisState struct {
+	Codes     []Code     `json:"codes"`
```
ethanfrey: Code and Contract are only used by GenesisState, correct?
I would move them here, it makes it easier to understand the code

ethanfrey:
I'm happy to merge this, but would love a test case... at least a simple one on the happy path. Dump state and restart is a horrible time to find a bug

ethanfrey:
Okay, I will merge without tests, as other PRs are incoming and I do not want merge conflicts.

But I left an open issue to finish up tests https://github.com/cosmwasm/wasmd/issues/20

sunnya97:
Oops sorry, was writing the test, but got busy with something else for a few days.  Opened PR for tests in #23

PR #9, opened 2020-01-09
Uncompress gzip data in tx body

alpe:
Resolves https://github.com/confio/go-cosmwasm/issues/20

As discussed before, I have implemented the decompression into wasmd. I would argue that this is part of the framework layer as the `MsgStoreCode` is.

______

For Admin Use:
- Reviewers Assigned


x/wasm/internal/keeper/ioutil.go
```diff
@@ -0,0 +1,38 @@
...
+	"io"
+	"io/ioutil"
+)
+
+// magic bytes to identify gzip.
+// See https://www.ietf.org/rfc/rfc1952.txt
+// and https://github.com/golang/go/blob/master/src/net/http/sniff.go#L186
+var gzipIdent = []byte("\x1F\x8B\x08")
+
+// limit max bytes read to prevent gzip bombs
```
ethanfrey: Yes, very good

x/wasm/internal/keeper/ioutil.go
```diff
@@ -0,0 +1,38 @@
...
+const maxSize = 400 * 1024
+
+// uncompress returns gzip uncompressed content or given src when not gzip.
+func uncompress(src []byte) ([]byte, error) {
+	if len(src) < 3 {
+		return src, nil
+	}
+	in := io.LimitReader(bytes.NewReader(src), maxSize)
+	buf := make([]byte, 3)
```
ethanfrey: This seems like a good way to ensure exactly 3 bytes read, but I wonder if it is guaranteed never to read an extra byte and drop it somewhere. https://golang.org/pkg/io/#ReadAtLeast

But I really doubt odd behavior when the input is a buffer not some network stream

x/wasm/internal/keeper/ioutil.go
```diff
@@ -0,0 +1,38 @@
...
+	if _, err := io.ReadAtLeast(in, buf, 3); err != nil {
+		return nil, err
+	}
+	if !bytes.Equal(gzipIdent, buf) {
+		return src, nil
+	}
+	zr, err := gzip.NewReader(io.MultiReader(bytes.NewReader(buf), in))
```
ethanfrey: You limit the input size with a Limit Reader, I think we need to wrap `zr` with a limit as well.
A zip bomb would have a relatively small input size (`src`) but a huge output size `ioutil.ReadAll(zr)`.

x/wasm/internal/keeper/ioutil_test.go
```diff
@@ -0,0 +1,63 @@
...
+		},
+		"handle big slice": {
+			src:       []byte(strings.Repeat("a", maxSize+1)),
+			expResult: []byte(strings.Repeat("a", maxSize+1)),
+		},
+		"handle gzip identifier only": {
+			src:      gzipIdent,
+			expError: io.ErrUnexpectedEOF,
+		},
+		"handle broken gzip": {
```
ethanfrey: Please add a test or two for zip bombs (after reading your comment, I'm thinking a lot about them)

I found the following famous zip bomb (42.zip): https://www.unforgettable.dk/ but it uses many files (recursively), so this might be rejected anyway (multistream=false).

Here is an article on non-recursive zip bombs, along with some links. One claims  42 kB → 5.5 GB

https://www.bamsoftware.com/hacks/zipbomb/ ([original paper](https://www.usenix.org/system/files/woot19-paper_fifield_0.pdf))

I'm not sure if these work on gzip, but it would be good to find a proper copy to validate the defense.

x/wasm/internal/keeper/ioutil_test.go
```diff
@@ -0,0 +1,63 @@
...
+	specs := map[string]struct {
+		src       []byte
+		expError  error
+		expResult []byte
+	}{
+		"handle wasm uncompressed": {
+			src:       wasmRaw,
+			expResult: wasmRaw,
+		},
+		"handle wasm compressed": {
```
ethanfrey: Can you also add an invalid zip file?
I think `wasmGzipped[:len(wasmGzipped)-5]` should produce an error (as there is a checksum).

ethanfrey: Besides that (and the zip bomb), nice test coverage

x/wasm/internal/keeper/keeper.go
```diff
@@ -58,6 +58,10 @@ func NewKeeper(cdc *codec.Codec, storeKey sdk.StoreKey, accountKeeper auth.Accou
 
 // Create uploads and compiles a WASM contract, returning a short identifier for the contract
 func (k Keeper) Create(ctx sdk.Context, creator sdk.AccAddress, wasmCode []byte) (codeID uint64, sdkErr sdk.Error) {
+	wasmCode, err := uncompress(wasmCode)
```
ethanfrey: amazing to do this with 4 lines :+1:

x/wasm/internal/keeper/keeper_test.go
```diff
@@ -39,6 +39,33 @@ func TestCreate(t *testing.T) {
...
+	wasmCode, err := ioutil.ReadFile("./testdata/contract.wasm.gzip")
+	require.NoError(t, err)
+
+	contractID, err := keeper.Create(ctx, creator, wasmCode)
+	require.NoError(t, err)
+	require.Equal(t, uint64(1), contractID)
+	// and verify content
+	storedCode, err := keeper.GetByteCode(ctx, contractID)
+	require.NoError(t, err)
+	rawCode, err := ioutil.ReadFile("./testdata/contract.wasm")
```
ethanfrey: Great test, ensuring we store the same wasm properly

x/wasm/internal/types/msg.go
```diff
@@ -10,6 +10,7 @@ const (
 
 type MsgStoreCode struct {
 	Sender       sdk.AccAddress `json:"sender" yaml:"sender"`
+	// WASMByteCode can be raw or gzip compressed
```
ethanfrey: :smiley:

x/wasm/internal/keeper/ioutil.go
alpe: From the [Reader Godoc](https://golang.org/pkg/io/#Reader):
> Read reads up to len(p) bytes into p.

x/wasm/internal/keeper/ioutil.go
alpe: ja, should have been limit on `zr` instead. 🐛

x/wasm/internal/keeper/ioutil_test.go
alpe: Added your case. I could not make a gzip bomb work with the code so I have added a test that checks a regular file with content > max.
It works all nicely with the `LimitedReader`. The only problem is that the error message does not reflect the `ContractLimitExceeded` constraint but is `ErrUnexpectedEOF`.
Checking `LimitedReader.N <=0` did not work as `N` was not updated for some reasons

x/wasm/internal/keeper/ioutil.go
ethanfrey: Can you replace lines 21-30 with:

```go
if !bytes.Equal(gzipIdent, src[:3]) {
  return src, nil
}
zr, err := gzip.NewReader(bytes.NewReader(src))
```

Same functionality, just cleaner

alpe: ja, much cleaner as src is not a stream. Thanks!

x/wasm/internal/keeper/ioutil_test.go
```diff
@@ -0,0 +1,88 @@
...
+		},
+		"handle incomplete gzip": {
+			src:      wasmGzipped[:len(wasmGzipped)-5],
+			expError: io.ErrUnexpectedEOF,
+		},
+		"handle big gzip output": {
+			src:      asGzip(strings.Repeat("a", maxSize+1)),
+			expError: io.ErrUnexpectedEOF,
+		},
+		"handle other big gzip output": {
```
ethanfrey: Thank you, this error: `io.ErrUnexpectedEOF` is a "Too big error" and can be tested against

PR #22, opened 2020-01-13
Improve wasm contract queries

### LINKED ISSUE #12: Improve `query wasm contract-state` CLI syntax
## Summary

There is a comment on https://www.cosmwasm.com/docs/getting-started/first-demo that this command must be improved. Furthermore, we will soon have binary keys. Let's clean this up.

## Problem Definition

There is a cli command: `wasmcli query wasm contract-state $CONTRACT` that dumps the entire contract state. This will quickly become unusable for all but the most simple cases. We also need to make sure the queries work well for the rest API.

## Proposal

The following subcommands, each can be done as a separate PR:

- [ ] `wasmcli query wasm contract-state raw $CONTRACT $KEY` - this will make access the given key inside the contracts store and return the value verbatum. 
- [ ] `wasmcli query wasm contract-state smart $CONTRACT $QUERY` - this will make a call to the given contract via the `query` handler (to be added in 0.6), and passes in `$QUERY` as the user message 
- [ ] `wasmcli query wasm contract-state all $CONTRACT` - behaves as the current one, but we need to add pagination to it. It should output (key, value) pairs (the above only output values), and should encode the key in `hex` format in all cases. 

Encoding formats: 

At least with 0.6, we will be dealing with `[]byte` that may be ascii or may or may not be binary. Until there are major changes in `cosmwasm`, we can assume that the value is always valid utf-8 encoded json and just wrap this as `json.RawMessage`. 

The `key` can be anything. To clarify this, we will add three flags, of which only one can be set. If none is set, we assume `ascii` for `$QUERY` and `hex` for `$KEY`. Queries are generally assumed to be json (ascii), while almost all keys will have non-ascii characters (when using prefixstore in `cw-erc20` contract or `cw-storage`):

* `--ascii` flag that will force decoding `$KEY`/`$QUERY` as ascii, eg. just cast to `[]byte` 
* `--hex` flag that will force decoding `$KEY`/`$QUERY` as hex (with or without the 0x prefix)
* `--b64` flag that will force decoding `$KEY`/`$QUERY` as base64
____

#### For Admin Use

- [ ] Not duplicate issue

alpe:
Resolves #12

In this PR are the 3 different query types implemented as defined in #12. This includes server side and cli code.

Example:
```sh
wasmcli query wasm contract-state all "${CONTRACT}"
wasmcli query wasm contract-state raw  "${CONTRACT}" config --ascii
wasmcli query wasm contract-state smart "${CONTRACT}" '{"raw":{"key":"config"}}' --ascii
```
They all return:
```json
[
  {
    "key": "config",
    "val": "{\"arbiter\":\"cosmos1s5k2p7nct27jzjrzvqgwy785wug0js9vdmdr5s\",\"recipient\":\"cosmos1p35xmxjl8ysxdp699ff4kmca8duwcve2jupxe3\",\"source\":\"cosmos1laapuw0u4vx3jjyq8vtf3pmzj24h3eqdfunyuk\",\"end_height\":0,\"end_time\":0}"
  }
]
```
______

For Admin Use:
- Reviewers Assigned

### DIFF
--- a/x/wasm/client/cli/query.go
+++ b/x/wasm/client/cli/query.go
@@ -2,6 +2,7 @@ package cli
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"strconv"
@@ -145,10 +146,26 @@ func GetCmdGetContractInfo(cdc *codec.Codec) *cobra.Command {
 
 // GetCmdGetContractState dumps full internal state of a given contract
 func GetCmdGetContractState(cdc *codec.Codec) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:                        "contract-state",
+		Short:                      "Querying commands for the wasm module",
+		DisableFlagParsing:         true,
+		SuggestionsMinimumDistance: 2,
+		RunE:                       client.ValidateCmd,
+	}
+	cmd.AddCommand(client.GetCommands(
+		GetCmdGetContractStateAll(cdc),
+		GetCmdGetContractStateRaw(cdc),
+	)...)
+	return cmd
+
+}
+
+func GetCmdGetContractStateAll(cdc *codec.Codec) *cobra.Command {
 	return &cobra.Command{
-		Use:   "contract-state [bech32_address]",
-		Short: "Prints out internal state of a contract given its address",
-		Long:  "Prints out internal state of a contract given its address",
+		Use:   "all [bech32_address]",
+		Short: "Prints out all internal state of a contract given its address",
+		Long:  "Prints out all internal state of a contract given its address",
 		Args:  cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			cliCtx := context.NewCLIContext().WithCodec(cdc)
@@ -159,6 +176,33 @@ func GetCmdGetContractState(cdc *codec.Codec) *cobra.Command {
 			}
 
 			route := fmt.Sprintf("custom/%s/%s/%s", types.QuerierRoute, keeper.QueryGetContractState, addr.String())
+			res, _, err := cliCtx.Query(route)
+			if err != nil {
+				return err
+			}
+			fmt.Println(string(res))
+			return nil
+		},
+	}
+}
+func GetCmdGetContractStateRaw(cdc *codec.Codec) *cobra.Command {
+	return &cobra.Command{
+		Use:   "raw [bech32_address] [key]",
+		Short: "Prints out internal state for key of a contract given its address",
+		Long:  "Prints out internal state for of a contract given its address",
+		Args:  cobra.ExactArgs(2),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			cliCtx := context.NewCLIContext().WithCodec(cdc)
+
+			addr, err := sdk.AccAddressFromBech32(args[0])
+			if err != nil {
+				return err
+			}
+			key := args[1]
+			if key == "" {
+				return errors.New("key must not be empty")
+			}
+			route := fmt.Sprintf("custom/%s/%s/%s/%s", types.QuerierRoute, keeper.QueryGetContractState, addr.String(), key)
 			res, _, err := cliCtx.Query(route)
 			if err != nil {
 				return err
--- a/x/wasm/client/cli/query.go
+++ b/x/wasm/client/cli/query.go
@@ -1,11 +1,15 @@
 package cli
 
 import (
+	"encoding/base64"
+	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"io/ioutil"
 	"strconv"
+
+	flag "github.com/spf13/pflag"
 
 	"github.com/spf13/cobra"
 
@@ -176,7 +180,7 @@ func GetCmdGetContractStateAll(cdc *codec.Codec) *cobra.Command {
 				return err
 			}
 
-			route := fmt.Sprintf("custom/%s/%s/%s", types.QuerierRoute, keeper.QueryGetContractState, addr.String())
+			route := fmt.Sprintf("custom/%s/%s/%s/%s", types.QuerierRoute, keeper.QueryGetContractState, addr.String(), keeper.QueryMethodContractStateAll)
 			res, _, err := cliCtx.Query(route)
 			if err != nil {
 				return err
@@ -188,24 +192,24 @@ func GetCmdGetContractStateAll(cdc *codec.Codec) *cobra.Command {
 }
 
 func GetCmdGetContractStateRaw(cdc *codec.Codec) *cobra.Command {
-	return &cobra.Command{
+	decoder := newArgDecoder(hex.DecodeString)
+	cmd := &cobra.Command{
 		Use:   "raw [bech32_address] [key]",
 		Short: "Prints out internal state for key of a contract given its address",
 		Long:  "Prints out internal state for of a contract given its address",
 		Args:  cobra.ExactArgs(2),
-		RunE: func(cmd *cobra.Command, args []string) error {
+		RunE: func(_ *cobra.Command, args []string) error {
 			cliCtx := context.NewCLIContext().WithCodec(cdc)
 
 			addr, err := sdk.AccAddressFromBech32(args[0])
 			if err != nil {
 				return err
 			}
-			key := args[1]
-			if key == "" {
-				return errors.New("key must not be empty")
+			queryData, err := decoder.DecodeString(args[1])
+			if err != nil {
+				return err
 			}
 			route := fmt.Sprintf("custom/%s/%s/%s/%s", types.QuerierRoute, keeper.QueryGetContractState, addr.String(), keeper.QueryMethodContractStateRaw)
-			queryData := []byte(key) // todo: open question: encode into json???
 			res, _, err := cliCtx.QueryWithData(route, queryData)
 			if err != nil {
 				return err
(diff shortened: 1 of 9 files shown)


x/wasm/client/cli/query.go
```diff
@@ -168,3 +185,30 @@ func GetCmdGetContractState(cdc *codec.Codec) *cobra.Command {
...
+
+			addr, err := sdk.AccAddressFromBech32(args[0])
+			if err != nil {
+				return err
+			}
+			key := args[1]
+			if key == "" {
+				return errors.New("key must not be empty")
+			}
+			route := fmt.Sprintf("custom/%s/%s/%s/%s", types.QuerierRoute, keeper.QueryGetContractState, addr.String(), key)
```
ethanfrey: It would be good to include `all`/`raw`/`smart` here to make it easier to differentiate.
Adding one length gets tricky when we add a third.

ethanfrey: btw, smart should upload a json blob, which doesn't really fit in the path domain, you can use `cliCtx.QueryWithData` to include a data field (this request body in HTTP land), which is nicer place to dump json

x/wasm/internal/keeper/querier.go
```diff
@@ -76,13 +88,15 @@ func queryContractState(ctx sdk.Context, bech string, req abci.RequestQuery, kee
 
 	var state []types.Model
 	for ; iter.Valid(); iter.Next() {
+		if !accept(iter.Key()) {
```
ethanfrey: Iterate over all keys just to get one? Super inefficient.
The db supports `Get`. You just have to expose this in the `Keeper`.
Feel free to add/refactor functions there to make the querier cleaner

x/wasm/client/cli/query.go
alpe: > The key can be anything

I would also move this to the body. Can either be raw or wrapped into some
```json
{
"key": "the data"
}
```
element to make it more restful. WDYT?

alpe: For the smart query as well. Should it be json or length prefixed amino? I guess later one

ethanfrey: > element to make it more restful. WDYT?

Hmmm... I wasn't thinking of restful at all with the cli, but then yes, the Querier implementation will be used by both, so it makes sense for something that works well as a REST backend. Of course the rest-server can make any needed transform from the user input to the Querier interface, so the most important things are to:

1. Document the API and
2. Provide a helper function to construct the querier call

The internal api between rest/cli and QueryHandler in the abci app is not so important (if helper function is exposed), but yeah, should not be too ugly. Use you judgement there

alpe: I would start with raw bytes. A REST server impl would also need to take the result types into account. I noticed an empty result is currently `nil` not `[]` for example.

x/wasm/client/cli/query.go
```diff
@@ -168,3 +186,62 @@ func GetCmdGetContractState(cdc *codec.Codec) *cobra.Command {
...
+			addr, err := sdk.AccAddressFromBech32(args[0])
+			if err != nil {
+				return err
+			}
+			key := args[1]
+			if key == "" {
+				return errors.New("key must not be empty")
+			}
+			route := fmt.Sprintf("custom/%s/%s/%s/%s", types.QuerierRoute, keeper.QueryGetContractState, addr.String(), keeper.QueryMethodContractStateRaw)
+			queryData := []byte(key) // todo: open question: encode into json???
```
ethanfrey: I think the issue mentions some flags to control how key is parsed (ascii, hex, base64), let's do that.

No need to encode into json, I think the raw bytes is good (parse the human cli into binary)

x/wasm/client/cli/query.go
```diff
@@ -168,3 +186,62 @@ func GetCmdGetContractState(cdc *codec.Codec) *cobra.Command {
...
+			addr, err := sdk.AccAddressFromBech32(args[0])
+			if err != nil {
+				return err
+			}
+			key := args[1]
+			if key == "" {
+				return errors.New("key must not be empty")
+			}
+			route := fmt.Sprintf("custom/%s/%s/%s/%s", types.QuerierRoute, keeper.QueryGetContractState, addr.String(), keeper.QueryMethodContractStateSmart)
+			var queryData []byte
```
ethanfrey: I guess you didn't finish this. Here, I would just set `queryData := []byte(key)` as we assume verbatim json.

However, you can also support flags as above you wish, and just default to ascii here

x/wasm/internal/keeper/keeper.go
```diff
@@ -128,22 +133,10 @@ func (k Keeper) Instantiate(ctx sdk.Context, creator sdk.AccAddress, codeID uint
...
-	if contractBz == nil {
-		return sdk.Result{}, types.ErrNotFound("contract")
-	}
-	var contract types.ContractInfo
-	k.cdc.MustUnmarshalBinaryBare(contractBz, &contract)
-
-	contractInfoBz := store.Get(types.GetCodeKey(contract.CodeID))
-	if contractInfoBz == nil {
-		return sdk.Result{}, types.ErrNotFound("contract info")
+	codeInfo, prefixStore, err := k.contractInstance(ctx, contractAddress)
```
ethanfrey: Nice refactor :+1:

x/wasm/internal/keeper/keeper.go
```diff
@@ -170,6 +160,42 @@ func (k Keeper) Execute(ctx sdk.Context, contractAddress sdk.AccAddress, caller
 	return types.CosmosResult(*res), nil
 }
 
+func (k Keeper) Query(ctx sdk.Context, contractAddr sdk.AccAddress, req []byte) (*wasmTypes.QueryResult, sdk.Error) {
```
ethanfrey: This only does smart queries, right?

I think it would be good to name this QuerySmart or such, and make a QueryRaw that handles the raw queries, even if it is just a few lines: https://github.com/cosmwasm/wasmd/pull/22/files#diff-4abe0cb83913fb6687ff80141ae33310R97-R101

We will expose these to be called from the smart contracts themselves in 0.7, so having them on the keeper is very useful

x/wasm/internal/keeper/keeper.go
```diff
@@ -170,6 +160,42 @@ func (k Keeper) Execute(ctx sdk.Context, contractAddress sdk.AccAddress, caller
...
+	}
+	res, gasUsed, qErr := k.wasmer.Query(codeInfo.CodeHash, req, prefixStore, gasForContract(ctx))
+	if qErr != nil {
+		return nil, types.ErrExecuteFailed(qErr)
+	}
+	consumeGas(ctx, gasUsed)
+	return res, nil
+}
+
+func (k Keeper) contractInstance(ctx sdk.Context, contractAddress sdk.AccAddress) (types.CodeInfo, prefix.Store, sdk.Error) {
```
ethanfrey: Yes.

x/wasm/internal/keeper/keeper.go
alpe: I will return `wasmTypes.QueryResult` in this case from the method. Please let me know if this is a bad idea. :-)

ethanfrey: In 0.6 go-cosmwasm, it just returns `[]byte`.
So do the simplest thing possible, and I will fix when I merge in 0.6 support

x/wasm/client/cli/query.go
alpe: cli updates are in the next commit. This was wip.

x/wasm/internal/keeper/querier_test.go
```diff
@@ -0,0 +1,131 @@
...
+		},
+		"query raw with unknown address": {
+			srcPath:     []string{QueryGetContractState, anyAddr.String(), QueryMethodContractStateRaw},
+			expModelLen: 0,
+		},
+		"query all with unknown address": {
+			srcPath:     []string{QueryGetContractState, anyAddr.String(), QueryMethodContractStateAll},
+			expModelLen: 0,
+		},
+		"query smart with unknown address": {
```
ethanfrey: I would assume all the "query with unknown address" would return the same `types.ErrNotFound("contract")` message

x/wasm/internal/keeper/querier_test.go
```diff
@@ -0,0 +1,131 @@
...
+			srcPath:     []string{QueryGetContractState, anyAddr.String(), QueryMethodContractStateSmart},
+			expModelLen: 0,
+			expErr:      types.ErrNotFound("contract"),
+		},
+	}
+
+	for msg, spec := range specs {
```
ethanfrey: Very nice use of table tests

x/wasm/internal/keeper/querier_test.go
```diff
@@ -0,0 +1,131 @@
...
+			expModelLen:      1,
+			expModelContains: []model{{Key: string([]byte{0x0, 0x1}), Value: string([]byte{0x2, 0x3})}},
+		},
+		"query smart": {
+			srcPath:     []string{QueryGetContractState, addr.String(), QueryMethodContractStateSmart},
+			srcReq:      abci.RequestQuery{Data: []byte(`{"raw":{"key":"config"}}`)},
+			expModelLen: 1,
+			//expModelContains: []model{}, // stopping here as contract internals are not stable
+		},
+		"query unknown raw key": {
```
ethanfrey: Why doesn't this return an empty slice?

alpe: This was the simplest implementation but not necessarily the best. I was under the assumption that the KV store allows nil values so this result make sense but it does not.

ethanfrey: Maybe you are right, but I would rather let len(0) = deleted. Even if not strictly true, it gives contracts a way to mark things as deleted

x/wasm/internal/keeper/querier_test.go
alpe: Agree `ErrNotFound ` would be nicer but the `prefix.Store` concatenates the prefix and key for the parent store so that it results in just another unknown key.
I would need to do a prefix scan on the parent to detect the issue. The previous implementation for `all` had the same "problem"

x/wasm/internal/keeper/querier_test.go
alpe: I added a patch. And also returning non nil model slices.

ethanfrey:
Thanks for the cleanup. Looks great

PR #26, opened 2020-01-17
Fix 14: Allow configuring LRU cache size in app.toml

### LINKED ISSUE #14: Allow configuring LRU cache size in app.toml
## Summary

Node operators should be able to configure LRU cache size based on their memory availability

## Problem Definition

Storing instances in the LRU will have no effect on the results (still deterministic), but should lower execution time at the cost of increased memory usage. We cannot pick universal parameters for this, so we should allow node operators to set it.

## Proposal

There seems to be an `app.toml` that the node operator can configure some subjective values like `halt-height` and `min-gas-fee` or such. Let's add one more value:

```toml
[wasm]
lru_size = 10
```

The default should be 0, which means disabled. This value must be parsed at start and is passed into the wasm.NewKeeper function. We currently have hardcoded the value to 3: https://github.com/cosmwasm/wasmd/blob/master/x/wasm/internal/keeper/keeper.go#L44
____

#### For Admin Use

- [ ] Not duplicate issue

anilCSE:
Closes #14 

- [x] Linked to github-issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Wrote tests

______

For Admin Use:
- Reviewers Assigned


app/app.go
```diff
@@ -181,7 +183,14 @@ func NewWasmApp(
 	// better way to get this dir???
 	homeDir := viper.GetString(cli.HomeFlag)
 	wasmDir := filepath.Join(homeDir, "wasm")
-	app.wasmKeeper = wasm.NewKeeper(app.cdc, keys[wasm.StoreKey], app.accountKeeper, app.bankKeeper, wasmRouter, wasmDir)
+
+	// Read LRU cache size from config (app.toml), defaults to `0`
```
ethanfrey: app.toml is parsed into a Config object here: https://github.com/cosmos/cosmos-sdk/blob/4d0dfe592b45b658d796d9b2eb6f2eb5cfe68fc1/server/config/config.go#L15-L45
We don't need to update that as well somehow?

app/app.go
```diff
@@ -181,7 +183,14 @@ func NewWasmApp(
...
+	cacheSize := uint64(viper.GetInt64(flagLRUCacheSize))
+	// Read query gas limit from config
+	smartQueryGasLimit := uint64(viper.GetInt64(flagQueryGasLimit))
+
+	app.wasmKeeper = wasm.NewKeeper(app.cdc, keys[wasm.StoreKey], app.accountKeeper, app.bankKeeper, wasmRouter,
```
ethanfrey: Even if we don't need to modify the sdk Config, how about putting these into a `WasmConfig` struct, so we don't need to add more args here, even if we add more options to the config (and I bet we will)

app/app.go
ethanfrey: Also, I note these have no defaults...
EIther we need defaults here, or (probably better) on the `WasmConfig` struct.

I would propose defaults: 

```go
WasmConfig {
  SmartQueryGasLimit: 300000,
  CacheSize: 0,
}
```

LRU cache should be opt-in. It should be impossible to set 0 gas limit (1 okay) for smart queries, so this means no issue with "unable to set 0 in config" issues.

anilCSE:
> Interesting, I didn't know it was so easy to add this to `app.toml`.
> 
> Can you make some changes to the code flow (defaults and config struct)?
> Maybe some flow like this would work to parse into wasmConfig:
> https://github.com/cosmos/cosmos-sdk/blob/4d0dfe592b45b658d796d9b2eb6f2eb5cfe68fc1/server/config/toml.go#L54-L61
> 
> Also, if possible, I would like to have a separate section in the toml:
> 
> ```toml
> [wasm]
> lru_size = 10
> ```
> 
> This should be easier to define with the Config format. If too hard, leave top-level, it is a minor preference.

As far as I see, there's no way to do this from `app`. Node operators should manually add the extra config entries to `app.toml` file after it gets created.

This should be raised as an issue against cosmos-sdk, to allow overrride or extend the `defaultConfigTemplate` to create custom app.toml. 
Needed a way to supply custom template here https://github.com/cosmos/cosmos-sdk/blob/4d0dfe592b45b658d796d9b2eb6f2eb5cfe68fc1/server/config/toml.go#L46
 or 
set/override `defaultConfigTemplate` from `app.go` before `init()`

app/app.go
anilCSE: Makes sense. Updated the required changes and its r4r

ethanfrey:
Yes, the config template is impossible to override without forking the sdk.

I added the `[wasm]` subsection and a little explanation in a README. This whole module needs more docs, but step by step.

PR #27, opened 2020-01-17
Fix #11: Add source ref

### LINKED ISSUE #11: Add source code reference to Code objects
## Summary

Help the developer annotate where the source code for a contract is.
This is not validated, but can allow a client to easily check the code out and validate if it matches (via deterministic build system).

## Problem Definition

It is important to validate the source code. Solutions like Etherscan store all this info on a centralized database. Adding a field or two can make this much more decentralized.

## Proposal

We want to add two fields:
* `Source string` is an optional field that, if present, must be a valid URI that can reference the source code
* `Builder string` is an optional field that, if present, is a docker tag for a deterministic build system. This will not be run blindly, but the client can check it against some whitelist. For example `cosmwasm-opt:0.6.0` or `cosmwasm-opt:0.5.2`. We need this info to properly verify (as we need the same build tag to properly reproduce).

Please update [`MsgStoreCode`](https://github.com/cosmwasm/wasmd/blob/master/x/wasm/internal/types/msg.go#L11-L14), [`CodeInfo`](https://github.com/cosmwasm/wasmd/blob/master/x/wasm/internal/types/types.go#L10-L13), and any relevant queries
____

#### For Admin Use

- [ ] Not duplicate issue

anilCSE:
Closes: #11 

- [x] Linked to github-issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Wrote tests

______

For Admin Use:
- Reviewers Assigned


x/wasm/client/cli/tx.go
```diff
@@ -21,13 +22,20 @@ import (
...
+	flagAmount  = "amount"
+	flagSource  = "source"
+	flagBuilder = "builder"
 )
 
 // limit max bytes read to prevent gzip bombs
 const maxSize = 400 * 1024
 
+// whitelist
+var validBuildTags = map[string]bool{
```
ethanfrey: Nice check, but I would prefer to whitelist all in valid format on the server, and let the clients evolve faster (rather than need a blockchain upgrade for 0.6.1 or 0.6.2

x/wasm/client/cli/tx.go
```diff
@@ -62,6 +70,26 @@ func StoreCodeCmd(cdc *codec.Codec) *cobra.Command {
 				return err
 			}
 
+			source := viper.GetString(flagSource)
+
+			// ensure source to be a valid uri
+			if source != "" {
+				_, err := url.Parse(source)
```
ethanfrey: Yes

x/wasm/client/cli/tx.go
```diff
@@ -62,6 +70,26 @@ func StoreCodeCmd(cdc *codec.Codec) *cobra.Command {
...
+					return fmt.Errorf("invalid url supplied for source %s", source)
+				}
+			}
+
+			builder := viper.GetString(flagBuilder)
+
+			// ensure builder to be a valid build tag
+			if builder != "" {
+				if !validBuildTags[builder] {
+					return fmt.Errorf("invalid tag supplied for builder %s", source)
```
ethanfrey: As mentioned above, cab you use a regexp here instead of whitelist?

x/wasm/internal/keeper/genesis.go
```diff
@@ -14,7 +14,7 @@ import (
 // CONTRACT: all types of accounts must have been already initialized/created
 func InitGenesis(ctx sdk.Context, keeper Keeper, data types.GenesisState) {
 	for _, code := range data.Codes {
-		newId, err := keeper.Create(ctx, code.CodeInfo.Creator, code.CodesBytes)
+		newId, err := keeper.Create(ctx, code.CodeInfo.Creator, code.CodesBytes, code.CodeInfo.Source, code.CodeInfo.Builder)
```
ethanfrey: Can you update the genesis test to include these fields?

x/wasm/internal/keeper/keeper_test.go
```diff
@@ -138,7 +138,7 @@ func TestExecute(t *testing.T) {
 	wasmCode, err := ioutil.ReadFile("./testdata/contract.wasm")
 	require.NoError(t, err)
 
-	contractID, err := keeper.Create(ctx, creator, wasmCode)
+	contractID, err := keeper.Create(ctx, creator, wasmCode, "", "")
```
ethanfrey: How about setting some of these to non empty in the tests

anilCSE:
> Nice start, this is what I had in mind.
> 
> However, as I see it, the format is only enforced in the cli and not in the keeper. (Unkess I missed that). Can you enforce that in eg. MsgCreate.ValidateBasic() and add some tests for that?
> 
> That an a few more tests and this is ready to merge

Updated `msg.ValidateBasic()` to handle this at `cli` and `handler` level. Also moved build regex from `cli` to `msg.ValidateBasic()`. May be we can replace with better regex expression.

x/wasm/internal/keeper/genesis.go
anilCSE: Done

x/wasm/client/cli/tx.go
```diff
@@ -62,11 +61,9 @@ func StoreCodeCmd(cdc *codec.Codec) *cobra.Command {
 				return err
 			}
 
-			// limit the input size
```
ethanfrey: Why is this check removed?

Otherwise, the function looks good

x/wasm/handler.go
```diff
@@ -40,7 +40,12 @@ func NewHandler(k Keeper) sdk.Handler {
 }
 
 func handleStoreCode(ctx sdk.Context, k Keeper, msg *MsgStoreCode) sdk.Result {
-	codeID, err := k.Create(ctx, msg.Sender, msg.WASMByteCode)
+	_err := msg.ValidateBasic()
```
ethanfrey: Why `_err`?
Custom is to use `sdkerr` for this case? (As `err` has different type, so you need two)

x/wasm/internal/types/msg.go
```diff
@@ -26,9 +36,26 @@ func (msg MsgStoreCode) ValidateBasic() sdk.Error {
...
 	}
+
 	if len(msg.WASMByteCode) > MaxWasmSize {
 		return sdk.ErrInternal("wasm code too large")
 	}
+
+	if msg.Source != "" {
+		fmt.Println("source: ", msg.Source)
+
+		if !govalidator.IsURL(msg.Source) {
```
ethanfrey: I'm not strongly against adding more dependencies, but I like to see the utility.
Bringing in govalidator for one line seems too much.

You can use: https://golang.org/pkg/net/url/#Parse from stdlib, which should have the same functionality

x/wasm/internal/types/msg.go
```diff
@@ -1,17 +1,27 @@
...
+
+	"github.com/asaskevich/govalidator"
+
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
 const (
-	MaxWasmSize = 500 * 1024
+	MaxWasmSize   = 500 * 1024
+	BuildTagRegex = "cosmwasm-opt:"
```
ethanfrey: Maybe `"^cosmwasm-opt:"` so it can't start with other characters? Otherwise, that seems fine enough... some limits but not terribly restrictive

x/wasm/client/cli/tx.go
ethanfrey: Nevermind... I saw this is now handled by `ValidateBasic`. 
Very proper to remove it here to avoid duplication.

x/wasm/internal/types/msg.go
anilCSE: We used `net/url.Parse()` previously but the method has a few weaknesses. If you change the url value to `somerandomstring#43243@` also `net/url.Parse()` function will still pass the broken URL as `valid`. This is NOT the right way to validate URL.

anilCSE: May be we can do the following.
- check if url is reachable
- check if the url is absolute (we shouldn't encourage relative urls for source anyway)

anilCSE: Added these changes and updated genesis tests to cover these cases.

PR #39, opened 2020-01-21
WIP: Dockerize

### LINKED ISSUE #38: Create working docker image for CI/testing
## Summary

The current docker build doesn't work (alpine) and only allows running one process.

We want:

1. to update the Dockerfile to work with wasmd (glibc and static linking seem to be issues)
2. produce a script so we can run `wasmd` and `wasmcli rest-server` together in one container. It should include some persistence (mount external volume) to store data and keys. 

This does not have to be production secure. It should be easy to set up for tests.

## Problem Definition

We want to allow contract developers and frontend developers an easy way to stand up a node to run against, without having to install a bunch of go packages locally and fumble with shell scripts.

## Proposal

See above
____

#### For Admin Use

- [ ] Not duplicate issue

ethanfrey:
Closes #38 

Check README for how to run local Docker image for testing

- [ ] Linked to github-issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Wrote tests

______

For Admin Use:
- Reviewers Assigned


ethanfrey:
@alpe I thought last night and there are at least 2 missing features here:

1. The `setup.sh` should take a list of addresses. The script can then loop through `$@` and run `add-genesis-account` for each of those. They can all have the same amount.
2. There should be two forms to run. One (current `run.sh`) assumes we mount an initialized directory (from `setup.sh` or a previous run) and just write to it. The other is designed for CI. `setup.sh` can write to one volume under `/root`. Then for run, we mount to eg. `/template` and the other `run_ci.sh` script can copy all files from `/template/.wasm{d,cli}` into `/root/*` (which is in the docker tempfs) and then start the normal run script. This lets us initialize setup once (which requires interactive password entering), and then use that initialized data for CI runs without overwriting it.

If you can think of anything else useful for local testing/CI, please add that as well.

alpe:
Very nice to see this working now! 🎉 

> 1. The `setup.sh` should take a list of addresses. The script can then loop through `$@` and run `add-genesis-account` for each of those. They can all have the same amount.

This would not be hard to implement but maybe hard to communicate. Instead of running a simple script people would need to think about arguments now and remember them for later use. My second note is about the password entry. It is already not very satisfying to run the script for generating a single key. If we can use a default password that's not an issue.

> 2. There should be two forms to run. One (current `run.sh`) assumes we mount an initialized directory (from `setup.sh` or a previous run) and just write to it. The other is designed for CI. `setup.sh` can write ...

Not sure if I understood it correct. For CI we would need some defined state. IMHO this can be achieved best by mounting predefined `.wasm{d,cli}/` folders from the repo into the container under `/root`.  No `setup.sh` execution during CI.

README.md
```diff
@@ -26,3 +26,30 @@ To set up a single node testnet, [look at the deployment documentation](./docs/d
...
+Build: `docker build  -t wasmd:manual .`
+
+Run:
+
+```sh
+docker volume rm -f wasmd_data
+
+# pick a simple (8 char) passphrase for testing.. you will type it many times
+docker run --rm -it \
+    --mount type=volume,source=wasmd_data,target=/root \
```
alpe: pp: the `-v` volume syntax is a bit more common. You can use `--volume wasmd_data:/root \` here and below as well.

README.md
```diff
@@ -26,3 +26,30 @@ To set up a single node testnet, [look at the deployment documentation](./docs/d
...
+    wasmd:manual ./setup.sh
+
+# This will start both wasmd and wasmcli rest-server, only wasmcli output is shown on the screen
+docker run --rm -it -p 26657:26657 -p 26656:26656 -p 1317:1317 \
+    --mount type=volume,source=wasmd_data,target=/root \
+    wasmd:manual ./run.sh
+
+# view wasmd logs in another shell
+docker run --rm -it \
+    --mount type=volume,source=wasmd_data,target=/root,readonly \
```
alpe: nice `read only` setup on the volume. This would be `--volume wasmd_data:/root:ro \`

Dockerfile_demo
```diff
@@ -0,0 +1,30 @@
+# Simple usage with a mounted data directory:
```
alpe: let's delete this file

Dockerfile
```diff
@@ -2,32 +2,31 @@
 # > docker build -t gaia .
 # > docker run -it -p 46657:46657 -p 46656:46656 -v ~/.wasmd:/root/.wasmd -v ~/.wasmcli:/root/.wasmcli gaia wasmd init
 # > docker run -it -p 46657:46657 -p 46656:46656 -v ~/.wasmd:/root/.wasmd -v ~/.wasmcli:/root/.wasmcli gaia wasmd start
-FROM golang:alpine AS build-env
+FROM golang:1.13-buster AS build-env
```
alpe: The comments above are outdated. Proper ones are in the README. let's remove them here.

Dockerfile
```diff
@@ -2,32 +2,31 @@
...
 
-# Copy over binaries from the build-env
-COPY --from=build-env /go/bin/wasmd /usr/bin/wasmd
-COPY --from=build-env /go/bin/wasmcli /usr/bin/wasmcli
+# rest server
+EXPOSE 1317
+# tendermint p2p
+EXPOSE 26656
+# tendermint rpc
+EXPOSE 26657
```
alpe: 👍 good for doc

Dockerfile_demo
alpe: and the the `docker_demo.md`

ethanfrey:
> Not sure if I understood it correct. For CI we would need some defined state. IMHO this can be achieved best by mounting predefined .wasm{d,cli}/ folders from the repo into the container under /root. No setup.sh execution during CI.

My idea is I, dev, can run setup.sh to make an initial CI state. And I save that to disk/commit to git repo. Thus, I also want setup to be easy to use for such cases.

In the CI I only use run with the predefined state and it doesn't need a password

ethanfrey:
> My second note is about the password entry. It is already not very satisfying to run the script for generating a single key. If we can use a default password that's not an issue.

If you can get the script to run non interactive without password entry, even default password, that would be great. Maybe another name to leave both options?

There was just a PR #40 about using Keyring headless, maybe that has some useful tips

README.md
ethanfrey: Sure, but all docker docs say it is deprecated.

Change if you wish, but I want to show with docker volumes not familiar dir and this was the only way i knew

Dockerfile
ethanfrey: Yes, let's remove all docs except readme

README.md
alpe: `volume` is not deprecated but `mount` is recommended. I didn't know that. thanks! Let's leave it like it is now.
> Even though there is no plan to deprecate --volume, usage of --mount is recommended.

https://docs.docker.com/engine/reference/commandline/run/

PR #58, opened 2020-01-30
Rename docker denoms

### LINKED ISSUE #57: Rename denom to ucosm and ustake
The [test chain from docker](https://hub.docker.com/r/cosmwasm/wasmd) comes with two tokens, COSM and STAKE. Both have 6 fractional digits. Thus the `denom` value should be µCOSM and µSTAKE, or "ucosm" and "ustake" for consistency with ATOMs.

This needs to be updated in client-side configuration once it is changed in the blockchain config.

ethanfrey:
Closes #57 

Makes `ustake` the staking/governance token. `ucosm` is a second fee token

- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Wrote tests

______

For admin use:

- [ ] Reviewers assigned

### DIFF
--- a/docker/setup.sh
+++ b/docker/setup.sh
@@ -5,10 +5,10 @@ PASSWORD=${PASSWORD:-1234567890}
 wasmd init --chain-id=testing testing
 (echo $PASSWORD; echo $PASSWORD) | wasmcli keys add validator
 # hardcode the validator account for this instance
-echo $PASSWORD | wasmd add-genesis-account validator 1000000000stake,1000000000cosm
+echo $PASSWORD | wasmd add-genesis-account validator 1000000000ustake,1000000000ucosm
 # (optionally) add a few more genesis accounts
 for addr in "$@"; do
-  wasmd add-genesis-account $addr 1000000000stake,1000000000cosm
+  wasmd add-genesis-account $addr 1000000000ustake,1000000000ucosm
 done
 # submit a genesis validator tx
 (echo $PASSWORD; echo $PASSWORD; echo $PASSWORD) | wasmd gentx --name validator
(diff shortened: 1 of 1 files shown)


docker/setup.sh
```diff
@@ -3,13 +3,15 @@
 PASSWORD=${PASSWORD:-1234567890}
 
 wasmd init --chain-id=testing testing
+# staking/governance token is hardcoded in config, change this
+sed -i 's/"stake"/"ustake"/' "$HOME"/.wasmd/config/genesis.json
```
webmaster128: no similar diff for cosm -> ucosm? Is the token created by the first account receiving it?

ethanfrey: The account balancea are defines un the creaste account command below.

The default genesis hard codes stake as token for staking validators and voting in governance. Sed fixes that (no cli flag)

ethanfrey: Actually, maybe we make both denoms args to this script, so we control in the package where we run in CI, not here. And then default to these denoms if $1 and $2 are undefined

webmaster128: But where is the COSM token configuted? At which point does the token come into existence?

ethanfrey: When you add it to a genesis account.

There is no explicit definition of tokens or minting.
They look at balances in genesis and those are the tokens on the chain.

I know, you miss weave :)

webmaster128: Okay cool, thanks, that answers my question.

And it answers much more than that. In order to get a list of chain tokens, you need to iterate over all accounts on the chain 🤯

ethanfrey: They added a supply module recently that basically does that iteration. I should give a list of denoms, but no fractional digits or other info

PR #67, opened 2020-02-09
Fix setup.sh

webmaster128:
Before, the optional positional arguments `$1` and `$2` confliced with the addresses read in `"$@"`


ethanfrey:
Facepalm.

Each were decent ideas, together not so....

PR #115, opened 2020-05-25
adjusted gas consumption logic in case of errors in go-cosmwasm

reuvenpo:
see:
https://github.com/CosmWasm/go-cosmwasm/pull/90
https://github.com/CosmWasm/cosmwasm/pull/359
and:
https://github.com/CosmWasm/go-cosmwasm/pull/91
https://github.com/CosmWasm/cosmwasm/pull/379

- [ ] Wrote tests

______

For admin use:

- [ ] Reviewers assigned


x/wasm/internal/keeper/keeper.go
```diff
@@ -138,15 +138,11 @@ func (k Keeper) Instantiate(ctx sdk.Context, codeID uint64, creator sdk.AccAddre
 
 	// instantiate wasm contract
 	gas := gasForContract(ctx)
-	res, err := k.wasmer.Instantiate(codeInfo.CodeHash, params, initMsg, prefixStore, cosmwasmAPI, querier, gas)
+	res, gasUsed, err := k.wasmer.Instantiate(codeInfo.CodeHash, params, initMsg, prefixStore, cosmwasmAPI, querier, gas)
+	consumeGas(ctx, gasUsed)
```
reuvenpo, line 142: Shouldn't this be divided by the `GasMultiplier`?

ethanfrey: That is done inside `consumeGas`

reuvenpo: Ah right, I forgot that when reviewing the code

x/wasm/internal/keeper/keeper.go
```diff
@@ -138,15 +138,11 @@ func (k Keeper) Instantiate(ctx sdk.Context, codeID uint64, creator sdk.AccAddre
...
 	if err != nil {
-		// TODO: wasmer doesn't return wasm gas used on error. we should consume it (for error on metering failure)
-		// Note: OutOfGas panics (from storage) are caught by go-cosmwasm, subtract one more gas to check if
-		// this contract died due to gas limit in Storage
-		consumeGas(ctx, GasMultiplier)
 		return contractAddress, sdkerrors.Wrap(types.ErrInstantiateFailed, err.Error())
 	}
-	consumeGas(ctx, res.GasUsed)
```
ethanfrey, line 149: You can remove `GasUsed` from `res` as we now return this as a separate return value from `go-cosmwasm` rather than a member of the struct.

reuvenpo: `res` is reported back to the user as part of the response, right? i think it can be useful for the user to know how much gas was actually consumed (although we should divide it by 100 i think)

ethanfrey: No, that is a `go-cosmwasm/types/Result` (we need to change the struct there).

We strip out most of the info there when converting it into a "cosmos SDK" result: https://github.com/CosmWasm/wasmd/blob/master/x/wasm/internal/types/types.go#L137-L155

We only need to propogate `Data` and `Log`

reuvenpo: ahhh, i see...
Sure, i'll remove the field completely

reuvenpo: fixed in go-cosmwasm and pulled here in 7758122

PR #124, opened 2020-06-03
Update contract admin

### LINKED ISSUE #117: Add MsgUpdateAdministrator
We will add another message type MsgUpdateAdministrator which can only be executed by the current administrator (if set) and will replace it with a new administrator (or empty). Empty administrator is the current  status quo and allows no migrations (immutable). If the administrator is set, it defines who can upgrade the contract. This may be under direct control of a private key, or may be controlled by a multisig contract (or eventually a native "group account"). This should allow flexibility for "committee governance" to manage this stage of the lifecycle.

alpe:
Resolves #117 

--
- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [x] Wrote tests
- [x] Added  relevant `godoc` [comments](https://blog.golang.org/godoc-documenting-go-code).
- [x] Added a relevant changelog entry to the `Unreleased` section in `CHANGELOG.md`
- [x] Re-reviewed `Files changed` in the Github PR explorer

______

For admin use:

- [ ] Reviewers assigned

### DIFF
--- a/x/wasm/handler.go
+++ b/x/wasm/handler.go
@@ -39,6 +39,11 @@ func NewHandler(k Keeper) sdk.Handler {
 		case MsgMigrateContract:
 			return handleMigration(ctx, k, &msg)
 
+		case *MsgUpdateAdministrator:
+			return handleUpdateContractAdmin(ctx, k, msg)
+		case MsgUpdateAdministrator:
+			return handleUpdateContractAdmin(ctx, k, &msg)
+
 		default:
 			errMsg := fmt.Sprintf("unrecognized wasm message type: %T", msg)
 			return nil, sdkerrors.Wrap(sdkerrors.ErrUnknownRequest, errMsg)
@@ -138,4 +143,21 @@ func handleMigration(ctx sdk.Context, k Keeper, msg *MsgMigrateContract) (*sdk.R
 	)
 	res.Events = append(events, ourEvent)
 	return res, nil
+}
+
+func handleUpdateContractAdmin(ctx sdk.Context, k Keeper, msg *MsgUpdateAdministrator) (*sdk.Result, error) {
+	if err := k.UpdateContractAdmin(ctx, msg.Contract, msg.Sender, msg.NewAdmin); err != nil {
+		return nil, err
+	}
+	events := filterMessageEvents(ctx.EventManager())
+	ourEvent := sdk.NewEvent(
+		sdk.EventTypeMessage,
+		sdk.NewAttribute(sdk.AttributeKeyModule, ModuleName),
+		sdk.NewAttribute(AttributeSigner, msg.Sender.String()),
+		sdk.NewAttribute(AttributeKeyContract, msg.Contract.String()),
+	)
+	return &sdk.Result{
+		//Data:   , TODO: anything to return here?
+		Events: append(events, ourEvent),
+	}, nil
 }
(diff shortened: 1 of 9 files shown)


x/wasm/handler.go
```diff
@@ -139,3 +144,20 @@ func handleMigration(ctx sdk.Context, k Keeper, msg *MsgMigrateContract) (*sdk.R
 	res.Events = append(events, ourEvent)
 	return res, nil
 }
+
+func handleUpdateContractAdmin(ctx sdk.Context, k Keeper, msg *MsgUpdateAdministrator) (*sdk.Result, error) {
+	if err := k.UpdateContractAdmin(ctx, msg.Contract, msg.Sender, msg.NewAdmin); err != nil {
+		return nil, err
+	}
+	events := filterMessageEvents(ctx.EventManager())
```
ethanfrey, line 152: This filtering stuff is only needed for the calls that dispatch to other handlers via the routers, as they would end up with ots of `sdk.EventTypeMessage` events that got merged together. I think you can just do

`events := ctx.EventManager().Events()`

here or something similar (like other modules do).

x/wasm/internal/keeper/keeper.go
```diff
@@ -255,6 +255,23 @@ func (k Keeper) Migrate(ctx sdk.Context, contractAddress sdk.AccAddress, caller
...
+		return sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, "unknown contract")
+	}
+	if contractInfo.Admin == nil {
+		return sdkerrors.Wrap(sdkerrors.ErrUnauthorized, "migration not supported by this contract")
+	}
+	if !contractInfo.Admin.Equals(caller) {
+		return sdkerrors.Wrap(sdkerrors.ErrUnauthorized, "no permission")
+	}
+	contractInfo.Admin = newAdmin
+	k.setContractInfo(ctx, contractAddress, *contractInfo)
```
ethanfrey, line 271: pp: `GetContactInfo` returns `*ContractInfo`, but `setContractInfo` accepts `ContractInfo`. Nice to have them both take pointers or both structs, not a mix.

But if there is a technical reason for it, no big deal, just seems like an odd mismatch

x/wasm/internal/types/msg.go
```diff
@@ -227,3 +227,43 @@ func (msg MsgMigrateContract) GetSignBytes() []byte {
...
+}
+
+func (msg MsgUpdateAdministrator) ValidateBasic() error {
+	if err := sdk.VerifyAddressFormat(msg.Sender); err != nil {
+		return sdkerrors.Wrap(err, "sender")
+	}
+	if err := sdk.VerifyAddressFormat(msg.Contract); err != nil {
+		return sdkerrors.Wrap(err, "contract")
+	}
+	if len(msg.NewAdmin) != 0 {
```
ethanfrey, line 252: good checks here

x/wasm/internal/keeper/keeper_test.go
```diff
@@ -568,6 +568,84 @@ func TestMigrate(t *testing.T) {
...
+	}
+	initMsgBz, err := json.Marshal(initMsg)
+	require.NoError(t, err)
+	specs := map[string]struct {
+		instAdmin            sdk.AccAddress
+		newAdmin             sdk.AccAddress
+		overrideContractAddr sdk.AccAddress
+		caller               sdk.AccAddress
+		expErr               *sdkerrors.Error
+	}{
```
ethanfrey, line 602: nice tests

x/wasm/internal/keeper/keeper.go
alpe: 🤔 `GetContactInfo` supports nil value as a result. I don't know if this is technically possible but refactoring this to a non pointer type would probably require an `error` return type as well. I will make them both pointers as a quick solution. I don't have a strong opinion. For me values are usually cleaner in api design as there are no modifications. But `setContractInfo` is a private method so does not matter much.

PR #125, opened 2020-06-03
Track gas in queries

reuvenpo:
This PR is complementary to https://github.com/CosmWasm/go-cosmwasm/pull/98 and adds the `GasConsumed` method to `QueryHandler`.
Once the go-cosmwasm PR is merged, a commit should be added here that points to go-cosmwasm dependency to the latest commit on the `0.9` branch

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/CosmWasm/wasmd
 go 1.13
 
 require (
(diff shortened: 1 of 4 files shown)


reuvenpo:
>But shouldn't you update the go-cosmwasm dep as well?
As I said, i'll update it after the go-cosmwasm PR is merged, and point them to the merge commit on 0.9

PR #137, opened 2020-06-09
Execute: propagate wasm events to tendermint

assafmo:

- [ ] Wrote tests

______

For admin use:

- [ ] Reviewers assigned

### DIFF
--- a/x/wasm/handler.go
+++ b/x/wasm/handler.go
@@ -6,7 +6,7 @@ import (
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
-	cosmwasmtypes "github.com/enigmampc/EnigmaBlockchain/go-cosmwasm/types"
+	cosmwasmtypes "github.com/CosmWasm/go-cosmwasm/types"
 )
 
 const (
--- a/x/wasm/handler.go
+++ b/x/wasm/handler.go
@@ -5,6 +5,11 @@ import (
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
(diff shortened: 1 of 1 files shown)


x/wasm/handler.go
```diff
@@ -125,6 +127,22 @@ func handleExecute(ctx sdk.Context, k Keeper, msg *MsgExecuteContract) (*sdk.Res
 	)
 
 	res.Events = append(events, ourEvent)
+	
+	var resp cosmwasmtypes.CosmosResponse
```
ethanfrey, line 131: Uh, this should all have been [handled in Keeper.Execute](https://github.com/CosmWasm/wasmd/blob/master/x/wasm/internal/keeper/keeper.go#L196-L198)

We `res.Data` in this function comes directly from the wasm result.
We also added all events from the wasmResult to the event manager and they are handled right above.

We have tested this with cosmwasm-js.

What seems to be happening to me is you have double-enoded the result.

`res.Data` is created by parsing the original result and then taking the Data field. So if you embed a serialized CosmosResponse inside another result you return from the contract I can see this happening. Double-check the return values from the contracts and debug the keeper.Execute code.

Also, a failing test case showing this is broken is a good first step :smile:

assafmo: Thanks, I'll take a look

assafmo: @ethanfrey I now see the on the version I have there was a bug inside `x/wasmd/internal/keeper/keeper.go` in `Execute`, where
```
var result wasmTypes.Result
err = json.Unmarshal(res, &result)
```
should have been
```
var result wasmTypes.CosmosResponse
err = json.Unmarshal(res, &result)
```
This cause the unmarshal to output an empty struct and this is why I reimplemented this in `handler.go`

Anyway, we're now pulling v0.9.0 so this will be fixed.

Thanks, @ethanfrey! :hearts:

PR #157, opened 2020-06-26
Split MsgUpdateAdministrator to MsgUpdateAdmin and MsgClearAdmin

### LINKED ISSUE #153: Rename MsgUpdateAdministrator -> MsgUpdateAdmin
The role should be called "admin" consistently. If there is "admin" in the state, this is what we want to update.

### LINKED ISSUE #155: Split message into MsgUpdateAdmin and MsgClearAdmin
At the JSON/Go boundary it is not possible to differentiate between an empty string, a null value and an unset value. This makes it impossible to differentiate between _new_admin not set_ and _new_admin is set to null_, which are are different things. Right now, sending this message clears the admin:

```json
{
  "type": "wasm/update-contract-admin",
  "value": {
    "contract": "cosmos1x20375el6y8kht67znuyxcmh53jp0g2av265nl",
    "sender": "cosmos14qemq0vw6y3gc3u3e0aty2e764u4gs5le3hada"
  }
}
```

but this could as well be a incorrectly encoded message (forgot to set the field). Right now, an additional field like `newadmin` causes a verification failure, but with protobuf this will not be the case anymore.

For most cases (e.g. the memo field), I think it is fine to treat unset and empty equally. But for clearing data I don't think the message expresses the intend. Thus I think we should split this into two messages:

- `MsgUpdateAdmin` where `new_admin` must be a valid address of this chain
- `MsgClearAdmin` explicitely clears the admin

ethanfrey:
Closes #153 
Closes #155 

- [ ] Wrote tests

______

For admin use:

- [ ] Reviewers assigned

### DIFF
--- a/x/wasm/client/cli/new_tx.go
+++ b/x/wasm/client/cli/new_tx.go
@@ -2,7 +2,6 @@ package cli
 
 import (
 	"bufio"
-	"errors"
 	"strconv"
 
 	"github.com/CosmWasm/wasmd/x/wasm/internal/types"
@@ -13,7 +12,6 @@ import (
 	"github.com/cosmos/cosmos-sdk/x/auth"
 	"github.com/cosmos/cosmos-sdk/x/auth/client/utils"
 	"github.com/spf13/cobra"
-	"github.com/spf13/viper"
 )
 
 // MigrateContractCmd will migrate a contract to a new code version
@@ -52,12 +50,12 @@ func MigrateContractCmd(cdc *codec.Codec) *cobra.Command {
 	return cmd
 }
 
-// UpdateContractAdminCmd sets or clears an admin for a contract
+// UpdateContractAdminCmd sets an new admin for a contract
 func UpdateContractAdminCmd(cdc *codec.Codec) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "set-contract-admin [contract_addr_bech32] [new_admin_addr_bech32]",
-		Short: "Set new admin for a contract. Can be empty to prevent further migrations",
-		Args:  cobra.RangeArgs(1, 2),
+		Short: "Set new admin for a contract",
+		Args:  cobra.ExactArgs(2),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			inBuf := bufio.NewReader(cmd.InOrStdin())
 			txBldr := auth.NewTxBuilderFromCLI(inBuf).WithTxEncoder(utils.GetTxEncoder(cdc))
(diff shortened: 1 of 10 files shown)


x/wasm/client/cli/new_tx.go
```diff
@@ -81,7 +81,7 @@ func UpdateContractAdminCmd(cdc *codec.Codec) *cobra.Command {
 				}
 			}
 
-			msg := types.MsgUpdateAdministrator{
+			msg := types.MsgUpdateAdmin{
```
webmaster128, line 84: The rest of this function needs some updates as well. E.g.
- text "Set new admin for a contract. Can be empty to prevent further migrations"
- Block starting with `// safety net to not accidentally clear an admin`

x/wasm/internal/keeper/keeper.go
```diff
@@ -263,7 +263,7 @@ func (k Keeper) Migrate(ctx sdk.Context, contractAddress sdk.AccAddress, caller
 	return types.ResultFromData(res.Data), nil
 }
 
-// UpdateContractAdmin sets the admin value on the ContractInfo. New admin can be nil to disable further migrations/ updates.
+// UpdateContractAdmin sets the admin value on the ContractInfo. It must be a valid address (use ClearContractAdmin to remove it)
 func (k Keeper) UpdateContractAdmin(ctx sdk.Context, contractAddress sdk.AccAddress, caller sdk.AccAddress, newAdmin sdk.AccAddress) error {
```
webmaster128, line 267: If that code worked for clearing before, don't we need an additional check now?

x/wasm/internal/types/msg.go
```diff
@@ -228,42 +228,71 @@ func (msg MsgMigrateContract) GetSigners() []sdk.AccAddress {
 	return []sdk.AccAddress{msg.Sender}
 }
 
-type MsgUpdateAdministrator struct {
+type MsgUpdateAdmin struct {
 	Sender   sdk.AccAddress `json:"sender" yaml:"sender"`
 	NewAdmin sdk.AccAddress `json:"new_admin,omitempty" yaml:"new_admin"`
```
webmaster128, line 233: We can remove the `,omitempty` here, right?

x/wasm/client/cli/new_tx.go
ethanfrey: Yup, and I added a clear admin cmd. Forgot about this cli stuff...

x/wasm/internal/keeper/keeper.go
ethanfrey: This is checked in `ValidateBasic`: https://github.com/CosmWasm/wasmd/pull/157/files#diff-47cb7ef32f0dc71704a9c4ac79c93817L252

I can check again, but that is guaranteed to be called before processing a message and is the standard way to enforce such guarantees (same with enforcing the signers). I personally dislike the magic/non-transparency but this is the cosmos-sdk-style

webmaster128: Alright, thanks for clarification. Makes sense how it is.

The whole idea of struct validation is questionable. A strict type should never allow instantiating invalid data. Then you only have one way to create that type and this does all the checks. But this is a discussion for some other place and time.

PR #165, opened 2020-06-30
Validate genesis model

alpe:
A data import via genesis file allows currently to bypass all validation checks of the models so that the internal state could end up in an inconsistent or broken state.

With this PR some syntax checks on the model were added as well as a sanity check on the DB import to prevent overwrites by duplicate data. Orphan contracts without a persistent codeInfo are prevented, now.

* Not covered:
The position of a `code_info` within the `code` block matters for the codeID generation during import. Switching two elements can therefore lead to different codeIDs and may cause contracts to point to the wrong one.
 See https://github.com/CosmWasm/wasmd/issues/166
______

For admin use:

- [ ] Reviewers assigned

### DIFF
--- a/x/wasm/internal/keeper/genesis_test.go
+++ b/x/wasm/internal/keeper/genesis_test.go
@@ -101,7 +101,7 @@ func TestFailFastImport(t *testing.T) {
 			}},
 			Contracts: nil,
 		}},
-		"happy path: code info and contract do match": {
+		"happy path: code id in info and contract do match": {
 			src: types.GenesisState{
 				Codes: []types.Code{{
 					CodeInfo: wasmTypes.CodeInfo{
@@ -112,13 +112,8 @@ func TestFailFastImport(t *testing.T) {
 				}},
 				Contracts: []types.Contract{
 					{
-						ContractAddress: addrFromUint64(1<<32 + 1),
-						ContractInfo: wasmTypes.ContractInfo{
-							CodeID:  1,
-							Creator: anyAddress,
-							Label:   "any",
-							Created: &types.AbsoluteTxPosition{BlockHeight: 1, TxIndex: 1},
-						},
+						ContractAddress: contractAddress(1, 1),
+						ContractInfo:    types.ContractInfoFixture(func(c *wasmTypes.ContractInfo) { c.CodeID = 1 }),
 					},
 				},
 			},
@@ -135,21 +130,11 @@ func TestFailFastImport(t *testing.T) {
 				}},
 				Contracts: []types.Contract{
 					{
-						ContractAddress: addrFromUint64(1<<32 + 1),
-						ContractInfo: wasmTypes.ContractInfo{
-							CodeID:  1,
-							Creator: anyAddress,
-							Label:   "any",
-							Created: &types.AbsoluteTxPosition{BlockHeight: 1, TxIndex: 1},
-						},
+						ContractAddress: contractAddress(1, 1),
+						ContractInfo:    types.ContractInfoFixture(func(c *wasmTypes.ContractInfo) { c.CodeID = 1 }),
 					}, {
-						ContractAddress: addrFromUint64(2<<32 + 1),
-						ContractInfo: wasmTypes.ContractInfo{
-							CodeID:  1,
-							Creator: anyAddress,
-							Label:   "any",
-							Created: &types.AbsoluteTxPosition{BlockHeight: 1, TxIndex: 1},
-						},
+						ContractAddress: contractAddress(2, 1),
+						ContractInfo:    types.ContractInfoFixture(func(c *wasmTypes.ContractInfo) { c.CodeID = 1 }),
 					},
 				},
 			},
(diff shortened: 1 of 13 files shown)


x/wasm/internal/keeper/genesis_test.go
```diff
@@ -67,6 +68,166 @@ func TestGenesisExportImport(t *testing.T) {
...
+						CodeHash: codeHash[:],
+						Creator:  anyAddress,
+					},
+					CodesBytes: wasmCode,
+				}},
+				Contracts: nil,
+			},
+			expSuccess: true,
+		},
+		"prevent code hash mismatch": {src: types.GenesisState{
```
ethanfrey, line 94: :+1:

x/wasm/internal/keeper/genesis_test.go
```diff
@@ -67,6 +68,166 @@ func TestGenesisExportImport(t *testing.T) {
...
+				Codes: []types.Code{{
+					CodeInfo: wasmTypes.CodeInfo{
+						CodeHash: codeHash[:],
+						Creator:  anyAddress,
+					},
+					CodesBytes: wasmCode,
+				}},
+				Contracts: []types.Contract{
+					{
+						ContractAddress: addrFromUint64(1<<32 + 1),
```
ethanfrey, line 115: We should probably expose a function that does the bitshift - `contractAddress(codeId, contractCounter)` or something

x/wasm/internal/keeper/genesis_test.go
```diff
@@ -67,6 +68,166 @@ func TestGenesisExportImport(t *testing.T) {
...
+							CodeID:  1,
+							Creator: anyAddress,
+							Label:   "any",
+							Created: &types.AbsoluteTxPosition{BlockHeight: 1, TxIndex: 1},
+						},
+					},
+				},
+			},
+		},
+		"prevent duplicate contracts": {
```
ethanfrey, line 142: :+1:

ethanfrey: Also good to show we do allow two contracts with different contractAddresses to point to the same CodeID

x/wasm/internal/keeper/genesis_test.go
```diff
@@ -67,6 +68,166 @@ func TestGenesisExportImport(t *testing.T) {
...
+				},
+				CodesBytes: wasmCode,
+			}},
+			Contracts: nil,
+		}},
+		"happy path: code info and contract do match": {
+			src: types.GenesisState{
+				Codes: []types.Code{{
+					CodeInfo: wasmTypes.CodeInfo{
+						CodeHash: codeHash[:],
```
ethanfrey, line 108: One issue still here, is the CodeID is not explicit in import or export. We rely on the ordering of the array.

So, if one CodeInfo was deleted, all references are broken. It would be nice to make this explicit (using a `GenesisCodeInfo` type or whatever), and use the explicit ID not the index in the array

x/wasm/internal/keeper/keeper.go
```diff
@@ -496,12 +510,26 @@ func (k Keeper) peekAutoIncrementID(ctx sdk.Context, lastIDKey []byte) uint64 {
...
+	if store.Has(lastIDKey) {
+		panic(fmt.Sprintf("duplicate autoincrement id: %s", string(lastIDKey)))
+	}
 	bz := sdk.Uint64ToBigEndian(val)
 	store.Set(lastIDKey, bz)
 }
 
+func (k Keeper) importContract(ctx sdk.Context, address sdk.AccAddress, c *types.ContractInfo, state []types.Model) {
+	if !k.containsCodeInfo(ctx, c.CodeID) {
+		panic(fmt.Sprintf("unknown code id: %d", c.CodeID))
```
ethanfrey, line 524: Welcome to the dark side :dagger: 

I knew the sdk's "panic before error" would get you one day :wink:

x/wasm/internal/types/msg.go
```diff
@@ -2,35 +2,11 @@ package types
...
 import (
 	"encoding/json"
-	"net/url"
-	"regexp"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 )
 
-const (
```
ethanfrey, line 12: nice to move this to a separate validation section

x/wasm/internal/keeper/genesis_test.go
alpe: Ja, we should address this issue proper. I am afraid this is more work and should be priotized proper. I had planned a separate PR for it after some discussion. See "not covered" section.

ethanfrey: Okay, happy for a new issue on that. Can you make a proper issue for the "not covered" part and link it to this PR?

x/wasm/internal/keeper/keeper.go
alpe: Ja, haha 👿  and there was not even a reason to panic here. I knew the error could not be propagated beyond the sdk boundary but still no need to panic here. 💩  will fix it. Thanks

x/wasm/internal/keeper/genesis_test.go
alpe: Done #166

alpe:
😅  code coverage is positive again.

PR #168, opened 2020-07-01
Genesis code import not position agnostic

### LINKED ISSUE #166: Genesis: code import should be more resilient
## Summary

The position of a code_info within the code block matters for the codeID generation during import. 

## Problem Definition
Switching two elements can therefore lead to different codeIDs and may cause contracts to point to the wrong one. This should be more resilient to manual edits.

## Proposal

We should not rely on the order but have the codeID defined explicitly.

____

#### For Admin Use

- [ ] Not duplicate issue

alpe:
Resolves #166 

The codeID is stored within the genesis data so that the order in the file does not matter anymore.

______

For admin use:

- [ ] Reviewers assigned

### DIFF
--- a/x/wasm/internal/keeper/genesis.go
+++ b/x/wasm/internal/keeper/genesis.go
@@ -1,8 +1,6 @@
 package keeper
 
 import (
-	"bytes"
-
 	"github.com/CosmWasm/wasmd/x/wasm/internal/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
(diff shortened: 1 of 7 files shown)


x/wasm/internal/keeper/genesis.go
```diff
@@ -48,11 +56,15 @@ func ExportGenesis(ctx sdk.Context, keeper Keeper) types.GenesisState {
 
 	maxCodeID := keeper.GetNextCodeID(ctx)
 	for i := uint64(1); i < maxCodeID; i++ {
+		if !keeper.containsCodeInfo(ctx, i) {
```
ethanfrey, line 59: okay, see how this can be slow if there are *huge* gaps. But this is export and only if someone decides to play around at genesis time

PR #173, opened 2020-07-03
Gov proposal types and handler

### LINKED ISSUE #162: Create Governance Proposals for Contract Lifecycle
## Summary

We wish to allow the chain governance to "act as root" on all contracts in the system. It can execute any privileged operation on any contract. To allow this to happen, we must define some new governance proposals that can perform the required actions, before potentially restricting them to other users in follow-up PRs.

The decision to allow governance control should be optional and configured by each chain at compile time. 

## Problem Definition

See #161 for the larger discussion.

You should consider the hub governance as root.  If you do not want this, then disable those governance proposal types in the binary, so the governance oversight / backdoor (depending on your view) doesn't exist. This means that governance can migrate/burn a contract even if it is marked "immutable". We will add at least three new proposal "content" types. The basics would be UploadCodeProposal, InstantiateContractProposal, MigrateContractProposal. We would also likely want to add an UpdateAdministratorProposal, as hub governance may not want to migrate directly but appoint a new committee as administrator to some contested contract after the instantiation (or change the committee responsible).

## Proposal

Add 5 new proposal types that mirror the 5 messages (besides Execute) that exist in the `x/wasm` module:

These two act like normal messages, except they can never be prohibited (that will be added in future PRs):

* `StoreCodeProposal` == `MsgStoreCode`
* `InstantiateContractProposal` == `MsgInstantiateContract`

These act like the normal message "as root", meaning they always succeed, regardless of who is set as "Admin" on the contracts:

* `MigrateContractProposal` ~ `MsgMigrateContract` 
* `UpdateAdminProposal` ~ `MsgUpdateAdmin`
* `ClearAdminProposal` ~ `MsgClearAdmin`

~~When compiling a chain, you should be able to enable None of these, All of these, or Some of these. By default, none of the proposal types are enabled. There should be one function to easily register all proposals with the application. And there should be another function (or functions) that can be used to selectively enable some of these. The last 3 - `MigrateContract`, `UpdateAdmin`, `ClearAdmin` can all be enabled/disabled as a block, as giving access to governance to do one is close to giving access to all.~~ (moved as a separate task into #175 )
____

#### For Admin Use

- [ ] Not duplicate issue

alpe:
Resolves #162 

Introduces new gov proposal types:
* StoreCodeProposal
* InstantiateContractProposal
* MigrateContractProposal
* UpdateAdminProposal
* ClearAdminProposal

with proposal handler and authz policy.

______

For admin use:

- [ ] Reviewers assigned


x/wasm/internal/types/proposal.go
```diff
@@ -0,0 +1,263 @@
...
+  Description: %s
+  Changes:
+`, p.Title, p.Description))
+	// todo: print all data
+	return b.String()
+}
+
+type InstantiateContractProposal struct {
+	GovProposal
+	// Creator is the address that pays the init funds
```
ethanfrey, line 99: Interesting... is it possible to use the "community fund" as a creator address here?

But of course, if I want to create a contract and ask governance to allow it, *I* should be paying the funds, not the chain governance.

x/wasm/internal/types/proposal.go
```diff
@@ -0,0 +1,263 @@
...
+
+	if p.Code == 0 {
+		return sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, "code_id is required")
+	}
+
+	if err := validateLabel(p.Label); err != nil {
+		return err
+	}
+
+	if p.InitFunds.IsAnyNegative() {
```
ethanfrey, line 131: I think there is a general "ValidateBasic" on sdk.Coins. Any reason not to use that? (It also checks denom and such)

x/wasm/internal/types/proposal.go
```diff
@@ -0,0 +1,263 @@
...
+  Changes:
+`, p.Title, p.Description))
+	// todo: print all data
+	return b.String()
+}
+
+type ClearAdminContractProposal struct {
+	GovProposal
+
+	NewAdmin sdk.AccAddress `json:"new_admin" yaml:"new_admin"`
```
ethanfrey, line 234: I believe this field is not needed

x/wasm/internal/types/proposal.go
alpe: I did not know about this method. The code was copied from `MsgInstantiateContract`. I will change it there as well to `IsValid`
```go
	if !p.InitFunds.IsValid() {
		return sdkerrors.ErrInvalidCoins
	}
```

x/wasm/internal/types/proposal.go
alpe: 🤔  Maybe I miss something but there are no further constraints defined in the proposal so you can put any address there. Later in the wasm keeper the bank keeper is used and bypasses all authorization.
With the "community fund" address it problematic as the account's balance would become less than what is stored in the fund's table.
For our messages the problem does not exist as we require the creators signature

alpe: Update: there is a list of blockedAddr setup that should be protected. In the app it contains the `maccPerms` as `ModuleAccountAddrs`

Maybe we should prevent taking tokens from any of them. WDYT?
See also https://github.com/cosmos/cosmos-sdk/blob/master/x/bank/keeper/send.go#L274

ethanfrey: My question as mainly on how the SDK works (which maybe neither know), and when we send from a ModuleAccount vs sending from an AccAddress that represents the account.

The link is on master, maybe it is cleaner there.. I remember Send and SendFromModuleAccount as two separate calls somewhere.

Anyway, more a random question than a change request on the code

ethanfrey:
It looks good to me. What is missing to take this out of draft? (Besides increasing test coverage a bit)

x/wasm/internal/types/proposal.go
alpe: I have created https://github.com/CosmWasm/wasmd/issues/177 to follow up on this.

ethanfrey:
Will review now.... looking good :+1:

app/app.go
```diff
@@ -257,6 +251,15 @@ func NewWasmApp(
 	supportedFeatures := "staking"
 	app.wasmKeeper = wasm.NewKeeper(app.cdc, keys[wasm.StoreKey], app.accountKeeper, app.bankKeeper, app.stakingKeeper, wasmRouter, wasmDir, wasmConfig, supportedFeatures, nil, nil)
 
+	if len(wasm.DefaultEnabledProposals) != 0 {
```
ethanfrey, line 254: Nit: maybe add a comment above this that people can define a list of enabled proposals in their `app.go` and not need to use the list from `x/wasm`

(As we encourage people to import `x/wasm` and use `app.go` as an integration reference)

x/wasm/internal/keeper/authz_policy.go
```diff
@@ -0,0 +1,42 @@
...
+func (p DefaultAuthorizationPolicy) CanCreateCode(config types.AccessConfig, actor sdk.AccAddress) bool {
+	return config.Allowed(actor)
+}
+
+func (p DefaultAuthorizationPolicy) CanInstantiateContract(config types.AccessConfig, actor sdk.AccAddress) bool {
+	return config.Allowed(actor)
+}
+
+func (p DefaultAuthorizationPolicy) CanModifyContract(admin, actor sdk.AccAddress) bool {
+	return admin != nil && admin.Equals(actor)
```
ethanfrey, line 26: :+1:

x/wasm/internal/keeper/proposal_handler.go
```diff
@@ -0,0 +1,141 @@
...
+	AttributeSigner      = "signer"
+)
+
+// NewWasmProposalHandler creates a new governance Handler for wasm proposals
+func NewWasmProposalHandler(k Keeper, enabledTypes map[string]struct{}) govtypes.Handler {
+	return func(ctx sdk.Context, content govtypes.Content) error {
+		if content == nil {
+			return sdkerrors.Wrap(sdkerrors.ErrUnknownRequest, "content must not be empty")
+		}
+		if _, ok := enabledTypes[content.ProposalType()]; !ok {
```
ethanfrey, line 25: Good way to filter it here flexibly

x/wasm/internal/keeper/proposal_handler.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	if err := k.setContractAdmin(ctx, p.Contract, p.Sender, p.NewAdmin, GovAuthorizationPolicy{}); err != nil {
+		return err
+	}
+
+	ourEvent := sdk.NewEvent(
+		sdk.EventTypeMessage,
+		sdk.NewAttribute(sdk.AttributeKeyModule, types.ModuleName),
+		//sdk.NewAttribute(AttributeSigner, p.Creator.String()),
```
ethanfrey, line 118: There is nothing here for other governance contracts either?

I'm fine, as long as it looks like the other code they ship.

x/wasm/internal/keeper/test_common.go
```diff
@@ -164,12 +172,27 @@ func CreateTestInput(t *testing.T, isCheckTx bool, tempDir string, supportedFeat
 	// add wasm handler so we can loop-back (contracts calling contracts)
 	router.AddRoute(wasmTypes.RouterKey, TestHandler(keeper))
 
+	govRouter := gov.NewRouter().
```
ethanfrey, line 175: good to add this

x/wasm/internal/keeper/proposal_handler.go
alpe: 🤔 I have not found anything in the other handlers:
* [community pool](https://github.com/cosmos/cosmos-sdk/blob/v0.38.3/x/distribution/keeper/proposal_handler.go)
* [Params](https://github.com/cosmos/cosmos-sdk/blob/v0.38.3/x/params/proposal_handler.go)
* [Upgrade](https://github.com/cosmos/cosmos-sdk/blob/v0.38.3/x/upgrade/handler.go)

What would make sense for our use cases? If somebody wants to watch all contract modifications for example then we should create an event.... it was just the attribute name "signer" that does not fit the the proposal context

app/app.go
alpe: I would address this with https://github.com/CosmWasm/wasmd/issues/175 . Just added a note there

x/wasm/internal/keeper/proposal_handler.go
ethanfrey: Fair enough, I agree with your decisions.

Issue #161, opened 2020-06-29
Governance / Permissioning Tracking Issue


ethanfrey:
## Summary

This is a general tracking issue for all the work around governance control of smart contracts
to make a "permissioned" smart contract system possible.

## Problem Definition

The requirements are set forth in the [Cosmos Hub Proposal 25](https://ipfs.io/ipfs/QmbD3bMajQCFmtDmkuRVWhmMWVdN2sK8QP2FoFCz9cjPiC/Cosmos%20Hub%20Proposal_%20CosmWasm%284%29.pdf) which passed, and which we need to implement

Some additional context may be present in the [forum discussion](https://forum.cosmos.network/t/proposal-cosmwasm-on-cosmos-hub/3629)

## Proposal

Exact steps will be defining in smaller issues. This defines the architecture for the whole project. Issues:

- [x] Add Governance Proposals #162 
- [x] Control who can instantiate code #163
- [x]  Control who can upload code #164 

We will add three new governance proposal types to handle the contract lifetime and some more metadata to control the permissioning. Note this does not affect executing contracts, but rather the larger "privileged" lifecycle: upload, instantiate, migrate, which often restricted to an owner.

You should consider the hub governance as root. It can execute any privileged operation on any contract. If you do not want this, then disable those governance proposal types in the binary, so the governance oversight / backdoor (depending on your view) doesn't exist. This means that governance can migrate/burn a contract even if it is marked "immutable". We will add at least three new proposal "content" types. The basics would be `UploadCodeProposal`, `InstantiateContractProposal`, `MigrateContractProposal`. We would also likely want to add an `UpdateAdministratorProposal`, as hub governance may not want to migrate directly but appoint a new committee as administrator to some contested contract after the instantiation (or change the committee responsible).

Beyond the root control, we also want to allow delegating to a subcommittee for quicker oversight. This ends up with upload and instantiate having a permission of Nobody | Somebody(Address) | Anybody and migrate of Nobody | Somebody(Address). The permissions for upload can be stored in module-level Params, which the permission level for instantiate can be stored on the CodeInfo. The Params should list the permission required for uploading and the default permission for instantiating (likely Anybody or Nobody depending on the chain). The CodeInfo permission level can also be set when uploading (overriding Params default).

[...]

PR #176, opened 2020-07-06
SPIKE: Gov perm spike

alpe:
🚨 Design spike

See for context  #161 

🚧  Not to be merged 🚧 

### New types
- New emum type `AccessType` : Undefined, Nobody, OnlyAddress, Everybody
- New type for `AccessType` and address pair: `AccessConfig`
- New `AuthorizationPolicy` with default constraints and Gov implementation

### Defaults and stored data
- Module params with: upload `AccessConfig` and instantiate default AccessType
- CodeInfo stores `InstantiateConfig AccessConfig` so that we could restrict to `owner` address
- ContractInfo stores `ModifyConfig AccessConfig ` that replaces the `admin` field

### Messages
Not touched, yet but we could customize the AccessTypes/ config for upload, instantiate easily

______

For admin use:

- [ ] Reviewers assigned


x/wasm/internal/types/types.go
```diff
@@ -53,20 +54,22 @@ func (c CodeInfo) ValidateBasic() error {
...
 }
 
 // ContractInfo stores a WASM contract instance
 type ContractInfo struct {
-	CodeID  uint64          `json:"code_id"`
-	Creator sdk.AccAddress  `json:"creator"`
-	Admin   sdk.AccAddress  `json:"admin,omitempty"`
+	CodeID       uint64         `json:"code_id"`
+	Creator      sdk.AccAddress `json:"creator"`
+	ModifyConfig AccessConfig   `json:"modify_config"`
```
ethanfrey, line 71: Okay, I don't see why we would ever want "Everybody" here, which why you broaden it. So this is not incorrect, but maybe a bit more verbose API than desired.

x/wasm/internal/types/params.go
```diff
@@ -0,0 +1,107 @@
...
+			panic(err)
+		}
+		return AccessConfig{Type: OnlyAddress, Address: addr}
+	case Everybody:
+		return AllowEverybody
+	}
+	panic("unsupported access type")
+}
+
+type AccessConfig struct {
```
ethanfrey, line 43: Okay, this makes sense for the enum

x/wasm/internal/keeper/authorization_policy.go
```diff
@@ -0,0 +1,42 @@
...
+
+func (p DefaultAuthorizationPolicy) CanCreateCode(config types.AccessConfig, actor sdk.AccAddress) bool {
+	return config.Allowed(actor)
+}
+
+func (p DefaultAuthorizationPolicy) CanInstantiateContract(config types.AccessConfig, actor sdk.AccAddress) bool {
+	return config.Allowed(actor)
+}
+
+func (p DefaultAuthorizationPolicy) CanModifyContract(config types.AccessConfig, actor sdk.AccAddress) bool {
```
ethanfrey, line 25: Since we have 3 calls here, maybe we can just use a different sig for `CanModifyContract` to behave with `admin = nil | Address` as now. And the create/instantiate using the new AccessConfig you created.

x/wasm/internal/keeper/keeper.go
```diff
@@ -72,14 +74,34 @@ func NewKeeper(cdc *codec.Codec, storeKey sdk.StoreKey, accountKeeper auth.Accou
...
+		authZPolicy:   DefaultAuthorizationPolicy{},
+		paramSpace:    ps,
 	}
 	keeper.queryPlugins = DefaultQueryPlugins(bankKeeper, stakingKeeper, keeper).Merge(customPlugins)
 	return keeper
 }
 
+func (k Keeper) getUploadAccessConfig(ctx sdk.Context) types.AccessConfig {
+	var a types.AccessConfig
+	k.paramSpace.Get(ctx, types.ParamStoreKeyUploadAccess, &a)
```
ethanfrey, line 86: Very nice to use params for this.

x/wasm/internal/keeper/keeper.go
```diff
@@ -72,14 +74,34 @@ func NewKeeper(cdc *codec.Codec, storeKey sdk.StoreKey, accountKeeper auth.Accou
...
+	return a
+}
+
 // Create uploads and compiles a WASM contract, returning a short identifier for the contract
 func (k Keeper) Create(ctx sdk.Context, creator sdk.AccAddress, wasmCode []byte, source string, builder string) (codeID uint64, err error) {
+	return k.create(ctx, creator, wasmCode, source, builder, k.authZPolicy)
+}
+
+func (k Keeper) create(ctx sdk.Context, creator sdk.AccAddress, wasmCode []byte, source string, builder string, authZ AuthorizationPolicy) (codeID uint64, err error) {
+	if !authZ.CanCreateCode(k.getUploadAccessConfig(ctx), creator) {
```
ethanfrey, line 102: Very nice here.

I would change the message from "no permission" to "cannot create" or such, as ErrUnauthorized default message is the same as "no permission" basically

x/wasm/internal/keeper/keeper.go
```diff
@@ -93,7 +115,7 @@ func (k Keeper) Create(ctx sdk.Context, creator sdk.AccAddress, wasmCode []byte,
 	}
 	store := ctx.KVStore(k.storeKey)
 	codeID = k.autoIncrementID(ctx, types.KeyLastCodeID)
-	codeInfo := types.NewCodeInfo(codeHash, creator, source, builder)
+	codeInfo := types.NewCodeInfo(codeHash, creator, source, builder, k.getInstantiateAccessConfig(ctx))
```
ethanfrey, line 118: Okay, this is a design question.

Do we allow the contract creator to set who can instantiate? Or do we use a global default?

If we have a global default with "Somebody", this should likley not be a hardcoded "Somebody", but rather the actual creator.

I would propose to add an optional `instantiatePermission` field to the StoreCode method. If not set, then use `k.getInstantiateAccessConfig(ctx)`.

If governance is required to upload code, then it can also check that this `instantiatePermssion` is acceptable (or unset)

x/wasm/internal/keeper/keeper.go
```diff
@@ -189,7 +218,12 @@ func (k Keeper) Instantiate(ctx sdk.Context, codeID uint64, creator, admin sdk.A
 
 	// persist instance
 	createdAt := types.NewCreatedAt(ctx)
-	instance := types.NewContractInfo(codeID, creator, admin, initMsg, label, createdAt)
+
+	modifyConfig := types.AllowNobody // TODO: How about default to creator?
```
ethanfrey, line 222: Default should be immutable. We only make "upgradeable contracts" when explicitly requested. This is an extra feature

ethanfrey: That said, with my previous comments, we can likely store the admin on contractInfo as now, and just update the store and instantiate permisisons as you already did

x/wasm/internal/types/types.go
```diff
@@ -30,10 +30,11 @@ func (m Model) ValidateBasic() error {
...
 type CodeInfo struct {
-	CodeHash []byte         `json:"code_hash"`
-	Creator  sdk.AccAddress `json:"creator"`
-	Source   string         `json:"source"`
-	Builder  string         `json:"builder"`
+	CodeHash          []byte         `json:"code_hash"`
+	Creator           sdk.AccAddress `json:"creator"`
+	Source            string         `json:"source"`
+	Builder           string         `json:"builder"`
+	InstantiateConfig AccessConfig   `json:"instantiate_config"`
```
ethanfrey, line 37: :+1:

x/wasm/internal/types/types.go
alpe: I don't see a use case for "Everybody" here other than somebody intentionally wanting to loose control over a contract.

alpe: If we ever want AccessConfig to contain a list of addresses or do other extensions in this field the re usage of AccessConfig makes sense. But I see your point

x/wasm/internal/keeper/authorization_policy.go
alpe: Actually I had this before before I unified to `AccessConfig`.

x/wasm/internal/keeper/keeper.go
alpe: IMHO it would make sense to let the contract creator decide on the permission setting. It is not required for Gov but can make sense to speed up the next step.
I kept message modifications out of scope in this spike but will push a another commit to be aligned.

x/wasm/internal/types/types.go
ethanfrey: >  If we ever want AccessConfig to contain a list of addresses or do other extensions in this field the re usage of AccessConfig makes sense. But I see your point

Okay. If the option was `Nobody | Somebody(Addr) | Somebodies(Addr[]) | Everybody` then I see your point that "all options - Everybody" != `Nobody | Somebody`.

ethanfrey: I would like to keep things simplest possible given the functionality. That means either:

* Making this one `Admin` field like before (which satisfies the current functionality) OR
* Extend the `AccessConfig` with another option now that make this required (please think a bit on what it should be)

ethanfrey: > I don't see a use case for "Everybody" here other than somebody intentionally wanting to loose control over a contract.

This is just chaos. And much more likely to be a typo/misconfiguration than an honest request

alpe: Let's keep things simple with the admin field. The changes are minimal on the server side and whenever we want to support multiple addresses we can do it easily. 👌

ethanfrey:
Okay, looks good to me except the merge conflicts. Want to take this out of draft, so I can approve?

Actually, maybe it makes sense to merge #173 to master first, then rebase this and merge to master - as those are separate issues.

ethanfrey:
This is now in #173 right? I saw very similar code

If done, please close it.

alpe:
Closing this PR now as work is now on the proper implementation. I will keep the branch for reference until the tasks are finished.

PR #180, opened 2020-07-08
Add configurable control for code instantiation

### LINKED ISSUE #163: Control who can instantiate Code
## Summary

As part of the governance meta-issue #161, we wish to have the ability to control *who* can instantiate a given `Code` once it is uploaded. Besides a governance proposal, this may be `Nobody | Somebody(Address) | Everybody`. ~~The current default setting is `Everybody`.~~ Update: The default setting of who can instantiate a CodeID is `Somebody(creator)`

## Problem Definition

This only really makes sense once #164 is done and you can control who can upload code in the first place, but this seemed simpler, so I placed it first. In a permissioned setting, you may have an cw20 contract approved, but not necessarily want to allow everyone to make tokens, so you could need a second permissioning for that. You could set it to `Nobody` to only allow chain governance approve instantiation, or you could use `Somebody(Address)` to give permission to some multisig/committee to approve instantiation much quicker than requiring full governance. `Everybody` remains the default for permissionless chains. (In fact, if there is no control over upload, there is no point to set it to anything other than `Everybody`. Someone could download the wasm code from the chain and upload it as a new CodeID with full permissions).

## Proposal

* Determine the best way to handle this enum (nil / non-nil doesn't handle the "everybody" case, so this may need to be explicit) 
* Add another field to `MsgStoreCode` and `StoreCodeProposal` to control who can instantiate contracts from this CodeID.
* Store this info in `CodeID` and expose it in the query endpoints
* Enforce this restriction when processing `MsgInstantiateContract` (but not on `InstantiateContractProposal`)

____

#### For Admin Use

- [ ] Not duplicate issue

alpe:
Resolves #163

______

For admin use:

- [ ] Reviewers assigned

### DIFF
--- a/x/wasm/internal/keeper/keeper_test.go
+++ b/x/wasm/internal/keeper/keeper_test.go
@@ -53,6 +53,59 @@ func TestCreate(t *testing.T) {
 	require.Equal(t, wasmCode, storedCode)
 }
 
+func TestCreateStoresInstantiatePermission(t *testing.T) {
+	wasmCode, err := ioutil.ReadFile("./testdata/contract.wasm")
+	require.NoError(t, err)
+	var (
+		deposit = sdk.NewCoins(sdk.NewInt64Coin("denom", 100000))
+		myAddr  = bytes.Repeat([]byte{1}, sdk.AddrLen)
+	)
+
+	specs := map[string]struct {
+		srcPermission types.AccessType
+		expInstConf   types.AccessConfig
+	}{
+		"default": {
+			srcPermission: types.DefaultParams().DefaultInstantiatePermission,
+			expInstConf:   types.AllowEverybody,
+		},
+		"everybody": {
+			srcPermission: types.Everybody,
+			expInstConf:   types.AllowEverybody,
+		},
+		"nobody": {
+			srcPermission: types.Nobody,
+			expInstConf:   types.AllowNobody,
+		},
+		"onlyAddress with matching address": {
+			srcPermission: types.OnlyAddress,
+			expInstConf:   types.AccessConfig{Type: types.OnlyAddress, Address: myAddr},
+		},
+	}
+	for msg, spec := range specs {
+		t.Run(msg, func(t *testing.T) {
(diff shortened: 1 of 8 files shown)


x/wasm/internal/keeper/keeper_test.go
```diff
@@ -53,6 +53,59 @@ func TestCreate(t *testing.T) {
...
+	var (
+		deposit = sdk.NewCoins(sdk.NewInt64Coin("denom", 100000))
+		myAddr  = bytes.Repeat([]byte{1}, sdk.AddrLen)
+	)
+
+	specs := map[string]struct {
+		srcPermission types.AccessType
+		expInstConf   types.AccessConfig
+	}{
+		"default": {
```
ethanfrey, line 68: Nice cases here!

x/wasm/internal/keeper/keeper_test.go
```diff
@@ -263,6 +316,70 @@ func TestInstantiate(t *testing.T) {
...
+	)
+
+	initMsg := InitMsg{
+		Verifier:    anyAddr,
+		Beneficiary: anyAddr,
+	}
+	initMsgBz, err := json.Marshal(initMsg)
+	require.NoError(t, err)
+
+	specs := map[string]struct {
```
ethanfrey, line 337: Covers everything I can think of. And well.

x/wasm/internal/keeper/proposal_integration_test.go
```diff
@@ -64,6 +64,7 @@ func TestInstantiateProposal(t *testing.T) {
 
 	ctx, keepers := CreateTestInput(t, false, tempDir, "staking", nil, nil)
 	govKeeper, wasmKeeper := keepers.GovKeeper, keepers.WasmKeeper
+	wasmKeeper.setParams(ctx, types.Params{UploadAccess: types.AllowNobody, DefaultInstantiatePermission: types.Nobody})
```
ethanfrey, line 67: Nice. Showing governance can always succeed regardless of the permissions

PR #188, opened 2020-07-14
Remove max gas

ethanfrey:
Closes #170 

- Targeted PR against correct branch (see [CONTRIBUTING.md](https://github.com/cosmos/gaia/blob/master/CONTRIBUTING.md#pr-targeting))

- [ ] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Wrote tests
- [ ] Updated relevant documentation (`docs/`)
- [ ] Added  relevant `godoc` [comments](https://blog.golang.org/godoc-documenting-go-code).
- [ ] Added a relevant changelog entry to the `Unreleased` section in `CHANGELOG.md`
- [ ] Re-reviewed `Files changed` in the Github PR explorer

______

For admin use:

- [ ] Added appropriate labels to PR (ex. `WIP`, `R4R`, `docs`, etc)
- [ ] Reviewers assigned
- [ ] Squashed all commits, uses message "Merge pull request #XYZ: [title]" ([coding standards](https://github.com/tendermint/coding/blob/master/README.md#merging-a-pr))

PR #193, opened 2020-07-15
Increase and document the MaxGas limit

### LINKED ISSUE #170: Remove keeper.MaxGas or write a comment why it is needed
The existence of [keeper.MaxGas](https://github.com/CosmWasm/wasmd/blob/v0.9.1/x/wasm/internal/keeper/keeper.go#L30-L31) is suspicious and is confusing people. @ethanfrey, why do we need a limit for the smart contract execution other than the gas limit of the block or the tx?

Even if the VM has an execution limit (which is [has right now](https://github.com/CosmWasm/cosmwasm/blob/v0.9.2/packages/vm/src/backends/singlepass.rs#L15-L23) but should not), I don't see a reason to maintain the value in two places. The VM then just stops once its limit is eached.

ethanfrey:
Closes #170 
Replaces #188 (which shows why we cannot just remove the field)

We document where the limit comes from and link to the rust code.

- [ ] Wrote tests

______

For admin use:

- [ ] Reviewers assigned

### DIFF
--- a/x/wasm/internal/keeper/keeper.go
+++ b/x/wasm/internal/keeper/keeper.go
@@ -28,8 +28,10 @@ import (
 // Rough timing have 88k gas at 90us, which is equal to 1k sdk gas... (one read)
 const GasMultiplier = 100
 
-// MaxGas for a contract is 900 million (enforced in rust)
-const MaxGas = 900_000_000
+// MaxGas for a contract is 10 billion wasmer gas (enforced in rust to prevent overflow)
+// The limit for v0.9.3 is defined here: https://github.com/CosmWasm/cosmwasm/blob/v0.9.3/packages/vm/src/backends/singlepass.rs#L15-L23
+// (this will be increased in future releases)
+const MaxGas = 10_000_000_000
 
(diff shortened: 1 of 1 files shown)


ethanfrey:
> Isn't there any way to move this into the testing framework?
> I don't think we should silently change the value for the user if unconfigured.

I don't understand.

In any production code, there will be a gas limit set and it will be below this.
If we use `InfiniteGasMeter` this will be an issue.

We cannot "patch the testing framework", as there is nothing to patch there. They setup a test context and call the keeper methods. I think any other change would be more invasive.

This was working for quite some time, I just added docs and increased the value (in 0.9->0.10 which allows breaking changes)

webmaster128:
> I don't understand.
> 
> In any production code, there will be a gas limit set and it will be below this.
> If we use `InfiniteGasMeter` this will be an issue.

The existence of `MaxGas` adds logic to production code, which should not be there for any production use. The variable and its usage is only needed because the testing framework needs it. So the question was if there is a way to get this out of production code. But given your reply this does not seem possible.

> I think any other change would be more invasive.

Yeah, if you don't see a way to move this into testing code then we have to live with it.

PR #195, opened 2020-07-16
Full history of contract code migrations 

### LINKED ISSUE #130: Full history of contract code migrations
## Summary
Contracts can be migrated to different code versions. A full history of the changes would be helpful for audits and debugging issues.

## Problem Definition
Follow up from [comment](https://github.com/CosmWasm/wasmd/issues/119#issuecomment-638729015)

## Proposal
Mandatory data would be `codeID` and `AbsoluteTxPosition` for each migration
This data should be available via rest/ cli as an ordered list.

____

#### For Admin Use

- [ ] Not duplicate issue

alpe:
Resolves #130 

A contract history is stored within ContractInfo for:
* Init
* Migrate
* Genesis import

Note: The genesis import/ export was more complex than storing some history data within the entity.

Moving the history out of the ContractInfo makes more sense to me to optimize the workflow and save gas. I will add a ticket to have this as a second step.
______

For admin use:

- [ ] Reviewers assigned

### DIFF
--- a/x/wasm/internal/keeper/genesis_test.go
+++ b/x/wasm/internal/keeper/genesis_test.go
@@ -104,9 +104,8 @@ func TestGenesisExportImport(t *testing.T) {
 func TestFailFastImport(t *testing.T) {
 	wasmCode, err := ioutil.ReadFile("./testdata/contract.wasm")
 	require.NoError(t, err)
-	codeHash := sha256.Sum256(wasmCode)
-	anyAddress := make([]byte, 20)
 
+	myCodeInfo := wasmTypes.CodeInfoFixture(wasmTypes.WithSHA256CodeHash(wasmCode))
 	specs := map[string]struct {
 		src        types.GenesisState
 		expSuccess bool
@@ -114,11 +113,8 @@ func TestFailFastImport(t *testing.T) {
 		"happy path: code info correct": {
 			src: types.GenesisState{
 				Codes: []types.Code{{
-					CodeID: 1,
-					CodeInfo: wasmTypes.CodeInfo{
-						CodeHash: codeHash[:],
-						Creator:  anyAddress,
-					},
+					CodeID:     1,
+					CodeInfo:   myCodeInfo,
 					CodesBytes: wasmCode,
 				}},
 				Sequences: []types.Sequence{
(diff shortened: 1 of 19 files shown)


x/wasm/internal/keeper/genesis_test.go
```diff
@@ -377,6 +391,116 @@ func TestFailFastImport(t *testing.T) {
...
+  "codes": [
+    {
+      "code_id": 1,
+      "code_info": {
+        "code_hash": "mrFpzGE5s+Qfn9Xe7OikXc0q169m7sMm4ZuV6pVA8Mc=",
+        "creator": "cosmos1qtu5n0cnhfkjj6l2rq97hmky9fd89gwca9yarx",
+        "source": "https://example.com",
+        "builder": "foo/bar:tag",
+        "instantiate_config": {
+          "type": 1,
```
ethanfrey, line 459: I thought you updated this to "Nobody" / "Everybody" instead of numbers? Or is that PR not merged?

x/wasm/internal/keeper/proposal_integration_test.go
```diff
@@ -169,9 +176,20 @@ func TestMigrateProposal(t *testing.T) {
 	cInfo := wasmKeeper.GetContractInfo(ctx, contractAddr)
 	require.NotNil(t, cInfo)
 	assert.Equal(t, uint64(2), cInfo.CodeID)
-	assert.Equal(t, uint64(1), cInfo.PreviousCodeID)
 	assert.Equal(t, anyAddress, cInfo.Admin)
 	assert.Equal(t, "testing", cInfo.Label)
+	expHistory := []types.ContractCodeHistoryEntry{{
+		Operation: types.GenesisContractCodeHistoryType,
```
ethanfrey, line 182: Good, you don't store any data from the genesis one here (can be huge) - but using a different type than init shows where it came from (and where the data is if you want to query it) :+1:

x/wasm/internal/keeper/querier.go
```diff
@@ -85,6 +81,15 @@ func queryContractInfo(ctx sdk.Context, bech string, req abci.RequestQuery, keep
 	return bz, nil
 }
 
+// redact clears all fields not in the public api
+func redact(info *types.ContractInfo) {
```
ethanfrey, line 85: :+1:

x/wasm/internal/keeper/genesis_test.go
alpe: yes, this is an old configuration I had before. Good 👁️ . It also shows that the field is not validated proper I will fix this

alpe:
Pushed some additional tests. They are not really related to this feature but red areas on the critical path.... should make codecov happy 😆

x/wasm/internal/types/params.go
```diff
@@ -147,7 +147,7 @@ func validateAccessType(i interface{}) error {
 
 func (v AccessConfig) ValidateBasic() error {
 	switch v.Type {
-	case Undefined:
+	case Undefined, "":
```
ethanfrey, line 150: Good catch

PR #212, opened 2020-07-21
A test that triggers null pointer exception on 3rd level of external query

assafmo:
Also need to add a print in this line: https://github.com/CosmWasm/go-cosmwasm/blob/master/api/callbacks.go#L415
```go
fmt.Printf("\n\n\ncQueryExternal: gasAfter: %d gasBefore: %d querier: %v\n\n\n", gasAfter,gasBefore, querier)
```

And then you'd see that `gasAfter == gasBefore` and that in the 3rd level of query that the vTable at the end of `querier` is `{<nil> <nil> <nil> <nil>}`, which leads to null pointer exception later on in `keeper.go/QuerySmart`.


assafmo:
Fixed by #213

ethanfrey:
I will review this. I think it would be helpful to add this (or a modified version) to wasmd to ensure we have no regressions. As the fix had no test with it.

Essentially don't delete the branch, I will bring it into master with possibly a different contract

assafmo:
Okay, for sure.
Just note that the original test was intended to test that the SGX OOM killer handles infinite external query loops gracefully, so you might want to create a different test that e.g. passes a number+1 to the next query and when it reaches 10 or 20 just return, and make sure that the first caller receives 10 or 20.

```rust
#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
#[serde(rename_all = "snake_case")]
pub enum QueryMsg {
    SendExternalQueryInfiniteLoop { to: HumanAddr, counter: u64 },
}

pub fn query<S: Storage, A: Api, Q: Querier>(deps: &Extern<S, A, Q>, msg: QueryMsg) -> QueryResult {
    match msg {
        QueryMsg::SendExternalQueryInfiniteLoop { to, counter } => {
            send_external_query_infinite_loop(deps, to, counter)
        }
    }
}

fn send_external_query_infinite_loop<S: Storage, A: Api, Q: Querier>(
    deps: &Extern<S, A, Q>,
    contract_addr: HumanAddr,
    counter: u64,
) -> QueryResult {
    if counter >= 10 {
        return Ok(Binary(counter.into()))
    }

    let answer = deps
        .querier
        .query::<u64>(&QueryRequest::Wasm(WasmQuery::Smart {
            contract_addr: contract_addr.clone(),
            msg: Binary(
                format!(
                    r#"{{"send_external_query_infinite_loop":{{"to":"{}","counter":{}}}}}"#,
                    contract_addr.clone().to_string(),
                    counter + 1,
                )
                .into(),
            ),
        }));

    match answer {
        Ok(counter) => Ok(Binary(counter.into())),
        Err(e) => Err(e),
    }
}
```

assafmo:
And in go code:
```go
counter, err = keeper.QuerySmart(ctx, contractAddress, []byte(fmt.Sprintf(`{"send_external_query_infinite_loop":{"to":"%s","counter":1}}`, contractAddress.String())))
require.Equal(t, 10, counter)
```
(Probably need to convert counter from bytes to uint64)

ethanfrey:
Thanks, I will add such to the hackatom contract

ethanfrey:
Thanks, I will add such to the hackatom contract

PR #214, opened 2020-07-22
Write a doc on how x/wasm uses IBC

ethanfrey:
This is a proposal of how we use ports and channels.

It is not fully complete and could use feedback. Let's define this well for our best ideas now and adapt
it if the code shows another approach is more advantageous.

Related to https://github.com/CosmWasm/cosmwasm/issues/147 which has a long discussion between Chris and I
in February, also setting forth the version negotiation that has since been implemented in IBC. Please read that for
some of the thinking - that is not final thinking, nor is this document final, but the should cover a lot of the reasoning
behind the design.

- [ ] Wrote tests

______

For admin use:

- [ ] Reviewers assigned

### DIFF
--- a/x/wasm/IBC.md
+++ b/x/wasm/IBC.md
@@ -36,7 +36,7 @@ as how contracts can properly identify their counterparty.
   (See [Channel Handshake Version Negotiation](https://docs.cosmos.network/master/ibc/custom.html#channel-handshake-version-negotiation))
 * Both the *Port* and the *Channel* are fully owned by one contract.
 * `x/wasm` will allow both *ORDERED* and *UNORDERED* channel and pass that mode
-  down to the contract in `OnTryOpenChannel`, so the contract can decide if it accepts
+  down to the contract in `OnChanOpenTry`, so the contract can decide if it accepts
   the mode. We will recommend the contract developers stick with *ORDERED* channels
   unless they can reason about async packet timing.
 * When sending a packet, the CosmWasm contract must specify the *ChannelID*
@@ -79,13 +79,18 @@ Packet Lifecycle:
   an IBCEnabled contract, this will be routed to the proper contract
   and call an exposed function for this purpose.
 * `IBCPacketAck` - The original sender of `IBCSendMsg` will
-  get this callback eventually if the message was successfully
-  processed on the other chain (this implies using an envelope for the 
-  IBC acknowledge message, so the Go code can differentiate between success and error)
-* `IBCPacketFailed` - The original sender of `IBCSendMsg` will 
+  get this callback eventually if the message was
+  processed on the other chain (this may be either a success or an error,
+  but comes from the app-level protocol, not the IBC protocol).
+* `IBCPacketDropped` - The original sender of `IBCSendMsg` will 
   get this callback eventually if the message failed to be
-  processed on the other chain (for timeout, closed channel, or the 
-  other side returning an error message in the IBC acknowledgement)
+  processed on the other chain (for timeout, closed channel, or 
+  other IBC-level failure)
+  
+Note: We may add some helpers inside the contract to map `IBCPacketAck` / `IBCPacketDropped`
+to `IBCPacketSucceeded` / `IBCPacketFailed` assuming they use the standard envelope. However,
+we decided not to enforce this on the Go-level, to allow contracts to communicate using protocols
+that do not use this envelope.
 
 Channel Lifecycle Hooks:
 
--- a/x/wasm/IBC.md
+++ b/x/wasm/IBC.md
@@ -17,41 +17,75 @@ and *Channel*
 
 Establishing *Clients* and *Connections* is out of the scope of this
 module and must be created by the same means as for `ibc-transfer`
-(via the cli or otherwise). `x/wasm` will bind one or more *Ports* to
-handle messages for the contracts. How *Channels* are established
-is an open question for discussion.
+(via the cli or otherwise). `x/wasm` will bind a unique *Port* for each
+"IBC Enabled" contract.
 
-The following actions are only available for `IBCEnabled` contracts:
+An "IBC Enabled" contract may dispatch the following messages not available
+to other contracts:
 
-* Dispatch `IBCSendMsg` - this sends an IBC packet over an established
-  channel. This is returned from `handle` just like any other `CosmosMsg`
-  (For mocks, we will trigger this externally, later only
-  valid contract addresses should be able to do so).
-* Handle `IBCRecvPacket` - when another chain sends a packet destined to
+* `IBCSendMsg` - this sends an IBC packet over an established channel. 
+* `IBCOpenChannel` - given a ConnectionID and a remote Port, this will request
+  to open a new channel with the given chain
+* `IBCCloseChannel` - given an existing channelID bound to this contract's Port,
+  initiate the closing sequence and reject all pending packets.
+
+They are returned from `handle` just like any other `CosmosMsg`
+(For mocks, we will trigger this externally, later only valid contract addresses 
+should be able to do so).
+
+An "IBC Enabled" contract must support the following callbacks from the runtime
+(we will likely multiplex many over one wasm export, as we do with handle, but these 
+are the different calls we must support):
+
+Packet Lifecycle:
+
+* `IBCRecvPacket` - when another chain sends a packet destined to
   an IBCEnabled contract, this will be routed to the proper contract
   and call an exposed function for this purpose.
-* Handle `IBCAckPacket` - The original sender of `IBCSendMsg` will
+* `IBCPacketAck` - The original sender of `IBCSendMsg` will
   get this callback eventually if the message was successfully
-  processed on the other chain
-* Handle `IBCErrorPacket` - The original sender of `IBCSendMsg` will 
+  processed on the other chain (this implies using an envelope for the 
+  IBC acknowledge message, so the Go code can differentiate between success and error)
+* `IBCPacketFailed` - The original sender of `IBCSendMsg` will 
   get this callback eventually if the message failed to be
-  processed on the other chain (for any reason)
+  processed on the other chain (for timeout, closed channel, or the 
+  other side returning an error message in the IBC acknowledgement)
+
+Channel Lifecycle Hooks:
+
+* `OnChanOpenTry` - this is called when another chain attempts to open
+  a channel with this contract. It provides the protocol versions that
+  the initiate expects, and the contract can reject it or accept it
+  and return the protocol version it will communicate with.
+* `OnChanOpenConfirm` - this is called when the other party has agreed
+  to the channel and all negotiation is over. This will be called exactly
+  once for any established channel, and the contract should record these
+  open channels in the internal state.
+* `OnChanClosed` - this is called when an existing channel is closed
+  for any reason, allowing the client to update the state there.
+  This will (likely? @cwgoes?) be accompanied by `IBCPacketFailed`
+  callbacks for all the in-progress packets that were not processed before
+  it closed, as well as all pending acks.
+
+We may want to expose some basic queries of IBC State to the contract.
+We should check if this is useful to the contract and if it opens up
+any possible DoS:
+
+* `ListMyChannels` - return a list of all channel IDs (along with remote PortID)
+  that are bound to this contract's Port.
+* `ListPendingPackets` - given a channelID owned by the contract, return all packets
+  in that channel, that were sent by this chain, for which no acknowledgement or timeout
+  has been received
 
 For mocks, all the `IBCxxxPacket` commands are routed to some
 Golang stub handler, but containing the contract address, so we
 can perform contract-specific actions for each packet.
 
-There are open questions on how exactly the various `IBCxxPacket`
-messages are routed to a contract, as well as related to the
-channel lifecycle. (Can a contract open a channel? How do we
-allow contracts to participate in protocol version negotiation
-when establishing a channel?)
-
 ## Ports and Channels
 
-There are two main proposals on how to map these concepts to
-contracts. We will describe both in detail, with arguments
-pro and contra, to select one approach.
+We decided on "one port per contract", especially after the IBC team raised
+the max length on port names to allow `wasm-<bech32 address>` to be a valid port.
+Here are the arguments for "one port for x/wasm" vs "one port per contract"
 
 ### One Port per Contract
 
@@ -70,10 +104,10 @@ as how contracts can properly identify their counterparty.
   [`ChanOpenTry` and `ChanOpenAck` phases](https://docs.cosmos.network/master/ibc/overview.html#channels).
   (See [Channel Handshake Version Negotiation](https://docs.cosmos.network/master/ibc/custom.html#channel-handshake-version-negotiation))
 * Both the *Port* and the *Channel* are fully owned by one contract.
-* `x/wasm` only accepts *ORDERED Channels* for simplicity of contract
-  correctness.
+* Question: `x/wasm` only accepts *ORDERED Channels* for simplicity of contract
+  correctness? Or should we allow the contract to decide?
 * When sending a packet, the CosmWasm contract must specify the *ChannelID*
-  (and remote *PortID*?), which is generally provided by the external client
+  and remote *PortID*, which is generally provided by the external client
   which understands which channels it wishes to communicate over.
 * When sending a packet, the contract can set a custom identifier (of
   max 64 characters) that it can use to look this up later.
@@ -81,8 +115,7 @@ as how contracts can properly identify their counterparty.
   *ChannelID* as well as remote *PortID* (and *ClientID*???) that is
   communicating.
 * When receiving an Ack or Error packet, the contract also receives the
-  same identifier that it set on Send (`x/wasm` handles mapping between this
-  and the native IBC packet IDs).
+  original packet that it set on Send 
   
 ### One Port per Module
 
--- a/x/wasm/IBC.md
+++ b/x/wasm/IBC.md
@@ -43,12 +43,10 @@ as how contracts can properly identify their counterparty.
   and remote *PortID*, which is generally provided by the external client
   which understands which channels it wishes to communicate over.
 * When receiving a Packet (or Ack or Error), the contracts receives the
-  *ChannelID* as well as remote *PortID* (and *ClientID*???) that is
-  communicating.
+  *ChannelID* as well as remote *PortID* with which it is communicating.
 * When receiving an Ack or Error packet, the contract also receives the
   original packet that it set on Send 
-  
-  
+
 ## Workflow
 
 Establishing *Clients* and *Connections* is out of the scope of this
--- a/x/wasm/IBC.md
+++ b/x/wasm/IBC.md
@@ -1,0 +1,128 @@
+# IBC specification
+
+This documents how CosmWasm contracts are expected to interact with IBC.
+
+## General Concepts
+
+**IBC Enabled** - when instantiating a contract, we detect if it supports IBC messages
+  (this is an optional extension). If so, it is considered "IBC Enabled".
+  This info should be public on all contracts. (For mock, we assume all
+  contracts are IBC enabled)
+  
+Also, please read the [IBC Docs](https://docs.cosmos.network/master/ibc/overview.html)
+for detailed descriptions of the terms *Port*, *Client*, *Connection*,
+and *Channel*
+  
+## Workflow
+
+Establishing *Clients* and *Connections* is out of the scope of this
+module and must be created by the same means as for `ibc-transfer`
+(via the cli or otherwise). `x/wasm` will bind one or more *Ports* to
+handle messages for the contracts. How *Channels* are established
+is an open question for discussion.
+
+The following actions are only available for `IBCEnabled` contracts:
+
+* Dispatch `IBCSendMsg` - this sends an IBC packet over an established
+  channel. This is returned from `handle` just like any other `CosmosMsg`
+  (For mocks, we will trigger this externally, later only
+  valid contract addresses should be able to do so).
+* Handle `IBCRecvPacket` - when another chain sends a packet destined to
+  an IBCEnabled contract, this will be routed to the proper contract
+  and call an exposed function for this purpose.
+* Handle `IBCAckPacket` - The original sender of `IBCSendMsg` will
+  get this callback eventually if the message was successfully
+  processed on the other chain
+* Handle `IBCErrorPacket` - The original sender of `IBCSendMsg` will 
+  get this callback eventually if the message failed to be
+  processed on the other chain (for any reason)
+
+For mocks, all the `IBCxxxPacket` commands are routed to some
+Golang stub handler, but containing the contract address, so we
+can perform contract-specific actions for each packet.
+
+There are open questions on how exactly the various `IBCxxPacket`
+messages are routed to a contract, as well as related to the
+channel lifecycle. (Can a contract open a channel? How do we
+allow contracts to participate in protocol version negotiation
+when establishing a channel?)
+
+## Ports and Channels
+
+There are two main proposals on how to map these concepts to
+contracts. We will describe both in detail, with arguments
+pro and contra, to select one approach.
+
+### One Port per Contract
+
+This may be the most straight-forward mapping, treating each contract 
+like a module. It does lead to very long portIDs however. Make special
+attention to both the Channel establishment (which should be compatible
+with standard ICS20 modules without changes on their part), as well
+as how contracts can properly identify their counterparty.
+
+* Upon `Instantiate`, if a contract is *IBC Enabled*, we dynamically 
+  bind a port for this contract. The port name is `wasm-<contract address>`,
+  eg. `wasm-cosmos1hmdudppzceg27qsuq707tjg8rkgj7g5hnvnw29`
+* If a *Channel* is being established with a registered `wasm-xyz` port,
+  the `x/wasm.Keeper` will handle this and call into the appropriate
+  contract to determine supported protocol versions during the
+  [`ChanOpenTry` and `ChanOpenAck` phases](https://docs.cosmos.network/master/ibc/overview.html#channels).
+  (See [Channel Handshake Version Negotiation](https://docs.cosmos.network/master/ibc/custom.html#channel-handshake-version-negotiation))
+* Both the *Port* and the *Channel* are fully owned by one contract.
+* `x/wasm` only accepts *ORDERED Channels* for simplicity of contract
+  correctness.
+* When sending a packet, the CosmWasm contract must specify the *ChannelID*
+  (and remote *PortID*?), which is generally provided by the external client
+  which understands which channels it wishes to communicate over.
+* When sending a packet, the contract can set a custom identifier (of
+  max 64 characters) that it can use to look this up later.
+* When receiving a Packet (or Ack or Error), the contracts receives the
+  *ChannelID* as well as remote *PortID* (and *ClientID*???) that is
+  communicating.
+* When receiving an Ack or Error packet, the contract also receives the
+  same identifier that it set on Send (`x/wasm` handles mapping between this
+  and the native IBC packet IDs).
+  
+### One Port per Module
+
+In this approach, the `x/wasm` module just binds one port to handle all
+modules. This can be well defined name like `wasm`. Since we always
+have `(ChannelID, PortID)` for routing messages, we can reuse one port
+for all contracts as long as we have a clear way to map the `ChannelID`
+to a specific contract when it is being established.
+
+
+* On genesis we bind the port `wasm` for all communication with the `x/wasm`
+  module.
+* The *Port* is fully owned by `x/wasm`
+* Each *Channel* is fully owned by one contract.
+* `x/wasm` only accepts *ORDERED Channels* for simplicity of contract
+  correctness.
+
+To clarify:
+
+* When a *Channel* is being established with port `wasm`, the
+  `x/wasm.Keeper` must be able to identify for which contract this
+  is destined. **how to do so**??
+  * One idea: the channel name must be the contract address. This means
+    (`wasm`, `cosmos13d...`) will map to the given contract in the wasm module.
+    The problem with this is that if two contracts from chainA want to
+    connect to the same contracts on chainB, they will want to claim the
+    same *ChannelID* and *PortID*. Not sure how to differentiate multiple
+    parties in this way.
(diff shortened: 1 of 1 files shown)


x/wasm/IBC.md
```diff
@@ -0,0 +1,128 @@
...
+* Handle `IBCRecvPacket` - when another chain sends a packet destined to
+  an IBCEnabled contract, this will be routed to the proper contract
+  and call an exposed function for this purpose.
+* Handle `IBCAckPacket` - The original sender of `IBCSendMsg` will
+  get this callback eventually if the message was successfully
+  processed on the other chain
+* Handle `IBCErrorPacket` - The original sender of `IBCSendMsg` will 
+  get this callback eventually if the message failed to be
+  processed on the other chain (for any reason)
+
```
alpe, line 39: We need to let contracts handle `timeouts` where the packet was not delivered in time for any reason so that they can revert their operation

x/wasm/IBC.md
```diff
@@ -0,0 +1,128 @@
...
+For mocks, all the `IBCxxxPacket` commands are routed to some
+Golang stub handler, but containing the contract address, so we
+can perform contract-specific actions for each packet.
+
+There are open questions on how exactly the various `IBCxxPacket`
+messages are routed to a contract, as well as related to the
+channel lifecycle. (Can a contract open a channel? How do we
+allow contracts to participate in protocol version negotiation
+when establishing a channel?)
```
alpe, line 48: This would be nice features for dynamic IBC

x/wasm/IBC.md
```diff
@@ -0,0 +1,128 @@
...
+* If a *Channel* is being established with a registered `wasm-xyz` port,
+  the `x/wasm.Keeper` will handle this and call into the appropriate
+  contract to determine supported protocol versions during the
+  [`ChanOpenTry` and `ChanOpenAck` phases](https://docs.cosmos.network/master/ibc/overview.html#channels).
+  (See [Channel Handshake Version Negotiation](https://docs.cosmos.network/master/ibc/custom.html#channel-handshake-version-negotiation))
+* Both the *Port* and the *Channel* are fully owned by one contract.
+* `x/wasm` only accepts *ORDERED Channels* for simplicity of contract
+  correctness.
+* When sending a packet, the CosmWasm contract must specify the *ChannelID*
+  (and remote *PortID*?), which is generally provided by the external client
```
alpe, line 76: We would need channel + port to send the packet

x/wasm/IBC.md
```diff
@@ -0,0 +1,128 @@
...
+
+This may be the most straight-forward mapping, treating each contract 
+like a module. It does lead to very long portIDs however. Make special
+attention to both the Channel establishment (which should be compatible
+with standard ICS20 modules without changes on their part), as well
+as how contracts can properly identify their counterparty.
+
+* Upon `Instantiate`, if a contract is *IBC Enabled*, we dynamically 
+  bind a port for this contract. The port name is `wasm-<contract address>`,
+  eg. `wasm-cosmos1hmdudppzceg27qsuq707tjg8rkgj7g5hnvnw29`
```
alpe, line 66: We should return this port either with ContractInfo or some other endpoint to make it explicit. I can imagine contract port lookups can become an issue if we leave this completely to the clients. The port naming schema   may change...?

x/wasm/IBC.md
```diff
@@ -0,0 +1,128 @@
...
+* When sending a packet, the CosmWasm contract must specify the *ChannelID*
+  (and remote *PortID*?), which is generally provided by the external client
+  which understands which channels it wishes to communicate over.
+* When sending a packet, the contract can set a custom identifier (of
+  max 64 characters) that it can use to look this up later.
+* When receiving a Packet (or Ack or Error), the contracts receives the
+  *ChannelID* as well as remote *PortID* (and *ClientID*???) that is
+  communicating.
+* When receiving an Ack or Error packet, the contract also receives the
+  same identifier that it set on Send (`x/wasm` handles mapping between this
```
alpe, line 84: I would prefer to pass the origin message payload here. It is available in the IBC packets. This would leave it to the contract to store and pass any identifier.

x/wasm/IBC.md
```diff
@@ -0,0 +1,128 @@
...
+    ```
+    Versions must be strings but can implement any versioning structure. 
+    If your application plans to have linear releases then semantic versioning is recommended.
+    ... 
+    Valid version selection includes selecting a compatible version identifier with a subset 
+    of features supported by your application for that version.
+    ...    
+    ICS20 currently implements basic string matching with a
+    single supported version.
+    ```
```
alpe, line 128: We could also put the sender/ receiver contract addresses as fields in the IBC packets we pass.

x/wasm/IBC.md
```diff
@@ -0,0 +1,128 @@
...
+channel lifecycle. (Can a contract open a channel? How do we
+allow contracts to participate in protocol version negotiation
+when establishing a channel?)
+
+## Ports and Channels
+
+There are two main proposals on how to map these concepts to
+contracts. We will describe both in detail, with arguments
+pro and contra, to select one approach.
+
```
alpe, line 55: I like "One Port per Contract" to start with now that the size limit is moved. Just thinking can there be scenarios scenarios where 1 port is not enough in dynamic IBC?

ethanfrey:
@alpe I made a lot of progress cleaning up the doc, and think I addressed all your issues as well as integrating ideas from your spike. Can you please review this again and give more feedback?

ethanfrey:
Okay, rebased this to focus on the stargate tracking branch, so we can look to merge this in, separate from the implementation work (which is a bit of mix of research, spikes, and useful code).

x/wasm/IBC.md
```diff
@@ -0,0 +1,253 @@
...
+  bind a port for this contract. The port name is `wasm-<contract address>`,
+  eg. `wasm-cosmos1hmdudppzceg27qsuq707tjg8rkgj7g5hnvnw29`
+* If a *Channel* is being established with a registered `wasm-xyz` port,
+  the `x/wasm.Keeper` will handle this and call into the appropriate
+  contract to determine supported protocol versions during the
+  [`ChanOpenTry` and `ChanOpenAck` phases](https://docs.cosmos.network/master/ibc/overview.html#channels).
+  (See [Channel Handshake Version Negotiation](https://docs.cosmos.network/master/ibc/custom.html#channel-handshake-version-negotiation))
+* Both the *Port* and the *Channel* are fully owned by one contract.
+* Question: `x/wasm` only accepts *ORDERED Channels* for simplicity of contract
+  correctness? Or should we allow the contract to decide?
```
alpe, line 200: IMHO there is no need to limit this in wasmd. Let the contract (framework) decide so that we have the most flexible solution in wasmd for future use cases.

x/wasm/IBC.md
```diff
@@ -0,0 +1,253 @@
...
+* When sending a packet, the CosmWasm contract must specify the *ChannelID*
+  and remote *PortID*, which is generally provided by the external client
+  which understands which channels it wishes to communicate over.
+* When sending a packet, the contract can set a custom identifier (of
```
alpe, line 204: This would go into the contract (application) frame. There is no need to support this in wasmd.

x/wasm/IBC.md
```diff
@@ -0,0 +1,253 @@
...
+
+**TODO**
+
+## Ports and Channels Discussion
+
+We decided on "one port per contract", especially after the IBC team raised
+the max length on port names to allow `wasm-<bech32 address>` to be a valid port.
+Here are the arguments for "one port for x/wasm" vs "one port per contract"
+
+### One Port per Contract
```
alpe, line 182: 👍  
As discussed on chat I want to add some ideas for future versions:
- multiple ports per contract
- elastic ports that can be assigned to different contracts
- transfer of channels to another contract
if there is some demand for this in the community...
We likely need a different naming schema for them to do proper routing

x/wasm/IBC.md
```diff
@@ -0,0 +1,253 @@
...
+Here we map out the workflow with the exact arguments passed with those calls 
+from the Go side (which we will use with our mock), and then a 
+proposal for multiplexing this over fewer wasm exports (define some rust types)
+
+Messages:
+
+```go
+type IBCSendMsg struct {
+    ChannelID string
+    RemotePort string // do we need both? isn't channel enough once it is established???
```
alpe, line 95: >... a single instance of an IBC module may communicate on the same port with any number of other modules and and IBC will correctly route all packets to the relevant module using the (channelID, portID tuple)

https://docs.cosmos.network/master/ibc/overview.html#capabilities

x/wasm/IBC.md
```diff
@@ -0,0 +1,253 @@
+# IBC specification
+
+This documents how CosmWasm contracts are expected to interact with IBC.
+
+## General Concepts
+
+**IBC Enabled** - when instantiating a contract, we detect if it supports IBC messages
```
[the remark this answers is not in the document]
ethanfrey: We use "allowed_features" in cosmwasm-vm already to ensure the contract is compatible with the vm (if it makes use of non-essential commands, the vm must also support them).

We expose which features the contract requested to go-cosmwasm, but currently don't pass that back to the sdk. We can just return that, and any contract that wishes to use the ibc messages will auto-request that feature and be detectable. We do this with staking, and chain-specific extensions now (like terra)

x/wasm/IBC.md
ethanfrey: Okay, we allow both ordered and unordered. And recommend ordered unless the developer can reason about distributed systems well.

x/wasm/IBC.md
ethanfrey: Agreed

x/wasm/IBC.md
ethanfrey: We can add a "future directions" section. But make it clear what the current spec is.

x/wasm/IBC.md
ethanfrey: @cwgoes can you explain this to me?

I understand **connection** as IP-level connection (the linked docs refers to **channel** as like IP connection)

A channel is like TCP/UDP connection which is established with source and remote ports via handshake. Once the channel is established, it has a well-defined port on both sides and simply the channelID should identify which client to use and the source/dest port, right?

ethanfrey:
Thanks for the feedback, I will update the docs

x/wasm/IBC.md
```diff
@@ -4,15 +4,51 @@ This documents how CosmWasm contracts are expected to interact with IBC.
...
+* Both the *Port* and the *Channel* are fully owned by one contract.
+* `x/wasm` will allow both *ORDERED* and *UNORDERED* channel and pass that mode
+  down to the contract in `OnTryOpenChannel`, so the contract can decide if it accepts
+  the mode. We will recommend the contract developers stick with *ORDERED* channels
+  unless they can reason about async packet timing.
+* When sending a packet, the CosmWasm contract must specify the *ChannelID*
+  and remote *PortID*, which is generally provided by the external client
+  which understands which channels it wishes to communicate over.
+* When receiving a Packet (or Ack or Error), the contracts receives the
+  *ChannelID* as well as remote *PortID* (and *ClientID*???) that is
```
alpe, line 46: ClientID is not available by the packet data out of the box but if this is required we should add a "get clientID by channel" to the supported queries.
Instead we have from the packet: 
```
	Sequence uint64
	// identifies the port on the sending chain.
	SourcePort string
	// identifies the channel end on the sending chain.
	SourceChannel string
	// block height after which the packet times out
	TimeoutHeight uint64
	// block timestamp (in nanoseconds) after which the packet times out
	TimeoutTimestamp uint64
```
and the destination PortID, ChannelID

x/wasm/IBC.md
```diff
@@ -79,13 +79,18 @@ Packet Lifecycle:
...
   get this callback eventually if the message failed to be
-  processed on the other chain (for timeout, closed channel, or the 
-  other side returning an error message in the IBC acknowledgement)
+  processed on the other chain (for timeout, closed channel, or 
+  other IBC-level failure)
+  
+Note: We may add some helpers inside the contract to map `IBCPacketAck` / `IBCPacketDropped`
+to `IBCPacketSucceeded` / `IBCPacketFailed` assuming they use the standard envelope. However,
+we decided not to enforce this on the Go-level, to allow contracts to communicate using protocols
+that do not use this envelope.
```
alpe, line 93: 👍  good to add the explanation

x/wasm/IBC.md
```diff
@@ -0,0 +1,281 @@
...
+* Upon `Instantiate`, if a contract is *IBC Enabled*, we dynamically 
+  bind a port for this contract. The port name is `wasm-<contract address>`,
+  eg. `wasm-cosmos1hmdudppzceg27qsuq707tjg8rkgj7g5hnvnw29`
+* If a *Channel* is being established with a registered `wasm-xyz` port,
+  the `x/wasm.Keeper` will handle this and call into the appropriate
+  contract to determine supported protocol versions during the
+  [`ChanOpenTry` and `ChanOpenAck` phases](https://docs.cosmos.network/master/ibc/overview.html#channels).
+  (See [Channel Handshake Version Negotiation](https://docs.cosmos.network/master/ibc/custom.html#channel-handshake-version-negotiation))
+* Both the *Port* and the *Channel* are fully owned by one contract.
+* `x/wasm` will allow both *ORDERED* and *UNORDERED* channel and pass that mode
```
cwgoes, line 38:
```suggestion
* `x/wasm` will allow both *ORDERED* and *UNORDERED* channels and pass that mode
```

x/wasm/IBC.md
```diff
@@ -0,0 +1,281 @@
...
+  down to the contract in `OnChanOpenTry`, so the contract can decide if it accepts
+  the mode. We will recommend the contract developers stick with *ORDERED* channels
+  unless they can reason about async packet timing.
+* When sending a packet, the CosmWasm contract must specify the *ChannelID*
```
cwgoes, line 42: If the contract specifies a local port ID & channel ID, that will uniquely determine the remote port ID & channel ID.

x/wasm/IBC.md
```diff
@@ -0,0 +1,281 @@
...
+Packet Lifecycle:
+
+* `IBCRecvPacket` - when another chain sends a packet destined to
+  an IBCEnabled contract, this will be routed to the proper contract
+  and call an exposed function for this purpose.
+* `IBCPacketAck` - The original sender of `IBCSendMsg` will
+  get this callback eventually if the message was
+  processed on the other chain (this may be either a success or an error,
+  but comes from the app-level protocol, not the IBC protocol).
+* `IBCPacketDropped` - The original sender of `IBCSendMsg` will
```
cwgoes, line 86: You can name it whatever you like, but this is all called a "timeout" in IBC (including channel closure)

x/wasm/IBC.md
```diff
@@ -0,0 +1,281 @@
...
+  a channel with this contract. It provides the protocol versions that
+  the initiate expects, and the contract can reject it or accept it
+  and return the protocol version it will communicate with.
+* `OnChanOpenConfirm` - this is called when the other party has agreed
+  to the channel and all negotiation is over. This will be called exactly
+  once for any established channel, and the contract should record these
+  open channels in the internal state.
+* `OnChanClosed` - this is called when an existing channel is closed
+  for any reason, allowing the client to update the state there.
+  This will (likely? @cwgoes?) be accompanied by `IBCPacketFailed`
```
cwgoes: Do you mean `IBCPacketDropped`? Yes, packets can be timed-out after a channel is closed

x/wasm/IBC.md
```diff
@@ -0,0 +1,281 @@
...
+Note: We may add some helpers inside the contract to map `IBCPacketAck` / `IBCPacketDropped`
+to `IBCPacketSucceeded` / `IBCPacketFailed` assuming they use the standard envelope. However,
+we decided not to enforce this on the Go-level, to allow contracts to communicate using protocols
+that do not use this envelope.
+
+Channel Lifecycle Hooks:
+
+* `OnChanOpenTry` - this is called when another chain attempts to open
+  a channel with this contract. It provides the protocol versions that
+  the initiate expects, and the contract can reject it or accept it
```
cwgoes:
```suggestion
  the initiating party expects, and the contract can reject it or accept it
```

x/wasm/IBC.md
cwgoes: Not quite; the terminology is a little different than TCP/IP.

A *connection* exists between two chains and has a unique identifier on each chain. Many channels may use one connection.

A *channel* has a unique identifier per-port, so the combination of port identifier & channel identifier is required to uniquely identify a channel on a single chain. A channel references a single connection, which it will utilise to verify state (that is, it will utilise the client associated with the connection).

Of course, if a contract is always bound to a fixed port, the contract can just identify channels by the channel identifier, and the wasmd runtime can look up the port, presumably.

ethanfrey:
> A channel has a unique identifier per-port, so the combination of port identifier & channel identifier is required to uniquely identify a channel on a single chain. A channel references a single connection, which it will utilise to verify state (that is, it will utilise the client associated with the connection).

This is the key point I didn't get. A `ChannelID` is not globally unique. So if I own portID `A`, I can open ChannelIDs `A`, `B`, `C`... and not worry if someone else claimed them first?

Now I get it. `(PortID, ChannelID)` is the globally unique identifier

x/wasm/IBC.md
ethanfrey: Okay, and since each contract only has one PortID, we can just auto-complete that part in `x/wasm`. The contract can specify `ChannelID` (it has a local namespace), and we can then generate `(PortID, ChannelID)` to use elsewhere as the global identifier.

I finally got this, makes a lot clearer

x/wasm/IBC.md
ethanfrey: Okay, I wasn't sure about all the different cases (timeout, channel closed, client no longer valid...)

I will use the standard naming.

cwgoes:
> > A channel has a unique identifier per-port, so the combination of port identifier & channel identifier is required to uniquely identify a channel on a single chain. A channel references a single connection, which it will utilise to verify state (that is, it will utilise the client associated with the connection).
> 
> This is the key point I didn't get. A `ChannelID` is not globally unique. So if I own portID `A`, I can open ChannelIDs `A`, `B`, `C`... and not worry if someone else claimed them first?
> 
> Now I get it. `(PortID, ChannelID)` is the globally unique identifier

Yes, exactly.

ethanfrey:
Thank you for the review @cwgoes 

I was actually doing a bit of more extension of the detailed callbacks while you wrote that. Will revisit that with my new understanding of why `(PortID, ChannelID)` are both required.

My last remaining question is how [`OnChanOpenTry`](https://github.com/cosmos/cosmos-sdk/blob/1a531cb6454a2e98147639655eb60bcd28b6d320/x/ibc-transfer/module.go#L210-L220) works.

It has a local channelID and a local version string (in addition to the  counterparty info). I guess the IBC module will auto-generate the channelID for it when a counterparty initiates a channel (is that true? if not where does it come from?). But shouldn't the module be able to set the `version`? In fact, I thought the main purpose for this callback was two-fold:

* Module can ensure that channel has a valid ordering
* Module checks the counterpartyVersion and either rejects it or accepts it and returns a compatible version we accept locally (maybe an older version, or a few features turned off). The way this callback is written, I would assume the relayer(?) is picking the version string here?

I cannot find much of this in the docs, (and this is sdk-specific stuff, not spec stuff). Maybe you can explain more?

cwgoes:
> It has a local channelID and a local version string (in addition to the counterparty info). I guess the IBC module will auto-generate the channelID for it when a counterparty initiates a channel (is that true? if not where does it come from?). But shouldn't the module be able to set the version? In fact, I thought the main purpose for this callback was two-fold:

By default, the initiator picks a remote port & channel identifier in addition to a local one - but of course the destination module can choose to reject the proposed identifier (and end the handshake). Per discussions with Agoric (https://github.com/cosmos/ics/issues/462) we plan to modify the handshake a bit to allow the destination chain to pick the identifier in a different handshake mode; this has not yet been done in the SDK.

> The way this callback is written, I would assume the relayer(?) is picking the version string here?

Yes, that's right - the version provided therein is in the message (datagram); if you want the contract to calculate it directly, you can use an architecture similar to [ADR 025](https://github.com/cosmos/cosmos-sdk/blob/master/docs/architecture/adr-025-ibc-passive-channels.md).

ethanfrey:
> By default, the initiator picks a remote port & channel identifier in addition to a local one

This is a bit tricky, as the initiator must know which channel ids are taken/free on the other side. This may make sense from the point of view of a relayer, but from a contract, one more external piece of info needed.

> Yes, that's right - the version provided therein is in the message (datagram); if you want the contract to calculate it directly, you can use an architecture similar to ADR 025.

Very nice ADR025. Thank you for referring me there. Is this possible in the current ibc implementation in the cosmos SDK just by controlling the handler in `x/wasm`? Or do I need to make some minor changes to the ibc-04-channel module as well?

Also, ADR025 allows the module to perform the version negotiation, but it still requires the initiating party of the channel to specify the channelID on the counterparty, correct?

In fact, more than just the versioning aspect,  this "passive relayer" approach is the only one that scales. There can be only 1 active relayer that creates the channel and that seems very centralized (and prone to crashing). On the other hand, there may be N different passive relayers, each relaying some subset of the channels between two chains and providing redundancy. But that is another topic. Looking forward to production quality relayers once stargate gets stabilized.

cwgoes:
> Very nice ADR025. Thank you for referring me there. Is this possible in the current ibc implementation in the cosmos SDK just by controlling the handler in x/wasm? Or do I need to make some minor changes to the ibc-04-channel module as well?

Some minor changes may be required, I'm not quite sure - paging @michaelfig, maybe there's room for collaboration here.

> Also, ADR025 allows the module to perform the version negotiation, but it still requires the initiating party of the channel to specify the channelID on the counterparty, correct?

Yes; but that will change with https://github.com/cosmos/ics/issues/462.

x/wasm/IBC.md
```diff
@@ -0,0 +1,425 @@
...
+```go
+package messages
+
+type IBCSendMsg struct {
+    // This is our contract-local ID
+    ChannelID string
+    Msg []byte
+    // optional fields (or do we need exactly/at least one of these?)
+    TimeoutHeight uint64
+    TimeoutTimestamp uint64
```
alpe, line 167: We need the `Sequence` here as well to have all fields for an [ibc packet](https://github.com/cosmos/cosmos-sdk/blob/master/x/ibc/04-channel/types/packet.go#L30)
`SourcePortID` is known to us from the contract context and we can lookup the counterpart with the channelID:portID pair

x/wasm/IBC.md
```diff
@@ -0,0 +1,425 @@
...
+
+}
+
+type IBCCloseChannel struct {
+    ChannelID string
+}
+
+// use type from https://github.com/cosmos/cosmos-sdk/blob/master/x/ibc/03-connection/types/version.go ?
+// or is there a better generic model?
+type Version string
```
alpe, line 195: `string` is the most flexible one. Let's not restrict this here.

x/wasm/IBC.md
```diff
@@ -0,0 +1,425 @@
...
+   ChannelIDs []ChannelMetadata
+}
+
+type ChannelMetadata struct {
+    // Local portID, channelID is our unique identifier
+    PortID string
+    ChannelID string
+    RemotePortID string
+    Order channeltypes.Order
+    CounterpartyVerson Version
```
alpe, line 325: The `channel.version` is not fully clear to me. Either it is the counterparty version, the local one or both as they are the same. Go doc states: " opaque channel version, which is agreed upon during the handshake"

x/wasm/IBC.md
```diff
@@ -0,0 +1,425 @@
...
+}
+
+type QueryPendingPackets struct {
```
alpe, line 328: Can be unrelayed acknowledgements or unrelayed packet commitments. 
The CLI has a ` --acknowledgements` flag for this

x/wasm/IBC.md
```diff
@@ -0,0 +1,425 @@
...
+    // Always required
+    ChannelID string
+
+    // exactly one of these must be set. ContractAddress is a shortcut to save the Contract->PortID mapping
+    PortID string
+	ContractAddress string
+}
```
alpe, line 335: the CLI has `--sequences :comma separated list of packet sequence numbers (default [])` flag. Not sure what it does.

x/wasm/IBC.md
```diff
@@ -0,0 +1,425 @@
...
+type IBCCloseChannel struct {
+    ChannelID string
+}
+
+// use type from https://github.com/cosmos/cosmos-sdk/blob/master/x/ibc/03-connection/types/version.go ?
+// or is there a better generic model?
+type Version string
+```
+
+Packet callbacks:
```
alpe, line 198: I will comment on the callbacks tomorrow with some more inspiration from the spike. Or we merge this and I do a PR on top

x/wasm/IBC.md
alpe: > During the handshake process, two ends of a channel come to agreement on a version bytestring associated with that channel. 

https://github.com/cosmos/ics/tree/master/spec/ics-004-channel-and-packet-semantics#versioning

x/wasm/IBC.md
```diff
@@ -0,0 +1,425 @@
...
+to `IBCPacketSucceeded` / `IBCPacketFailed` assuming they use the standard envelope. However,
+we decided not to enforce this on the Go-level, to allow contracts to communicate using protocols
+that do not use this envelope.
+
+### Channel Lifecycle Hooks
+
+If you look at the [4 step process](https://docs.cosmos.network/master/ibc/overview.html#channels) for
+channel handshakes, we simplify this from the view of the contract:
+
+1. Channels *cannot* be opened by external clients, only the contract can initiate opening
```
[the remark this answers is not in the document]
ethanfrey: Sure, I guess that makes sense and is flexible. Then the contract on the initiating end should also have a say if this is a reasonable connection before sending the init message to the other chain

x/wasm/IBC.md
```diff
@@ -0,0 +1,425 @@
...
+   if we realize this doesn't work. Or save the info (we may need to define this channel uses an older version
+   of the protocol for example).
+4. `OnChanClosed` is called on both sides if the channel is closed for any reason, allowing them to
+   perform any cleanup. This will be followed by `IBCPacketTimeout` callbacks for all the in-progress 
+   packets that were not processed before it closed, as well as all pending acks (pending the relayer
+   to provide that).
+
+We require the following callbacks on the contract
+
+* `OnChanNegotiate` - called on receiving end for `OnChanOpenTry`.
```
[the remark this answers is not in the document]
ethanfrey: Is there much difference between `OnChanOpenInit` (if the version / channel / etc is provided by a 3rd party) and `OnChanOpenTry`? Maybe `OnChanNegotiate` can just be a check that the ordering and version (and remote port?) is acceptable.

x/wasm/IBC.md
```diff
@@ -0,0 +1,425 @@
...
+I thought the module would accept the Version and decide on it's own CounterpartyVersion. We assume the
+relayer makes that decision????
+
+### Queries
+
+We may want to expose some basic queries of IBC State to the contract.
+We should check if this is useful to the contract and if it opens up
+any possible DoS:
+
+* `GetPortID` - return PortID given a contract address
```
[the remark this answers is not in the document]
ethanfrey: Yeah, this also let's you get the port id for other contracts you know of. But mainly to pull it out of some custom Env struct.

x/wasm/IBC.md
```diff
@@ -0,0 +1,425 @@
...
+* `ListChannels` - return a list of all (portID, channelID) pairs
+  that are bound to a given port.
+* `ListPendingPackets` - given a (portID, channelID) identifier, return all packets
+  in that channel, that have been sent by this chain, but for which no acknowledgement 
+  or timeout has yet been received
```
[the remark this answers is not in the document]
ethanfrey: `GetCounterparty` returns the (portID, channelID) on the remote chain?

x/wasm/IBC.md
ethanfrey: The sequence should not be set by the contract, but auto-assigned by the keeper/ibc module. 

>  SourcePortID is known to us from the contract context and we can lookup the counterpart with the channelID:portID pair

Yes, I should add that as a document. The ChannelID (and context-contained PortID) should be enough to uniquely identify the channel and remote counterparty.

x/wasm/IBC.md
ethanfrey: Great, we can just add some suggestions of optional versioning schemes

x/wasm/IBC.md
ethanfrey: You are right, I was confused with `ChannelInfo` above (which is used *during* the negotiation process).

Once established, there is just a Version.

x/wasm/IBC.md
ethanfrey: Good point. I would make 2 separate queries here.

* One for pending packets (our requests).
* The other for pending acknowlegdements (our responses)

x/wasm/IBC.md
ethanfrey: Interesting find. I guess you can select details of some by sequence number.

I would leave that out and add it later if needed. Even adding these query packet functionality seems questionably useful for a contract (maybe to see if there is congestion). Hunting specific packets seems more useful for client side visualization, but not here

ethanfrey (commented):
Thanks for the review Alex, I will integrate some of this tomorrow morning.

I would appreciate a PR on top of this with the PacketCallback adjustments you had.

I may merge this first and then merge that second, or the other way around, depending on our timing.

x/wasm/IBC.md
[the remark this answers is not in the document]
alpe: yes, exactly

x/wasm/IBC.md
alpe: 👍  correct. we can use `channelKeeper.GetNextSequenceSend(ctx, sourcePort, sourceChannel)` to get the next sequence to build the package. Do I understand the workflow correct?
1. the contract returns an IBCSendMsg
2. in our handler_plugin we create the `channeltypes.Packet`
3. we send the packet to `channelKeeper.SendPacket`
3a. within the handler_plugin
3b return it and send it from the keeper
4. relayer sends packet to other chain....

ethanfrey: That sounds correct.

Now that you mention the details of the HandlerPlugin interface, it doesn't by itself have the ability to query into the channelKeeper to get that info. I don't want to push that responsibility down to the contract and I guess we will have to extend the HandlerPlugin interface to have more capabilities during the transform.

PR #216, opened 2020-07-23
removed reseting of gas meter each time we enter a query

reuvenpo:
This PR removes a line that was reseting the gas meter for each query, which meant that recursive queries would not report any gas usage, since it would be charged on a new meter, which the caller has no access to.
We now observe that gas is charged for external queries.
I expect some tests will fail after this first commit. Let;s see if they do. Are there tests that run such contract-to-contract queries?

### DIFF
--- a/x/wasm/internal/keeper/keeper.go
+++ b/x/wasm/internal/keeper/keeper.go
@@ -397,8 +397,6 @@ func (k Keeper) GetContractHistory(ctx sdk.Context, contractAddr sdk.AccAddress)
 func (k Keeper) QuerySmart(ctx sdk.Context, contractAddr sdk.AccAddress, req []byte) ([]byte, error) {
 	ctx.GasMeter().ConsumeGas(InstanceCost, "Loading CosmWasm module: query")
 
-	ctx = ctx.WithGasMeter(sdk.NewGasMeter(k.queryGasLimit))
-
 	codeInfo, prefixStore, err := k.contractInstance(ctx, contractAddr)
 	if err != nil {
 		return nil, err


ethanfrey:
@reuvenpo thank you for bringing this forward.

I am working on https://github.com/CosmWasm/cosmwasm/pull/494 to add such test potential to our standard test contract (your recursion idea plus a variable gas burning - sha256 hash loop). That is on 0.10.0 and we need to update go-cosmwasm https://github.com/CosmWasm/go-cosmwasm/pull/123 before we can use it.

I saw this last night after you pointed the bug. We need a different entry point for the external smart queries than the internal keeper. OR we do... but it is there at the external query handler that we need to set the gas limit, not in the keeper.

I built the contract to demo this issue. Unless you are really blocked by this, I would like to update everything to 0.10.0-alpha and get the test contract here with failing test cases before we start fixing bugs. Best way to ensure we squash them and there are no regressions

assafmo:
I don't think there's a need for different entry point. I guess up until now queries didn't charge any gas and no one cared?

reuvenpo:
We are not blocked by this. These last few PRs have been backports of changes that we have made in our repo, back to yours. We're fixing these bugs on our side, so we don't even need to pull in the full set of updates from your side yet.

ethanfrey:
I just merged #218 that expos ees 0.10 cosmwasn here and also the new gas metering. And the recursive query for testing.

I will look at this soon, tests first

ethanfrey:
Thanks for finding the issue and pushing a patch.

I fixed this in #226 while still enforcing the limit on external queries. And adding a number of tests to ensure it works.

assafmo:
Cool :+1:

PR #220, opened 2020-07-24
Add wasm lcd smoke test

alpe:
Test wasm store code via rest call.
______

For admin use:

- [ ] Reviewers assigned

### DIFF
--- a/lcd_test/wasm_test.go
+++ b/lcd_test/wasm_test.go
@@ -1,0 +1,79 @@
+package lcdtest
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"testing"
+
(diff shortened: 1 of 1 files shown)


alpe:
👌 I will update this one or add one on the the handler level

ethanfrey:
We can also merge this into master (as I'm going through all PRs) and do a follow-up PR with more coverage. This is better than before.

alpe:
I checked how to adopt the test to run against the issue we had with the missing fields but it would not make sense to squeeze that into this one. 
The LCD tests call the rest handler for creating the message json before signing. To replicate the issue we would have to bypass that step and create a custom json payload.
That is a valuable test but we would not necessary need Go for this. Let's discuss if integrating cosmJS into the CI pipeline makes sense, instead.

PR #226, opened 2020-07-27
Fix gas calculation on QuerySmart

ethanfrey:
Inspired by and replaces #216 

* We only enforce the query gas limit (with a different context) when calling in from external queries (see Querier)
* Add tests that queries charge gas
* Add tests that queries charge gas when recusing one level
* Add tests that recursing 4 times doesn't cause errors (fixed in #213 but lacking tests)

- [ ] Wrote tests

______

For admin use:

- [ ] Reviewers assigned

### DIFF
--- a/x/wasm/internal/keeper/recurse_test.go
+++ b/x/wasm/internal/keeper/recurse_test.go
@@ -2,6 +2,7 @@ package keeper
 
 import (
 	"encoding/json"
+	abci "github.com/tendermint/tendermint/abci/types"
 	"io/ioutil"
 	"os"
 	"testing"
@@ -81,6 +82,7 @@ func TestGasCostOnQuery(t *testing.T) {
 				Depth: 4,
 				Work:  50,
 			},
+			// this is (currently) 244_708 gas
 			expectedGas: 5*GasWork50 + 4*GasReturnHashed,
 		},
 	}
@@ -116,6 +118,9 @@ func TestGasCostOnQuery(t *testing.T) {
 	for name, tc := range cases {
 		tc := tc
 		t.Run(name, func(t *testing.T) {
+			// external limit has no effect (we get a panic if this is enforced)
+			keeper.queryGasLimit = 1000
+
 			// make sure we set a limit before calling
 			ctx = ctx.WithGasMeter(sdk.NewGasMeter(tc.gasLimit))
 			require.Equal(t, uint64(0), ctx.GasMeter().GasConsumed())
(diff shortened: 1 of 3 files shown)


x/wasm/internal/keeper/recurse_test.go
```diff
@@ -0,0 +1,259 @@
...
+
+			// make sure we set a limit before calling
+			ctx = ctx.WithGasMeter(sdk.NewGasMeter(tc.gasLimit))
+			require.Equal(t, uint64(0), ctx.GasMeter().GasConsumed())
+
+			recurse := tc.msg
+			recurse.Contract = contractAddr
+			msg := buildQuery(t, recurse)
+
+			// this should throw out of gas exception (panic)
```
alpe, line 132: q: is this a leftover from the other test?

x/wasm/internal/keeper/recurse_test.go
```diff
@@ -0,0 +1,259 @@
...
+		Verifier:    fred,
+		Beneficiary: bob,
+	}
+	initMsgBz, err := json.Marshal(initMsg)
+	require.NoError(t, err)
+	contractAddr, err := keeper.Instantiate(ctx, codeID, creator, nil, initMsgBz, "recursive contract", deposit)
+	require.NoError(t, err)
+
+	for name, tc := range cases {
+		tc := tc
```
alpe, line 119: nit: This is not needed

x/wasm/internal/keeper/recurse_test.go
```diff
@@ -0,0 +1,259 @@
...
+			require.Equal(t, uint64(0), ctx.GasMeter().GasConsumed())
+
+			recurse := tc.msg
+			recurse.Contract = contractAddr
+			msg := buildQuery(t, recurse)
+
+			// this should throw out of gas exception (panic)
+			data, err := keeper.QuerySmart(ctx, contractAddr, msg)
+			require.NoError(t, err)
+			var resp recurseResponse
```
alpe, line 135: pp: As you testing for the gas cost the response value does not really contribute to this goal. I would remove L135-146 to stay focused here

x/wasm/internal/keeper/recurse_test.go
```diff
@@ -0,0 +1,259 @@
...
+		Verifier:    fred,
+		Beneficiary: bob,
+	}
+	initMsgBz, err := json.Marshal(initMsg)
+	require.NoError(t, err)
+	contractAddr, err := keeper.Instantiate(ctx, codeID, creator, nil, initMsgBz, "recursive contract", deposit)
+	require.NoError(t, err)
+
+	for name, tc := range cases {
+		tc := tc
```
alpe, line 224: nit: not needed

x/wasm/internal/keeper/recurse_test.go
```diff
@@ -0,0 +1,259 @@
...
+					// this should run out of gas
+					_, _ = NewQuerier(keeper)(ctx, path, req)
+				})
+				return
+			}
+
+			// otherwise, make sure we get a good success
+			data, err := NewQuerier(keeper)(ctx, path, req)
+			require.NoError(t, err)
+			var resp recurseResponse
```
alpe, line 247: pp: end test with NoError in line before. No need to check the response here

x/wasm/internal/keeper/recurse_test.go
ethanfrey: Oh yes it is.. at least with the golangci linter on the cosmos-sdk.

I guess I can keep it out of here, but am a bit traumatized by that linter (and the long feedback cycle)

x/wasm/internal/keeper/recurse_test.go
ethanfrey: I want to make sure the right stuff is coming out, but I could do that in another test.
I will remove the check for the higher-level external queries

PR #227, opened 2020-07-27
Cleanup dockerfile and add coral support

ethanfrey:
* Remove ledger support from built artifacts (longer buildtime, deps, not useful in docker)
* Add `coral`/`corald` binaries

- [ ] Wrote tests

______

For admin use:

- [ ] Reviewers assigned

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,6 +34,8 @@ COPY --from=rust-builder /lib/libgo_cosmwasm_muslc.a /lib/libgo_cosmwasm_muslc.a
 
 # force it to use static lib (from above) not standard libgo_cosmwasm.so file
 RUN LEDGER_ENABLED=false BUILD_TAGS=muslc make build
+# we also (temporarily) build the testnet binaries here
+RUN LEDGER_ENABLED=false BUILD_TAGS=muslc make build-coral
 
 # --------------------------------------------------------
 FROM alpine:3.12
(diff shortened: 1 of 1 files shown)


ethanfrey:
I would like multiple dockerfiles sometime in the future, but that makes the build time huge 

Let's revisit this later, but I would do this for coral and gaiaflex

Dockerfile
```diff
@@ -24,16 +24,18 @@ RUN export GO_WASM_DIR=$(go list -f "{{ .Dir }}" -m github.com/CosmWasm/go-cosmw
 FROM cosmwasm/go-ext-builder:0.8.2-alpine AS go-builder
 
 RUN apk add git
-# without this, build with LEDGER_ENABLED=false
-RUN apk add libusb-dev linux-headers
+# NOTE: add these to run with LEDGER_ENABLED=true
+# RUN apk add libusb-dev linux-headers
```
webmaster128, line 28: There is another block like this a few lines above (line 6-7). Should we do the same there as well?

ethanfrey: Good call. I can remove that one entirely.
I guess I'll throw that in an existing branch - #229 to avoid a micro-pr

ethanfrey: https://github.com/CosmWasm/wasmd/pull/229/commits/08aef4a650afb94b4a80f64d80d185087629348e

PR #229, opened 2020-07-27
Add gaiaflex to Makefile and Dockerfile

### LINKED ISSUE #210: Combine wasmd and wasmgovd into one code package
## Summary

It is important to have these two binaries, but it seems heavy to require to copy all code to do so. 

## Problem Definition

Since the only difference is one configuration option, which is `[]string{}`, we could pass this in from the cli on the build step. 

## Proposal

Update https://github.com/CosmWasm/wasmd/blob/master/cmd/wasmd/main.go

```go
var EnabledProposals = ""

func getEnabledProposals() []string {
  if len(EnabledProposals) == 0 {
    return []string{}
  }
  return strings.Split(EnabledProposals, ",")
}
```

Then we [set a linker flag](https://github.com/CosmWasm/wasmd/blob/master/Makefile#L50-L55) in the Makefile, like

`-X github.com/CosmWasm/wasmd/cmd/wasmd.EnabledProposals=StoreCode,InstantiateContract`

And build two different binaries from the same `cmd/wasmd` directory, just using a different build flag
____

#### For Admin Use

- [ ] Not duplicate issue

ethanfrey:
Closes #210 (removed wasmgovd)

* Create a new Makefile target `build-gaiaflex`
* Add it to the Dockerfile
* Allow us to set the enabled proposals via build flags:

`-X github.com/CosmWasm/wasmd/app.ProposalsEnabled=true` to enable all governance proposals (default all disabled)

`-X github.com/CosmWasm/wasmd/app.EnableSpecificProposals=MigrateContract,UpdateAdmin,ClearAdmin` to enable some governance proposals


ethanfrey:
@orkunkl you can follow this to see when gaiaflex is in the Dockerfile

Makefile
```diff
@@ -97,6 +114,15 @@ else
 	go build -mod=readonly $(CORAL_BUILD_FLAGS) -o build/coral ./cmd/wasmcli
 endif
 
+build-gaiaflex: go.sum
+ifeq ($(OS),Windows_NT)
+    # wasmd nodes not supported on windows, maybe the cli?
```
webmaster128, line 119: Wrong indent here (must be tabs). Doesn't this break Makefiles?

ethanfrey: Fixed

PR #230, opened 2020-07-27
Remove wasmgovd binary

### LINKED ISSUE #210: Combine wasmd and wasmgovd into one code package

ethanfrey:
(This is an optional addition to the previous PR #229)

Closes #210 

* Use `gaiaflex` instead of `wasmgovd` as reference.
* Remove all `wasmgovd` code
* Document how to use the ldflags to configure this in the Makefile


README.md
```diff
@@ -129,6 +129,27 @@ docker run --rm -it \
     cosmwasm/wasmd:latest ./logs.sh
 ```
 
+## Runtime flags
```
alpe, line 132: 🌷 Good to have this documented

PR #231, opened 2020-07-27
Alex ibc spike2

ethanfrey:
This is not my branch, but I made a PR so I can comment on this better and get more visibility


ethanfrey:
Okay, some learnings of mine going through this....

[`OnAcknowledgementPacket`](https://github.com/cosmos/cosmos-sdk/blob/master/x/ibc-transfer/module.go#L323-L334) is called every on **both** success and failure. This makes the callback flow a bit more confusing.

Exactly one callback will come from a given packet. If [`OnTimeoutPacket`](https://github.com/cosmos/cosmos-sdk/blob/master/x/ibc-transfer/keeper/relay.go#L207-L209) is called, then we treat it as an error. If  [`OnAcknowledgementPacket`](https://github.com/cosmos/cosmos-sdk/blob/master/x/ibc-transfer/keeper/relay.go#L200-L205) is called, then we check the data packet and handle either success or error.

This does make sense for lower-level coding, but for the contracts, it would be nicer to expose something like `onAckSucceed` with *only success packets* and `onAckError` with *only failed packets*. I dream of making some standard envelope for `OnAcknowledgementPacket` such that we can parse success/error in a generic way, then pass arbitrary "sucess bytes" to the success case.

x/wasm/internal/keeper/cosmwasm/contract.go
```diff
@@ -0,0 +1,35 @@
...
+	Log []wasmTypes.LogAttribute `json:"log"`
+}
+type OnTimeoutIBCResponse struct {
+	Messages []sdk.Msg `json:"messages"`
+
+	// log message to return over abci interface
+	Log []wasmTypes.LogAttribute `json:"log"`
+}
+
+type AcceptChannelResponse struct {
```
ethanfrey, line 31: This would be one more contract callback? for the `OnTryInitChannel` or whatever it is called?

x/wasm/internal/keeper/cosmwasm/query.go
```diff
@@ -0,0 +1,18 @@
+package cosmwasm
+
+type IBCQuery struct {
```
ethanfrey, line 3: Yeah,  didn't even think about the queries. 

This is a good point, we should define them well.

x/wasm/internal/keeper/ibc.go
```diff
@@ -77,28 +80,176 @@ func (k Keeper) ClaimCapability(ctx sdk.Context, cap *capabilitytypes.Capability
...
+	OnAcknowledgement(hash []byte, params cosmwasm.Env, originalData []byte, acknowledgement []byte, store prefix.Store, api wasm.GoAPI, querier QueryHandler, meter sdk.GasMeter, gas uint64) (*cosmwasm.OnAcknowledgeIBCResponse, uint64, error)
+	OnTimeout(hash []byte, params cosmwasm.Env, msg []byte, store prefix.Store, api wasm.GoAPI, querier QueryHandler, meter sdk.GasMeter, gas uint64) (*cosmwasm.OnTimeoutIBCResponse, uint64, error)
+	// IBC channel livecycle
+	AcceptChannel(hash []byte, params cosmwasm.Env, order channeltypes.Order, version string, connectionHops []string, store prefix.Store, api wasm.GoAPI, querier QueryHandler, meter sdk.GasMeter, gas uint64) (*cosmwasm.AcceptChannelResponse, uint64, error)
+	//OnConnect(hash []byte, params cosmwasm.Env, msg []byte, store prefix.Store, api wasm.GoAPI, querier QueryHandler, meter sdk.GasMeter, gas uint64) (*cosmwasm.OnTimeoutIBCResponse, uint64, error)
+	//OnClose(hash []byte, params cosmwasm.Env, msg []byte, store prefix.Store, api wasm.GoAPI, querier QueryHandler, meter sdk.GasMeter, gas uint64) (*cosmwasm.OnTimeoutIBCResponse, uint64, error)
```
ethanfrey, line 91: Hmmm.... we probably do want this one.

I'm not sure when `OnConnect` would be called, or what it would be used for. You can get the same info from an IBCQuery and I don't want to give the contracts too much agency to randomly run code in arbitrary lifecycle events.

But if you have a communication channel with another module and it closes, it is probably nice to allow some cleanup? Or will I simply see the channel gone on the next query and receive `OnTimeout` events for all pending packets? That may be enough for cleanup, but I could see wanting to do some cleanup.

The issue would be the contract just erroring on `OnClose` and never letting the IBC module do the cleanup. Hmmm...

x/wasm/internal/keeper/ibc.go
```diff
@@ -77,28 +80,176 @@ func (k Keeper) ClaimCapability(ctx sdk.Context, cap *capabilitytypes.Capability
...
+}
+
+var MockContracts = make(map[string]IBCCallbacks, 0)
+
+func (k Keeper) AcceptChannel(ctx sdk.Context, contractAddr sdk.AccAddress, order channeltypes.Order, version string, connectionHops []string, ibcInfo cosmwasm.IBCInfo) ([]string, error) {
```
ethanfrey, line 96: we want to enforce `order == ORDERED`

x/wasm/internal/keeper/ibc.go
```diff
@@ -77,28 +80,176 @@ func (k Keeper) ClaimCapability(ctx sdk.Context, cap *capabilitytypes.Capability
...
+		Ctx:     ctx,
+		Plugins: k.queryPlugins,
+	}
+
+	gas := gasForContract(ctx)
+	mock, ok := MockContracts[contractAddr.String()]
+	if !ok { // hack for testing without wasmer
+		panic("not supported")
+	}
+	res, gasUsed, execErr := mock.AcceptChannel(codeInfo.CodeHash, params, order, version, connectionHops, prefixStore, cosmwasmAPI, querier, ctx.GasMeter(), gas)
```
ethanfrey, line 116: Nice way to map this out.

I would not pass in order (can be too confusing for devs) and enforce ORDERED for now. Also, nicer to embed lots of different args in a struct `order, version, connectionHops` like the `IBCInfo` struct above. (not identical, but something like that level of embedding)

x/wasm/internal/keeper/ibc.go
```diff
@@ -77,28 +80,176 @@ func (k Keeper) ClaimCapability(ctx sdk.Context, cap *capabilitytypes.Capability
...
+		Ctx:     ctx,
+		Plugins: k.queryPlugins,
+	}
+
+	gas := gasForContract(ctx)
+	mock, ok := MockContracts[contractAddr.String()]
+	if !ok { // hack for testing without wasmer
+		panic("not supported")
+	}
+	res, gasUsed, execErr := mock.OnReceive(codeInfo.CodeHash, params, payloadData, prefixStore, cosmwasmAPI, querier, ctx.GasMeter(), gas)
```
ethanfrey, line 147: This will need to know what channel the message is coming in on. that is not included in payload data. Any other context you think as useful?

x/wasm/internal/keeper/ibc.go
```diff
@@ -77,28 +80,176 @@ func (k Keeper) ClaimCapability(ctx sdk.Context, cap *capabilitytypes.Capability
...
+		Ctx:     ctx,
+		Plugins: k.queryPlugins,
+	}
+
+	gas := gasForContract(ctx)
+	mock, ok := MockContracts[contractAddr.String()] // hack for testing without wasmer
+	if !ok {
+		panic("not supported")
+	}
+	res, gasUsed, execErr := mock.OnAcknowledgement(codeInfo.CodeHash, params, payloadData, acknowledgement, prefixStore, cosmwasmAPI, querier, ctx.GasMeter(), gas)
```
ethanfrey, line 186: As above, we need to pass minimally the channelID here.

Also, I would love to be able to detect ack as success/error in our go code (in a fixed manner) before sending to the contract

x/wasm/internal/keeper/ibc.go
```diff
@@ -77,28 +80,176 @@ func (k Keeper) ClaimCapability(ctx sdk.Context, cap *capabilitytypes.Capability
...
-	ctx sdk.Context,
-	sourcePort,
-	sourceChannel string,
-	sender sdk.AccAddress,
-	receiver sdk.AccAddress,
-	timeoutHeight,
-	timeoutTimestamp uint64,
-	msg json.RawMessage,
-) error {
+func (k Keeper) IBCCallFromContract(ctx sdk.Context, sourcePort, sourceChannel string, sender sdk.AccAddress, timeoutHeight, timeoutTimestamp uint64, msg json.RawMessage) error {
```
ethanfrey, line 244: Cool idea, so this is another `CosmosMsg` type that the contract can return.

`IBCSendPacket` and `IBCOpenChannel` ?

Good to look at this code, I need to combine this with the comments on #214 and get a clear API spec'd out.

x/wasm/internal/keeper/ibc_mocks_test.go
```diff
@@ -0,0 +1,143 @@
...
+)
+
+type mockContract struct {
+	app          *app.WasmApp
+	t            *testing.T
+	contractAddr sdk.AccAddress
+}
+
+func (c *mockContract) AcceptChannel(hash []byte, params cosmwasm2.Env, order channeltypes.Order, version string, connectionHops []string, store prefix.Store, api cosmwasm.GoAPI, querier keeper.QueryHandler, meter sdk.GasMeter, gas uint64) (*cosmwasm2.AcceptChannelResponse, uint64, error) {
+	if order != channeltypes.ORDERED {
```
ethanfrey, line 30: Hmm... maybe we let the contract use UNORDERED as well?

x/wasm/internal/keeper/ibc.go
ethanfrey: We also need some way to map back to the original packet that we sent. This is not just an unsolicited acknowledgement (like receive), but an acknowledgement that ties to one specific packet.

In ibc transfer they take the [packet metadata (port and channel) and use it to determine the refund](https://github.com/cosmos/cosmos-sdk/blob/master/x/ibc-transfer/keeper/relay.go#L207-L219)

x/wasm/internal/keeper/ibc_mocks_test.go
```diff
@@ -0,0 +1,143 @@
...
+
+func MockContract(t *testing.T, contractAddr sdk.AccAddress, app *app.WasmApp) *mockContract {
+	c := &mockContract{t: t,
+		contractAddr: contractAddr,
+		app:          app,
+	}
+	keeper.MockContracts[contractAddr.String()] = c
+	return c
+}
+func (c *mockContract) OnReceive(hash []byte, params cosmwasm2.Env, msg []byte, store prefix.Store, api cosmwasm.GoAPI, querier keeper.QueryHandler, meter sdk.GasMeter, gas uint64) (*cosmwasm2.OnReceiveIBCResponse, uint64, error) {
```
ethanfrey, line 47: Nice mocks.. this whole file

alpe:
> OnAcknowledgementPacket is called every on both success and failure. This makes the callback flow a bit more confusing.

This is the application level. An error like "ErrInsufficientFunds" in ibc-transfer would be handled there. I think about this as an RPC call where the other side returns some method specific values. The response can be like (ok, error) or more complex. It would also serve as a commit in a DB transaction. Regardless of the payload, the IBC package was delivered successfully.

The `OnTimeoutPacket` is the transaction rollback equivalent.

I assume there can be IBC related errors like "the port is no longer able to accept channels, such as if the Connection to the remote chain has failed, perhaps because a consensus failure was observed" as the third group.



> I dream of making some standard envelope for OnAcknowledgementPacket such that we can parse success/error in a generic way, then pass arbitrary "sucess bytes" to the success case.

I see this in the cosmwasm rust framework layer. We could provide handlers and other tooling there and keep wasmd as dynamic as possible

x/wasm/internal/keeper/cosmwasm/contract.go
```diff
@@ -0,0 +1,35 @@
+package cosmwasm
+
+import (
+	"encoding/json"
+
+	wasmTypes "github.com/CosmWasm/go-cosmwasm/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+type OnReceiveIBCResponse struct {
```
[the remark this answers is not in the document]
alpe: 🤔 Success and Error are a bit ambiguous as an Ack can contain an application level error message.

x/wasm/internal/keeper/cosmwasm/contract.go
alpe: I have used this for `OnChanOpenInit` `OnChanOpenTry` but channel handshake was only touched in this spike.
My idea was to have the contracts engaged here as well. For example if a contract is end-of-life (burned, escrow released,...) it would make sense to not accept any new channels.

x/wasm/internal/keeper/cosmwasm/query.go
alpe: Not sure if this is a useful scenario but I thought a contract may want to query all it's channels and broadcast to it's counter parties. Or may close them all (end-of-life scenario)

x/wasm/internal/keeper/ibc.go
alpe: >  You can get the same info from an IBCQuery 
With a query I would get the channel date but I can only act on some other message that would be received later.  This callback on the other hand does not need an additional message. I would rather see this as an extension point to let contracts implement their own behaviour and state changes.

x/wasm/internal/keeper/ibc.go
alpe: I have added this to the [environment](https://github.com/CosmWasm/wasmd/blob/alex_ibc_spike2/x/wasm/internal/keeper/cosmwasm/env.go) for the contract. 
There are also a couple of message related information that are passed there, too.

x/wasm/internal/keeper/ibc.go
alpe: The original payload from the contract is passed in the `payloadData`. The contract has it's source data and the ack so that it can match it proper. 
(Assuming the data contains an unique identifier. This is also an area where the rust cosmwasm framework may provide some tooling for non natural identifiers)

alpe: The sender's channel/port are in `IBCPacketInfo` and passed via Environment

x/wasm/internal/keeper/cosmwasm/contract.go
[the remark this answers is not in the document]
ethanfrey: Exactly, the IBC/wasmd entry points make sense for the protocol level.

I would like to use more semantically interesting ones for the contracts.
And yes, I am only thinking of the interface we expose to the contracts, and figure we provide an adaptor between the official IBC and our desired API.

x/wasm/internal/keeper/ibc.go
ethanfrey: Ah... so payload data is read from the local store from the original input, and isn't something sent from the other side (that is acknowledgement)

x/wasm/internal/keeper/ibc.go
ethanfrey: Make sense. 

I wonder a bit when 90% of the time the Environment doesn't have this info.
Or do you propose two Environment types?

The original `cosmwasm.Env` that is sent for init/handle/migrate and then what you coded here (maybe rename `IBCEnvironment`), which is a superset of `cosmwasm.Env` and passed into the `ibc_handle` method for example.

This makes it clear what is called when

x/wasm/internal/keeper/ibc.go
alpe: I assume that Send and CloseChannel are possible from a contract. The `OpenChannel` process is not that clear to me! 

A channel is based on a connection and that on a client. Both are out of scope of our module so far. In GoS the relayer would establish them based on configuration. 
Technically the contract could store and provide the necessary data but calling out to other chains would mix with the relayers job and feels wrong at this stage.

from the cli:
` wasmcli tx ibc channel open-init [port-id] [channel-id] [counterparty-port-id] [counterparty-channel-id] [connection-hops] [flags]`

From the spec:
`The connectionHops stores the list of connection identifiers, in order, along which packets sent on this channel will travel. At the moment this list must be of length 1. In the future multi-hop channels may be supported.`

ethanfrey:
> I see this in the cosmwasm rust framework layer. We could provide handlers and other tooling there and keep wasmd as dynamic as possible

`wasmd` is likely the only flex point the *blockchain developer* has and the contract developer can only build upon what `cosmwasm-std` provides.

You suggest the mapping of ack -> success/error should happen inside the contract itself, not in the wasm framework?

x/wasm/internal/keeper/ibc.go
alpe: Ja, why not have different ones as we call different callbacks anyway. Good idea

The IBC PortID should be somehow available to the contract in all environments as well as IBC queries. An outgoing 
 `MsgWasmIBCCall` would start in a non IBC env. (if it works as I assume)

x/wasm/internal/keeper/ibc.go
alpe: I have not looked deep enough through the IBC impl to find where the origin message comes from in the ack process but it is available to us in the IBC callbacks.

alpe:
> You suggest the mapping of ack -> success/error should happen inside the contract itself, not in the wasm framework?

Ja, this is application layer logic in my understanding. But we can help the contract developer by providing a rust framework/toolset that comes with some defaults/ conventions to make the real contract code simple and fast to build.
If the rust contract framework provides a default ack frame then we could read from that and pass to simpler methods.
The challenge is to come up with good and reusable components. This will evolve with the use cases IMHO.

ethanfrey:
Closes in favor of #253

PR #232, opened 2020-07-27
Round up in consumeGas

assafmo:
This makes for a much more consistent behavior. Now out-of-gas inside of WASM will result in out-of-gas in cosmos instead of returning an "out of gas" string to the user.
If `gas_used == gas_limit`, cosmos doesn't throw an out of gas and instead an StdError with an "out of gas" string returns to the user.

### DIFF
--- a/x/wasm/internal/keeper/keeper.go
+++ b/x/wasm/internal/keeper/keeper.go
@@ -569,7 +569,7 @@ func gasForContract(ctx sdk.Context) uint64 {
 }
 
 func consumeGas(ctx sdk.Context, gas uint64) {
-	consumed := gas / GasMultiplier
+	consumed := (gas / GasMultiplier) + 1
 	ctx.GasMeter().ConsumeGas(consumed, "wasm contract")
 	// throw OutOfGas error if we ran out (got exactly to zero due to better limit enforcing)
 	if ctx.GasMeter().IsOutOfGas() {


ethanfrey:
This is not a "round up" but a "+1"

There are two issues you are concerned about:

(1) we spend exactly the amount of gas requested (cuz wasmer shut us down) - we should get an OutOfGas panic. This is now handled directly below your change with the lines

```go
	// throw OutOfGas error if we ran out (got exactly to zero due to better limit enforcing)
	if ctx.GasMeter().IsOutOfGas() {
              panic(sdk.ErrorOutOfGas{"Wasmer function execution"})
        }
```

(2) we spend 8098 wasmer gas and only 80 sdk gas is charged, not 81. This issue still exists, but a simple +1 doesn't solve it. If you find this important, we can add it. But please check that `gas % gasMultiplier != 0` before adding 1, if you want it to be called "round up"

assafmo:
You're right. It's oversimplified as 1 gas unit wouldn't hurt anybody but will solve the out of gas not throwing panic issue.
I think doing a +1 here obsoletes the check one line below and fixes the issue of not throwing out of gas because wasmer cuts you off.
Your call.

ethanfrey:
>  1 gas unit wouldn't hurt anybody

I don't know. I try not to add random +1s. I can accept a proper "round up on divide" logic or even "round off: +1 if gas % GasMulitiplier >= GasMultiplier/2". But I have seen random +1 thrown in to fix symptoms and hide deeper issues in plenty of code

assafmo:
Yeah I generally agree with you.
It's good to be verbose sometimes.

My only issue is the implementation of the `GasMeter`.
When you consume gas that results in `consumed == inlimit`, it doesn't throw: https://github.com/cosmos/cosmos-sdk/blob/18890a225b46260a9adc587be6fa1cc2aff101cd/store/types/gas.go#L86 (because overflow is false)
But than when you check `.IsOutOfGas()` you get true because `consumed >= limit` 🤷: https://github.com/cosmos/cosmos-sdk/blob/18890a225b46260a9adc587be6fa1cc2aff101cd/store/types/gas.go#L102

assafmo:
Closing this PR because I agree that the issue is already fixed (and I also see that the `cpuloop` test is fixed and now checks for panic instead of a returned error). ❤️

ethanfrey:
> My only issue is the implementation of the GasMeter.

I agree this is inconsistent. Want to make an issue on cosmos-sdk?

assafmo:
Yeah, should we send a PR?

PR #233, opened 2020-07-28
Restrict querier redispatched gas

ethanfrey:
Closes https://github.com/CosmWasm/cosmwasm/issues/456

So far this just adds a test to see when the recursive queries stop. I am not sure how to assert the depth that it hits before
the panic, as the output is panic in both times. The only difference is if it calls Query 5 times or 23 times. 

@alpe Any ideas how to test this better?

### DIFF
--- a/x/wasm/internal/keeper/recurse_test.go
+++ b/x/wasm/internal/keeper/recurse_test.go
@@ -314,7 +314,8 @@ func TestLimitRecursiveQueryGas(t *testing.T) {
 			// if we expect out of gas, make sure this panics
 			if tc.expectOutOfGas {
 				require.Panics(t, func() {
-					_, _ = keeper.QuerySmart(ctx, contractAddr, msg)
+					_, err := keeper.QuerySmart(ctx, contractAddr, msg)
+					t.Logf("Got error not panic: %#v", err)
 				})
 				assert.Equal(t, tc.expectQueriesFromContract, totalWasmQueryCounter)
 				return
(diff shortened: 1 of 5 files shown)


x/wasm/internal/keeper/recurse_test.go
```diff
@@ -243,3 +224,88 @@ func TestGasOnExternalQuery(t *testing.T) {
...
+			msg: Recurse{
+				Depth: 5,
+				Work:  2000,
+			},
+			expectedGas: GasWork2k + 5*(GasWork2k+GasReturnHashed),
+		},
+		// this is where we expect an error...
+		// it has enough gas to run 4 times and die on the 5th
+		// however, if we don't charge the cpu gas before sub-dispatching, we can recurse over 20 times
+		// TODO: figure out how to asset how deep it went
```
ethanfrey, line 268: This is what needs to be tested... any ideas how?

Once I get this failing (more than 5 levels), then I can work on fixing the go-cosmwasm levels and proving this worked.

alpe: 🤔  You could modify the `Querier` with a custom `QueryPlugin` and increment a counter on each call so that you would know how often it was called.

ethanfrey: That's an interesting idea.
I was thinking to add some global stats counter to the keeper, but that seems cleaner if possible

ethanfrey: I started this approach here: https://github.com/CosmWasm/wasmd/pull/234 but it seems to require many deeper changes

ethanfrey: Oof, my idea also requires making it a pointer, seems like I have to suck that up and do it.

PR #234, opened 2020-07-28
Attempt counting query plugin

ethanfrey:
I tried to modify the QueryPlugins, which worked, except we only use `Keeper`, so it only affects the first level. To do this properly (and maybe that would be a good thing), we need to use `*Keeper`, but this was causing lots of changes, so I wonder if this is the best approach.


ethanfrey:
Discarded for a much more clever solution from @alpe

PR #235, opened 2020-07-28
Move constants to params

### LINKED ISSUE #202: Move constants to params
## Summary
We have a couple of hard coded values that could be moved into params so that we can update them without deployment. 
An incomplete list:

- max contract size
- gas multiplier
- instance cost
- compile cost
...

____

#### For Admin Use

- [ ] Not duplicate issue

alpe:
🚧  WIP

Resolves #202

In Params now: 
* GasMultiplier
* InstanceCost
* CompileCost

NOTE: the changes come with a significant increase in gas costs  (in our test setup). The work is on hold for >0.10

TODO:
- move other constants
- proper tests

### DIFF
--- a/x/wasm/alias.go
+++ b/x/wasm/alias.go
@@ -23,7 +23,6 @@ const (
 	MaxBuildTagSize               = types.MaxBuildTagSize
 	CustomEventType               = types.CustomEventType
 	AttributeKeyContractAddr      = types.AttributeKeyContractAddr
-	GasMultiplier                 = keeper.GasMultiplier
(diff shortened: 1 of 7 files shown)


ethanfrey:
I would leave this on hold for a bit to think if we want to. The gas costs are an interesting factor (5x reads?)

It would need to be for 0.11 in any case.

ethanfrey:
Nice idea. Let's not do this for 1.0. Maybe 2.0

ethanfrey:
Do not delete branch

PR #238, opened 2020-07-30
Suffix testnet specific client executables with cli

orkunkl:
Let's keep the convention on...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,8 @@ Ref: https://keepachangelog.com/en/1.0.0/
 
 ## [Unreleased]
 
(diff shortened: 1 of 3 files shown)


ethanfrey:
I had a discussion with simon on this

I find the cli very ugly. And not used anywhere else 

Mongod and mongocli?

Sshd and sshcli?

Psqld and psqlcli?

I keep it for wasmd as that is familiar, but I prefer nicer names for our testnet. Do you see a big issue with the names? The seem nicer to me

orkunkl:
If the choice is intentional, no need to revert. I thought it got slipped somehow.

PR #246, opened 2020-08-06
Support gaps in query code list

alpe:
With the genesis import we allow gaps in the codeID range. Before the query handler would return early on a gap but now it will handle gaps and return all elements.


x/wasm/internal/keeper/querier.go
```diff
@@ -205,22 +205,16 @@ type ListCodeResponse struct {
...
 	var info []ListCodeResponse
-
-	var i uint64
-	for true {
-		i++
-		res := keeper.GetCodeInfo(ctx, i)
-		if res == nil {
-			break
-		}
+	keeper.IterateCodeInfos(ctx, func(i uint64, res types.CodeInfo) bool {
```
ethanfrey, line 208: Ah, and the function was already there.

PR #249, opened 2020-08-07
0.9.1 to cosmos stargate cli fixes

alpe:
* Fix wasmcli + wasmd cli context
* Upgrade sdk to `3322e26`

The encoding issues on `wasmd start` are not solved with this, yet. Waiting for sdk updates to apply/ adopt them.


app/app.go
```diff
@@ -265,7 +265,7 @@ func NewWasmApp(
 	// TODO: remove amino codec dependency once Tendermint version is upgraded with
```
ethanfrey, line 265: Can this todo go away? Or is appCodec still amino?

(having app.cdc and appCodec was super confusing)

app/encoding.go
```diff
@@ -1,14 +1,15 @@
...
-	"github.com/cosmos/cosmos-sdk/simapp"
 	simparams "github.com/cosmos/cosmos-sdk/simapp/params"
+	"github.com/cosmos/cosmos-sdk/std"
 )
 
 func MakeEncodingConfig() simparams.EncodingConfig {
-	encodingConfig := simapp.MakeEncodingConfig() // todo: this is the simapp !!!
-	wasm.RegisterCodec(encodingConfig.Amino)
-	wasm.RegisterInterfaces(encodingConfig.InterfaceRegistry)
+	encodingConfig := simparams.MakeEncodingConfig() // todo: this is the simapp !!!
```
ethanfrey, line 9: Should we not copy the type and the `simparams.MakeEncodingConfig()` ?

Also, this no longer registered all modules in the simapp (so ModuleBasics won't cause conflicts)? If tests pass, then this looks good, but seems a bit odd... I'm almost prefer just copying the simappparams code here so the whole flow is clear.

cmd/wasmcli/main.go
```diff
@@ -43,7 +44,7 @@ func main() {
 		WithInput(os.Stdin).
 		WithAccountRetriever(authtypes.NewAccountRetriever(encodingConfig.Marshaler)).
 		WithBroadcastMode(flags.BroadcastBlock).
-		WithHomeDir(app.DefaultNodeHome)
+		WithHomeDir(app.DefaultCLIHome)
```
ethanfrey, line 47: Very good call.

ethanfrey: Is this a bug/issue in master as well?

docker/setup.sh
```diff
@@ -19,5 +19,5 @@ for addr in "$@"; do
   wasmd add-genesis-account "$addr" "1000000000$STAKE,1000000000$FEE"
 done
 # submit a genesis validator tx
-(echo "$PASSWORD"; echo "$PASSWORD"; echo "$PASSWORD") | wasmd gentx --name validator --amount "250000000$STAKE"
+(echo "$PASSWORD"; echo "$PASSWORD"; echo "$PASSWORD") | wasmd gentx validator --chain-id=testing --amount "250000000$STAKE"
```
ethanfrey, line 22: This works now?

alpe: The `gentx` does work. It is the `start` command that fails

app/encoding.go
alpe: agree, I will copy the code

cmd/wasmcli/main.go
alpe: No, that was me migrating the client context 🙈

alpe:
> This works to launch an app? Or is there an issue in genesis parsing still? 

Unfortunately not. It still fails with `wasmd start`.  I will add a TODO

PR #257, opened 2020-08-20
module manager had evidence module added twice.

puneet2019:


alpe:
💐 Good eye. Thanks for your help!

PR #258, opened 2020-08-20
Propose contract methods

alpe:
Design goals:
- close to other cosmwasm signatures like a return type per method
- Use of Environment for IBC related metadata

### DIFF
--- a/x/wasm/IBC.md
+++ b/x/wasm/IBC.md
@@ -178,96 +178,98 @@ type IBCCloseChannel struct {
 type Version string
 ```
 
-Packet callbacks:
+Channel andPacket callbacks to the wasm contract
 
 ```go
-package packets
 
-// for reference: this is more like what we pass to go-cosmwasm
-// func (c *mockContract) OnReceive(params cosmwasm2.Env, msg []byte, store prefix.Store, api cosmwasm.GoAPI, 
-//         querier keeper.QueryHandler, meter sdk.GasMeter, gas uint64) (*cosmwasm2.OnReceiveIBCResponse, uint64, error) {}
-// below is how we want to expose it in x/wasm:
+// IBCContractCallbacks defines the methods for go-cosmwasm to interact with the wasm contract.
+// A mock contract would implement the interface to fully simulate a wasm contract's behaviour.
+type IBCContractCallbacks interface {
+  // Package livecycle
 
-// IBCRecvPacket is called when we receive a packet sent from the other end of a channel
-// The response bytes listed here will be returned to the caller as "result" in IBCPacketAck
-// What do we do with error?
-//
-// If we were to assume/enforce an envelope, then we could wrap response/error into the acknowledge packet,
-// but we delegated that encoding to the contract
-func IBCRecvPacket(ctx sdk.Context, k *wasm.Keeper, contractAddress sdk.AccAddress, env IBCPacketInfo, msg []byte) 
-    (response []byte, err error) {}
+  // OnIBCPacketReceive handles an incoming IBC package
+  OnIBCPacketReceive(hash []byte, params cosmwasm.Env, msg []byte, store prefix.Store, api wasm.GoAPI, querier QueryHandler, meter sdk.GasMeter, gas uint64) (*cosmwasm.IBCPacketReceiveResponse, uint64, error)
+  // OnIBCPacketAcknowledgement handles a IBC package execution on the counterparty chain
+  OnIBCPacketAcknowledgement(hash []byte, params cosmwasm.Env, originalData []byte, acknowledgement []byte, store prefix.Store, api wasm.GoAPI, querier QueryHandler, meter sdk.GasMeter, gas uint64) (*cosmwasm.IBCPacketAcknowledgementResponse, uint64, error)
+  // OnIBCPacketTimeout reverts state when the IBC package execution does not come in time
+  OnIBCPacketTimeout(hash []byte, params cosmwasm.Env, msg []byte, store prefix.Store, api wasm.GoAPI, querier QueryHandler, meter sdk.GasMeter, gas uint64) (*cosmwasm.IBCPacketTimeoutResponse, uint64, error)
+  // channel livecycle
 
-// how to handle error here? if we push it up the ibc stack and fail the transaction (normal handling),
-// the packet may be posted again and again. just log and ignore failures here? what does a failure even mean?
-// only realistic one I can imagine is OutOfGas (panic) to retry with higher gas limit.
-//
-// if there any point in returning a response, what does it mean?
-func IBCPacketAck(ctx sdk.Context, k *wasm.Keeper, contractAddress sdk.AccAddress, env IBCPacketInfo, 
-    originalMsg []byte, result []byte) error {}
+  // OnIBCChannelOpen does the protocol version negotiation during channel handshake phase
+  OnIBCChannelOpen(hash []byte, params cosmwasm.Env, order channeltypes.Order, version string, store prefix.Store, api wasm.GoAPI, querier QueryHandler, meter sdk.GasMeter, gas uint64) (*cosmwasm.IBCChannelOpenResponse, uint64, error)
+  // OnIBCChannelConnect callback when a IBC channel is established
+  OnIBCChannelConnect(hash []byte, params cosmwasm.Env, counterpartyPortID, counterpartyChannelID string, store prefix.Store, api wasm.GoAPI, querier QueryHandler, meter sdk.GasMeter, gas uint64) (*cosmwasm.IBCChannelConnectResponse, uint64, error)
+  // OnIBCChannelConnect callback when a IBC channel is closed
+  OnIBCChannelClose(ctx sdk.Context, hash []byte, params cosmwasm.Env, counterpartyPortID, counterpartyChannelID string, store prefix.Store, api wasm.GoAPI, querier QueryHandler, meter sdk.GasMeter, gas uint64) (*cosmwasm.IBCChannelCloseResponse, uint64, error)
+}
 
-// same question as for IBCPacketAck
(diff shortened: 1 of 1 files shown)


x/wasm/IBC.md
```diff
@@ -178,96 +178,98 @@ type IBCCloseChannel struct {
...
+  Log []wasmTypes.LogAttribute `json:"log"`
+}
+
+type IBCChannelCloseResponse struct {
+  Messages []sdk.Msg `json:"messages"`
+  Log []wasmTypes.LogAttribute `json:"log"`
+}
+
+// Environment data for a contract
+type Env struct {
```
ethanfrey, line 246: Okay, this is supposed to extend the normla cosmwasm Env with this extra info?

x/wasm/IBC.md
```diff
@@ -178,96 +178,98 @@ type IBCCloseChannel struct {
...
+  IBC *IBCInfo `json:"ibc,omitempty"`
 }
+
+type IBCInfo struct {
+  // Port of the contract
+  Port string `json:"port"`
+  // Channel to the contract
+  Channel string         `json:"channel"`
+  // Optional packet meta data when called on IBC packet functions
+  Packet  *IBCPacketInfo `json:"packet,omitempty"`
```
ethanfrey, line 258: 3 levels Env -> IBCInfo -> IBCPacketInfo seems overly complex. I had a much simpler version that covers most of this.

Let's flatten it, and we don't need both (Port, Channel) and (SourcePort, SourceChannel). Just show the local side and they can use `GetCounterParty` if they need it. Let's trim this to the essential

x/wasm/IBC.md
```diff
@@ -178,96 +178,98 @@ type IBCCloseChannel struct {
...
+  Messages []sdk.Msg `json:"messages"`
+  Log []wasmTypes.LogAttribute `json:"log"`
 }
 
-type ChannelClosedInfo struct {
-    // local id for the Channel that is being shut down
-    ChannelID string
+type IBCChannelOpenResponse struct {
+  // Result contains a boolean if the channel would be accepted
+  Result                       bool   `json:"result"`
```
ethanfrey, line 230: I would use `Accepted` or `Approved` here. Result implies something more than a boolean, with real data

ethanfrey: Maybe:

```
struct {
  Success bool
  Error string
}
```

Where Error is only set if Success is false

x/wasm/IBC.md
```diff
@@ -178,96 +178,98 @@ type IBCCloseChannel struct {
...
+  // OnIBCPacketAcknowledgement handles a IBC package execution on the counterparty chain
+  OnIBCPacketAcknowledgement(hash []byte, params cosmwasm.Env, originalData []byte, acknowledgement []byte, store prefix.Store, api wasm.GoAPI, querier QueryHandler, meter sdk.GasMeter, gas uint64) (*cosmwasm.IBCPacketAcknowledgementResponse, uint64, error)
+  // OnIBCPacketTimeout reverts state when the IBC package execution does not come in time
+  OnIBCPacketTimeout(hash []byte, params cosmwasm.Env, msg []byte, store prefix.Store, api wasm.GoAPI, querier QueryHandler, meter sdk.GasMeter, gas uint64) (*cosmwasm.IBCPacketTimeoutResponse, uint64, error)
+  // channel livecycle
+
+  // OnIBCChannelOpen does the protocol version negotiation during channel handshake phase
+  OnIBCChannelOpen(hash []byte, params cosmwasm.Env, order channeltypes.Order, version string, store prefix.Store, api wasm.GoAPI, querier QueryHandler, meter sdk.GasMeter, gas uint64) (*cosmwasm.IBCChannelOpenResponse, uint64, error)
```
ethanfrey, line 199: If we stay closer to the go-cosmwasm side, then let's wrap `order channeltypes.Order, version string` into a struct (all info will pass in som json object, not lots of different args, doesn't work well with wasm)

x/wasm/IBC.md
```diff
@@ -178,96 +178,98 @@ type IBCCloseChannel struct {
...
+  // OnIBCChannelConnect callback when a IBC channel is established
+  OnIBCChannelConnect(hash []byte, params cosmwasm.Env, counterpartyPortID, counterpartyChannelID string, store prefix.Store, api wasm.GoAPI, querier QueryHandler, meter sdk.GasMeter, gas uint64) (*cosmwasm.IBCChannelConnectResponse, uint64, error)
```
ethanfrey, line 201: Same as above (wrap in stuct): `counterpartyPortID, counterpartyChannelID string` and please add the final `version` string here.. the second side may change it, so the original side must get that info at the end. And the `ordering` enum as well - so we have all info in one place when confirming the connection

x/wasm/IBC.md
```diff
@@ -178,96 +178,98 @@ type IBCCloseChannel struct {
...
+  // OnIBCChannelConnect callback when a IBC channel is closed
+  OnIBCChannelClose(ctx sdk.Context, hash []byte, params cosmwasm.Env, counterpartyPortID, counterpartyChannelID string, store prefix.Store, api wasm.GoAPI, querier QueryHandler, meter sdk.GasMeter, gas uint64) (*cosmwasm.IBCChannelCloseResponse, uint64, error)
```
ethanfrey, line 203: `counterpartyPortID, counterpartyChannelID string` do we need these? We get our own side in Env and can query counterParty if needed. We should just use the channelID locally to look up everything.

I can see passing this explicitly in `OnIBCChannelConnect` to provide all relevant info on connect to be use for setup if needed. But once it is set up, do we usually care? And if so we can query

x/wasm/IBC.md
ethanfrey: Okay, this is passed on all calls, connections as well as packets, which is why you have the PacketInfo optional.

This is getting confusing. I would like things always there or not - clear at compile time - not based on out-of-bands understanding

Maybe just wrap the Packet Info around the Msg

x/wasm/IBC.md
```diff
@@ -178,96 +178,98 @@ type IBCCloseChannel struct {
...
-	
-    // Note: Timeout if guaranteed ONLY to be exceeded iff we are processing IBCPacketDropped
-    // otherwise, this is just interesting metadata
+// IBCContractCallbacks defines the methods for go-cosmwasm to interact with the wasm contract.
+// A mock contract would implement the interface to fully simulate a wasm contract's behaviour.
+type IBCContractCallbacks interface {
+  // Package livecycle
+
+  // OnIBCPacketReceive handles an incoming IBC package
+  OnIBCPacketReceive(hash []byte, params cosmwasm.Env, msg []byte, store prefix.Store, api wasm.GoAPI, querier QueryHandler, meter sdk.GasMeter, gas uint64) (*cosmwasm.IBCPacketReceiveResponse, uint64, error)
```
ethanfrey, line 191: See comments at the end, please use this instead of `msg`

```go
type IBCPacket struct {
  Msg []byte
  Sequence uint64
  TimeoutHeight uint64
  // block timestamp (in nanoseconds) after which the packet times out
  TimeoutTimestamp uint64
}
```

x/wasm/IBC.md
```diff
@@ -178,96 +178,98 @@ type IBCCloseChannel struct {
...
+  // OnIBCPacketAcknowledgement handles a IBC package execution on the counterparty chain
+  OnIBCPacketAcknowledgement(hash []byte, params cosmwasm.Env, originalData []byte, acknowledgement []byte, store prefix.Store, api wasm.GoAPI, querier QueryHandler, meter sdk.GasMeter, gas uint64) (*cosmwasm.IBCPacketAcknowledgementResponse, uint64, error)
+  // OnIBCPacketTimeout reverts state when the IBC package execution does not come in time
+  OnIBCPacketTimeout(hash []byte, params cosmwasm.Env, msg []byte, store prefix.Store, api wasm.GoAPI, querier QueryHandler, meter sdk.GasMeter, gas uint64) (*cosmwasm.IBCPacketTimeoutResponse, uint64, error)
```
ethanfrey, line 195: Maybe same as `IBCPacket` above, or maybe more info?

x/wasm/IBC.md
```diff
@@ -178,96 +178,98 @@ type IBCCloseChannel struct {
...
-    // otherwise, this is just interesting metadata
+// IBCContractCallbacks defines the methods for go-cosmwasm to interact with the wasm contract.
+// A mock contract would implement the interface to fully simulate a wasm contract's behaviour.
+type IBCContractCallbacks interface {
+  // Package livecycle
+
+  // OnIBCPacketReceive handles an incoming IBC package
+  OnIBCPacketReceive(hash []byte, params cosmwasm.Env, msg []byte, store prefix.Store, api wasm.GoAPI, querier QueryHandler, meter sdk.GasMeter, gas uint64) (*cosmwasm.IBCPacketReceiveResponse, uint64, error)
+  // OnIBCPacketAcknowledgement handles a IBC package execution on the counterparty chain
+  OnIBCPacketAcknowledgement(hash []byte, params cosmwasm.Env, originalData []byte, acknowledgement []byte, store prefix.Store, api wasm.GoAPI, querier QueryHandler, meter sdk.GasMeter, gas uint64) (*cosmwasm.IBCPacketAcknowledgementResponse, uint64, error)
```
ethanfrey, line 193: See comments at the end, please use this for the originalData and acknowledgement

```go
type IBCPacket struct {
  OriginalData []byte
  Acknowledgement []byte
  Sequence uint64
  TimeoutHeight uint64
  // block timestamp (in nanoseconds) after which the packet times out
  TimeoutTimestamp uint64
}
```

ethanfrey (commented):
Made a number of comments on the API details.

We cannot pass tons of arguements to wasm contracts. We need to wrap things into structs we json-encode and send down the line. Made a number of suggestions to improve the API

Issue #255, opened 2020-08-19
Complete IBC Mock testing


ethanfrey:
This builds on what is done in #253 and adds some simpler demos as defined in our milestone:

Add a mockHandler that plays ping-pong:
  * version is "ping" or "pong" (both acceptable)
  * on version negotiation the "Try" side will switch the version (if "ping" requested, counterparty does "pong" and visa versa)
  * stores a counter in state (starts at 0)
  * Packets are JSON, only valid ones look like `{"ping": 5}`
  * Receiver will add the value to the local state and return acknowledgement
  * Acknowledgement is either `{"error": "message here"}` or `{"success": {"pong": 14}}` or whatever the updated state is
  * On receiving acknowledgement, if a success, then that "remoteState" is stored. If error, increment "numErrors"

This handler has "localState" and "remoteState". Receiving a "ping" packet will increment "localState". Getting a "pong" acknowlegdement will set "remoteState".

Once this is build, a few scenarios:
  * Proper connection (show version is swapped from original request), sending packets with coordinators, it works well
  * Send `{"random": "data"}` packet and verify we get an `"error"` acknowledgement back (numErrors == 1)
  * Try connecting with invalid version (transfer) and show channel will not get established

PR #259, opened 2020-08-22
Add IBC Ping Pong contract demo

alpe:
* Adds `relay_pingpong_test.go` as happy path for #255 
* Contains contract interface redesign and 🐛 fixes that I had started already

### DIFF
--- a/x/wasm/ibc.go
+++ b/x/wasm/ibc.go
@@ -29,9 +29,11 @@ func (i IBCHandler) OnChanOpenInit(ctx sdk.Context, order channeltypes.Order, co
 		return sdkerrors.Wrapf(err, "contract port id")
 	}
 
-	err = i.keeper.OnOpenChannel(ctx, contractAddr, order, version, connectionHops, cosmwasm.IBCInfo{
-		Port:    portID,
-		Channel: channelID,
+	err = i.keeper.OnOpenChannel(ctx, contractAddr, cosmwasm.IBCChannel{
+		Endpoint:             cosmwasm.IBCEndpoint{Port: portID, Channel: channelID},
+		CounterpartyEndpoint: cosmwasm.IBCEndpoint{Port: counterParty.PortId, Channel: counterParty.ChannelId},
+		Order:                order,
+		Version:              version,
 	})
 	if err != nil {
 		return err
@@ -50,9 +52,12 @@ func (i IBCHandler) OnChanOpenTry(ctx sdk.Context, order channeltypes.Order, con
 		return sdkerrors.Wrapf(err, "contract port id")
 	}
 
-	err = i.keeper.OnOpenChannel(ctx, contractAddr, order, version, connectionHops, cosmwasm.IBCInfo{
-		Port:    portID,
-		Channel: channelID,
+	err = i.keeper.OnOpenChannel(ctx, contractAddr, cosmwasm.IBCChannel{
+		Endpoint:             cosmwasm.IBCEndpoint{Port: portID, Channel: channelID},
+		CounterpartyEndpoint: cosmwasm.IBCEndpoint{Port: counterParty.PortId, Channel: counterParty.ChannelId},
+		Order:                order,
+		Version:              version,
+		CounterpartyVersion:  &counterpartyVersion,
 	})
 	if err != nil {
 		return err
@@ -73,9 +78,12 @@ func (i IBCHandler) OnChanOpenAck(ctx sdk.Context, portID, channelID string, cou
 	if !ok {
 		return sdkerrors.Wrap(types.ErrInvalidCounterparty, "not found")
 	}
-	return i.keeper.OnConnectChannel(ctx, contractAddr, channelInfo.Counterparty, counterpartyVersion, cosmwasm.IBCInfo{
-		Port:    portID,
-		Channel: channelID,
+	return i.keeper.OnConnectChannel(ctx, contractAddr, cosmwasm.IBCChannel{
+		Endpoint:             cosmwasm.IBCEndpoint{Port: portID, Channel: channelID},
+		CounterpartyEndpoint: cosmwasm.IBCEndpoint{Port: channelInfo.Counterparty.PortId, Channel: channelInfo.Counterparty.ChannelId},
+		Order:                channelInfo.Ordering,
+		Version:              channelInfo.Version,
+		CounterpartyVersion:  &counterpartyVersion,
 	})
 }
 
@@ -88,9 +96,11 @@ func (i IBCHandler) OnChanOpenConfirm(ctx sdk.Context, portID, channelID string)
 	if !ok {
 		return sdkerrors.Wrap(types.ErrInvalidCounterparty, "not found")
 	}
-	return i.keeper.OnConnectChannel(ctx, contractAddr, channelInfo.Counterparty, channelInfo.Version, cosmwasm.IBCInfo{
-		Port:    portID,
-		Channel: channelID,
+	return i.keeper.OnConnectChannel(ctx, contractAddr, cosmwasm.IBCChannel{
+		Endpoint:             cosmwasm.IBCEndpoint{Port: portID, Channel: channelID},
+		CounterpartyEndpoint: cosmwasm.IBCEndpoint{Port: channelInfo.Counterparty.PortId, Channel: channelInfo.Counterparty.ChannelId},
+		Order:                channelInfo.Ordering,
+		Version:              channelInfo.Version,
 	})
 }
 
@@ -116,7 +126,7 @@ func (i IBCHandler) OnRecvPacket(ctx sdk.Context, packet channeltypes.Packet) (*
 	if err != nil {
 		return nil, nil, sdkerrors.Wrapf(err, "contract port id")
 	}
-	msgBz, err := i.keeper.OnRecvPacket(ctx, contractAddr, packet.Data, ibcInfoFromPacket(packet))
+	msgBz, err := i.keeper.OnRecvPacket(ctx, contractAddr, newIBCPacket(packet))
 	if err != nil {
 		return nil, nil, err
 	}
@@ -140,12 +150,15 @@ func (i IBCHandler) OnRecvPacket(ctx sdk.Context, packet channeltypes.Packet) (*
 }
 
 func (i IBCHandler) OnAcknowledgementPacket(ctx sdk.Context, packet channeltypes.Packet, acknowledgement []byte) (*sdk.Result, error) {
-	contractAddr, err := ContractFromPortID(packet.DestinationPort)
+	contractAddr, err := ContractFromPortID(packet.SourcePort)
 	if err != nil {
 		return nil, sdkerrors.Wrapf(err, "contract port id")
 	}
 
-	err = i.keeper.OnAckPacket(ctx, contractAddr, packet.Data, acknowledgement, ibcInfoFromPacket(packet))
+	err = i.keeper.OnAckPacket(ctx, contractAddr, cosmwasm.IBCAcknowledgement{
+		Acknowledgement: acknowledgement,
+		OriginalPacket:  newIBCPacket(packet),
+	})
 	if err != nil {
 		return nil, err
 	}
@@ -180,7 +193,7 @@ func (i IBCHandler) OnTimeoutPacket(ctx sdk.Context, packet channeltypes.Packet)
 	if err != nil {
 		return nil, sdkerrors.Wrapf(err, "contract port id")
 	}
-	err = i.keeper.OnTimeoutPacket(ctx, contractAddr, packet.Data, ibcInfoFromPacket(packet))
+	err = i.keeper.OnTimeoutPacket(ctx, contractAddr, newIBCPacket(packet))
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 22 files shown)


x/wasm/ibc.go
```diff
@@ -29,9 +29,11 @@ func (i IBCHandler) OnChanOpenInit(ctx sdk.Context, order channeltypes.Order, co
 		return sdkerrors.Wrapf(err, "contract port id")
 	}
 
-	err = i.keeper.OnOpenChannel(ctx, contractAddr, order, version, connectionHops, cosmwasm.IBCInfo{
-		Port:    portID,
-		Channel: channelID,
+	err = i.keeper.OnOpenChannel(ctx, contractAddr, cosmwasm.IBCChannel{
```
ethanfrey, line 32: Nice

x/wasm/ibc_testing/chain.go
```diff
@@ -334,10 +334,11 @@ func (chain *TestChain) AddTestConnection(clientID, counterpartyClientID string)
 // format:
 // connectionid<index>
 func (chain *TestChain) ConstructNextTestConnection(clientID, counterpartyClientID string) *TestConnection {
-	connectionID := ConnectionIDPrefix + strconv.Itoa(len(chain.Connections))
+	connectionID := chain.ChainID + ConnectionIDPrefix + strconv.Itoa(len(chain.Connections))
```
ethanfrey, line 337: ChainID is here just to make them unique for debugging tests, right? And not needed by the protocol?

If they have to be globally unique over all chains, good to document that

ethanfrey: (Good to ensure difference in test, so we don't accidentally mix up local and remote ids)

x/wasm/internal/keeper/cosmwasm/contract.go
```diff
@@ -1,42 +1,74 @@
...
+	Messages []CosmosMsg `json:"messages,omitempty"`
 	// Log contains event attributes to expose over abci interface
-	Log []wasmTypes.LogAttribute `json:"log,omitempty"`
+	Log []cosmwasmv1.LogAttribute `json:"log,omitempty"`
 }
 
 type IBCPacketAcknowledgementResponse struct {
-	Messages []sdk.Msg                `json:"messages"`
-	Log      []wasmTypes.LogAttribute `json:"log"`
+	Messages []CosmosMsg               `json:"messages"`
```
ethanfrey, line 50: Ah, nice naming here, makes it clear which ones are the same as current, which are new... Like the new CosmosMsg type

x/wasm/internal/keeper/cosmwasm/msg.go
```diff
@@ -0,0 +1,56 @@
...
+	Custom  json.RawMessage        `json:"custom,omitempty"`
+	Staking *cosmwasmv1.StakingMsg `json:"staking,omitempty"`
+	Wasm    *cosmwasmv1.WasmMsg    `json:"wasm,omitempty"`
+	IBC     *IBCMsg                `json:"wasm_ibc,omitempty"`
+}
+
+type IBCMsg struct {
+	SendPacket   *IBCSendMsg         `json:"execute,omitempty"`
+	CloseChannel *IBCCloseChannelMsg `json:"instantiate,omitempty"`
+	Transfer     *IBCTransferMsg     `json:"instantiate,omitempty"`
```
ethanfrey, line 22: Transfer is for the ics20 / ibc-transfer module? Or something else?

Can you add a comment there for the purpose? SendMsg and CloseChannel are clear. And if you want to add in OpenChannel you can (I decided you were right and just put that responsibility back to the relayer)

x/wasm/internal/keeper/handler_plugin.go
```diff
@@ -4,12 +4,16 @@ import (
 	"encoding/json"
 	"fmt"
 
-	wasmTypes "github.com/CosmWasm/go-cosmwasm/types"
+	cosmwasmv1 "github.com/CosmWasm/go-cosmwasm/types"
+	cosmwasmv2 "github.com/CosmWasm/wasmd/x/wasm/internal/keeper/cosmwasm"
```
ethanfrey, line 8: Nice

x/wasm/internal/keeper/handler_plugin.go
```diff
@@ -19,32 +23,46 @@ type MessageHandler struct {
...
+	}
+}
+
+type BankEncoder func(sender sdk.AccAddress, msg *cosmwasmv1.BankMsg) ([]sdk.Msg, error)
 type CustomEncoder func(sender sdk.AccAddress, msg json.RawMessage) ([]sdk.Msg, error)
-type StakingEncoder func(sender sdk.AccAddress, msg *wasmTypes.StakingMsg) ([]sdk.Msg, error)
-type WasmEncoder func(sender sdk.AccAddress, msg *wasmTypes.WasmMsg) ([]sdk.Msg, error)
+type StakingEncoder func(sender sdk.AccAddress, msg *cosmwasmv1.StakingMsg) ([]sdk.Msg, error)
+type WasmEncoder func(sender sdk.AccAddress, msg *cosmwasmv1.WasmMsg) ([]sdk.Msg, error)
+type IBCEncoder func(ctx sdk.Context, sender sdk.AccAddress, source cosmwasmv2.IBCEndpoint, msg *cosmwasmv2.IBCMsg) ([]sdk.Msg, error)
```
ethanfrey, line 45: This is all the extra info we need? Not an external keeper?

Nice to make this simple as possible

x/wasm/internal/keeper/handler_plugin.go
```diff
@@ -19,32 +23,46 @@ type MessageHandler struct {
...
-func DefaultEncoders() MessageEncoders {
-	return MessageEncoders{
+func DefaultEncoders(channelKeeper types.ChannelKeeper, capabilityKeeper types.CapabilityKeeper) MessageEncoders {
+	e := MessageEncoders{
 		Bank:    EncodeBankMsg,
 		Custom:  NoCustomMsg,
 		Staking: EncodeStakingMsg,
 		Wasm:    EncodeWasmMsg,
 	}
+	if channelKeeper != nil { // todo: quick hack to keep tests happy
```
ethanfrey, line 62: I like this, so it is only there and needed for the IBC tests and you don't have to rebuild everything

x/wasm/internal/keeper/handler_plugin.go
```diff
@@ -229,7 +267,62 @@ func EncodeWasmMsg(sender sdk.AccAddress, msg *wasmTypes.WasmMsg) ([]sdk.Msg, er
 	return nil, sdkerrors.Wrap(types.ErrInvalidMsg, "Unknown variant of Wasm")
 }
 
-func (h MessageHandler) Dispatch(ctx sdk.Context, contractAddr sdk.AccAddress, msg wasmTypes.CosmosMsg) error {
+func EncodeIBCMsg(channelKeeper types.ChannelKeeper, capabilityKeeper types.CapabilityKeeper) IBCEncoder {
```
ethanfrey, line 270: Ah, you pass the keepers in here. Okay, this makes more sense. You do add all the power into this one.

x/wasm/internal/keeper/ibc.go
```diff
@@ -69,32 +69,32 @@ type IBCContractCallbacks interface {
 	// Package livecycle
 
 	// OnIBCPacketReceive handles an incoming IBC package
-	OnIBCPacketReceive(hash []byte, params cosmwasm.Env, msg []byte, store prefix.Store, api wasm.GoAPI, querier QueryHandler, meter sdk.GasMeter, gas uint64) (*cosmwasm.IBCPacketReceiveResponse, uint64, error)
+	OnIBCPacketReceive(hash []byte, params cosmwasm.Env, packet cosmwasm.IBCPacket, store prefix.Store, api wasm.GoAPI, querier QueryHandler, meter sdk.GasMeter, gas uint64) (*cosmwasm.IBCPacketReceiveResponse, uint64, error)
```
ethanfrey, line 72: Yes, I like the cleanup here

x/wasm/internal/keeper/reflect_test.go
```diff
@@ -137,7 +137,7 @@ func TestMaskReflectContractSend(t *testing.T) {
 }
 
 func TestMaskReflectCustomMsg(t *testing.T) {
-	t.Skip("Alex: fails with `cannot protobuf JSON decode unsupported type: *types.Msg: failed to unmarshal JSON bytes`")
+	t.Skip("Alex: fails with `cannot protobuf JSON decode unsupported type: *types.Data: failed to unmarshal JSON bytes`")
```
ethanfrey, line 140: Yeah, this was a bit of an edge-case one. Leave it as skip and we address it before 2.0 with a different reflect contract likely.

ethanfrey (commented):
Start review, cover internal package.
Will do follow up review on ping_pong test

x/wasm/relay_pingpong_test.go
```diff
@@ -0,0 +1,323 @@
...
+	)
+	clientA, clientB, connA, connB := coordinator.SetupClientConnections(chainA, chainB, clientexported.Tendermint)
+	connA.NextChannelVersion = ping
+	connB.NextChannelVersion = pong
+
+	channelA, channelB := coordinator.CreateChannel(chainA, chainB, connA, connB, sourcePortID, counterpartyPortID, channeltypes.UNORDERED)
+	var err error
+
+	const startValue uint64 = 100
+	s := startGame{
```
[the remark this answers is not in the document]
ethanfrey: After talking, it seems this is what ibc-transfer does. But can we just use the lifetime callback and use startValue 0?

x/wasm/relay_pingpong_test.go
```diff
@@ -0,0 +1,323 @@
...
+	// parse received data and store
+	var receivedBall hit
+	if err := json.Unmarshal(packet.Data, &receivedBall); err != nil {
+		return &cosmwasmv2.IBCPacketReceiveResponse{
+			Acknowledgement: hitAcknowledgement{Error: err.Error()}.GetBytes(),
+			// no hit msg, we stop the game
+		}, 0, nil
+	}
+
+	otherCount := receivedBall[counterParty(p.actor)]
```
ethanfrey, line 249: Let's add a max hits (3 or 4) and if this is higher, then we return `IBCPacketReceiveResponse` with a "failed" response. In this case, we do not initiate another round. It will allow us to test error handling in acknowledge.

x/wasm/relay_pingpong_test.go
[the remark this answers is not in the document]
ethanfrey: Not a blocker if we cannot do this, but it would be nice to set up the channel info on the callback.

Then, the execute message can just contain start value (and max hits?). No need for all source and counterparty, just local channel ID (rest stored from OnIBCChannelConnect)

ethanfrey (commented):
Comments on relay test

x/wasm/ibc_testing/chain.go
alpe: yes, the intend was to make them unique for debugging when working with 2 chains. Protocol wise they don't have to be globally unique

x/wasm/internal/keeper/handler_plugin.go
alpe: `ChannelKeeper` and `CapabilityKeeper` were passed to the constructor.

x/wasm/relay_pingpong_test.go
```diff
@@ -0,0 +1,323 @@
...
+
+	wasmkeeper.MockContracts[pingContractAddr.String()] = pingContract
+	wasmkeeper.MockContracts[pongContractAddr.String()] = pongContract
+
+	var (
+		sourcePortID       = wasmkeeper.PortIDForContract(pingContractAddr)
+		counterpartyPortID = wasmkeeper.PortIDForContract(pongContractAddr)
+	)
+	clientA, clientB, connA, connB := coordinator.SetupClientConnections(chainA, chainB, clientexported.Tendermint)
+	connA.NextChannelVersion = ping
```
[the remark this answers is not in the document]
alpe: Exactly. This is the version string that is used when the next channel is created from this connection

x/wasm/relay_pingpong_test.go
```diff
@@ -193,37 +199,81 @@ func (p *player) Execute(hash []byte, params wasmTypes.Env, data []byte, store p
...
+	return nil
+}
+
+func (p player) storeEndpoint(store prefix.Store, channel cosmwasmv2.IBCChannel) {
+	var counterparties []connectedChannelsModel
+	if b := store.Get(ibcEndpointsKey); b != nil {
+		require.NoError(p.t, json.Unmarshal(b, &counterparties))
+	}
+	counterparties = append(counterparties, connectedChannelsModel{Our: channel.Endpoint, Their: channel.CounterpartyEndpoint})
+	bz, err := json.Marshal(&counterparties)
```
ethanfrey, line 272: Yup, just like a wasm contract would

x/wasm/relay_pingpong_test.go
```diff
@@ -193,37 +199,81 @@ func (p *player) Execute(hash []byte, params wasmTypes.Env, data []byte, store p
...
 	if !ok {
 		return nil, 0, sdkerrors.Wrap(channeltypes.ErrChannelCapabilityNotFound, "module does not own channel capability")
 	}
+
 	service := NewHit(p.actor, start.Value)
 	p.t.Logf("[%s] starting game with: %d: %v\n", p.actor, start.Value, service)
 
 	var seq uint64 = 1
-	packet := channeltypes.NewPacket(service.GetBytes(), seq, start.Source.Port, start.Source.Channel, start.CounterParty.Port, start.CounterParty.Channel, doNotTimeout, 0)
+	packet := channeltypes.NewPacket(service.GetBytes(), seq, endpoints.Our.Port, endpoints.Our.Channel, endpoints.Their.Port, endpoints.Their.Channel, doNotTimeout, 0)
```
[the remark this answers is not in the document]
ethanfrey: Nice we could load them all and this `OnIBCChannelConnect` was useful.

x/wasm/relay_test.go
```diff
@@ -196,7 +196,7 @@ func (c *receiverContract) OnIBCPacketReceive(hash []byte, params cosmwasmv2.Env
 	ctx := c.chain.GetContext() // HACK: please note that this is not reverted after checkTX
 	err := c.chain.App.TransferKeeper.OnRecvPacket(ctx, ibcPacket, src)
 	if err != nil {
-		return nil, 0, sdkerrors.Wrap(err, "within our smart contract")
+		return nil, 0, sdkerrors.Wrap(err, "within Our smart contract")
```
ethanfrey, line 199: Maybe these should not be capitalized

ethanfrey (approved):
Looks good to merge except for some errand text renamings.

Two things needed in follow-up PRs:

* Run the ping pong until maxValue, so we get an failure and stop. This failure should be properly relayed and not return an error to the ibc handler (and rollback the tx).
* Try version negotiation (ADR25) and see if that works or not - if not, make an issue about it. This can be done separately in a minor PR so we can link the failing code without blocking on it being fixed.

Also submit the issue about ordered channels not working (maybe linking to another intentionally broken pr)

PR #260, opened 2020-08-25
Ping pong with error

### LINKED ISSUE #255: Complete IBC Mock testing

alpe:
Resolves #255 

Add tests for unhappy path

### DIFF
--- a/x/wasm/relay_pingpong_test.go
+++ b/x/wasm/relay_pingpong_test.go
@@ -88,7 +88,6 @@ func TestPinPong(t *testing.T) {
 
 		err = coordinator.RelayPacket(chainA, chainB, clientA, clientB, pkg, ack.GetBytes())
 		require.NoError(t, err)
-		//coordinator.CommitBlock(chainA, chainB)
 		err = coordinator.UpdateClient(chainA, chainB, clientA, clientexported.Tendermint)
 		require.NoError(t, err)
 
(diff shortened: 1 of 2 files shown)


x/wasm/relay_pingpong_test.go
```diff
@@ -121,6 +120,189 @@ func TestPinPong(t *testing.T) {
...
+		activePlayer  = ping
+		pingBallValue = startValue
+	)
+	for i := 1; i <= rounds-1; i++ { // play some rounds before reaching max value
+		t.Logf("++ round: %d\n", i)
+		ball := NewHit(activePlayer, pingBallValue)
+
+		seq := uint64(i)
+		pkg := channeltypes.NewPacket(ball.GetBytes(), seq, channelA.PortID, channelA.ID, channelB.PortID, channelB.ID, doNotTimeout, 0)
+		ack := ball.BuildAck()
```
ethanfrey, line 181: I just noticed this. The NewPacket creates and sends a packet, right, but it doesn't return the ack bytes?
You manually create the ack in the test code (not from the contract).

Is this due to unclear APIs? Is it not possible to read the packet/ack from the chain?

x/wasm/relay_pingpong_test.go
```diff
@@ -121,6 +120,189 @@ func TestPinPong(t *testing.T) {
...
+
+		err = coordinator.RelayPacket(chainA, chainB, clientA, clientB, pkg, ack.GetBytes())
```
ethanfrey, line 183: This is odd, you pass in the ack instead of it coming from the contract

ethanfrey (approved):
Good test cases, but some comments on where the packet/ack come from?

x/wasm/relay_pingpong_test.go
alpe: This is a "workaround" in ibc tests. The relayer should watch the chain instead and read the packet/ack information from the events exposed there. Unfortunately this is not implemented into the coordinator (yet) and also not a simple task that I could squeeze in easily.
This setup is also used in other ibc tests so we are in good company at least.

x/wasm/relay_pingpong_test.go
alpe: `channeltypes.NewPacket` does not sent a packet here. It only creates an object that is used below with the coordinator to be passed to the other chain (B). This object creation would not be necessary in a full relayer demo. See my other comment.

x/wasm/relay_pingpong_test.go
ethanfrey: Fair enough, limits of test environment.

PR #265, opened 2020-09-03
Fix CORS paramete name for docker sh

alpe:
The name has changed.


ethanfrey (approved):
Thanks. I just noticed this now to merge it.

PR #273, opened 2020-10-06
Wasm querier integration tests

### LINKED ISSUE #269: Add high-level integration tests for contract querier usage
We currently use the `reflect` contract to re-dispatch arbitary `CosmosMsg` and make sure those are handled well as callbacks.

When https://github.com/CosmWasm/cosmwasm/issues/526 is implemented, we can do the same with QueryRequest. It would be good to test the following:

* `WasmQuery::*`
* `StakingQuery::*`

These are tested little-to-no in the current tests

ethanfrey:
Closes #269 

- [x] Test `WasmQuery::Smart` and `WasmQuery::Raw` via 
- [x] Test basic staking queries with reflect contract

Not done:
-  Update docs explaining how to parse RawQuery (maybe update types in CosmWasm? maybe change from this `[]{Key, Value}` return value...) - this is moved to #275

### DIFF
--- a/x/wasm/internal/keeper/staking_test.go
+++ b/x/wasm/internal/keeper/staking_test.go
@@ -477,7 +477,33 @@ func TestQueryStakingInfo(t *testing.T) {
 	require.Contains(t, valInfo.MaxCommission, "0.200")
 	require.Contains(t, valInfo.MaxChangeRate, "0.010")
 
+	// test to get all my delegations
+	reflectAllDelegationsQuery := MaskQueryMsg{Chain: &ChainQuery{Request: &wasmTypes.QueryRequest{Staking: &wasmTypes.StakingQuery{
+		AllDelegations: &wasmTypes.AllDelegationsQuery{
+			Delegator: contractAddr.String(),
+		},
+	}}}}
+	reflectAllDelegationsBin := buildMaskQuery(t, &reflectAllDelegationsQuery)
+	res, err = keeper.QuerySmart(ctx, maskAddr, reflectAllDelegationsBin)
+	require.NoError(t, err)
+	// first we pull out the data from chain response, before parsing the original response
+	mustParse(t, res, &reflectRes)
+	var allDelegationsRes wasmTypes.AllDelegationsResponse
+	mustParse(t, reflectRes.Data, &allDelegationsRes)
+	require.Equal(t, 1, len(allDelegationsRes.Delegations))
+	delInfo := allDelegationsRes.Delegations[0]
+	fmt.Printf("%#v\n", delInfo)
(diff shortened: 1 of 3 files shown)


x/wasm/internal/keeper/staking_test.go
```diff
@@ -404,6 +406,106 @@ func TestReinvest(t *testing.T) {
...
+	initInfo := initializeStaking(t)
+	defer initInfo.cleanup()
+	ctx, valAddr, contractAddr := initInfo.ctx, initInfo.valAddr, initInfo.contractAddr
+	keeper, stakingKeeper, accKeeper := initInfo.wasmKeeper, initInfo.stakingKeeper, initInfo.accKeeper
+	distKeeper := initInfo.distKeeper
+
+	// initial checks of bonding state
+	val, found := stakingKeeper.GetValidator(ctx, valAddr)
+	require.True(t, found)
+	//initPower := val.GetDelegatorShares()
```
alpe, line 420: commented out code

x/wasm/internal/keeper/staking_test.go
```diff
@@ -404,6 +406,106 @@ func TestReinvest(t *testing.T) {
...
+	assert.Equal(t, val.Tokens, sdk.NewInt(1000000), "%s", val.Tokens)
```
alpe, line 421: nit: this and others: for better failure messages define expected value first:  `Equal(t, expectedVal, gotVal)

x/wasm/internal/keeper/staking_test.go
```diff
@@ -404,6 +406,106 @@ func TestReinvest(t *testing.T) {
...
+	require.Equal(t, contractAddr.String(), delInfo.Delegator)
+	// this is a different Coin type, with String not BigInt, compare field by field
+	require.Equal(t, funds[0].Denom, delInfo.Amount.Denom)
+	require.Equal(t, funds[0].Amount.String(), delInfo.Amount.Amount)
+
+	// TODO: more delegation queries
+	// BondedDenom -> Atom
+	// Delegation -> With detailed info
+}
+
```
alpe, line 508: pp: there was quite some effort to setup the test. I was wondering if a staking keeper mock would be an easier fit or what was your intention?

alpe (approved):
🌷 Very nice.

x/wasm/internal/keeper/staking_test.go
ethanfrey: We need to see how the query callbacks from the contracts really work.

I have a contract that will take the user input and use that to trigger a query via `Querier`. This allows us to see what a contract would see. And we need to see what would happen against a real staking keeper (as there are some bugs... we want to fix how those work, full stack for compatibility).

PR #274, opened 2020-10-06
Prevent integer overflow

alpe:
Backported from [Terra](https://github.com/terra-project/core/pull/402)

Background:
This was a finding of [Certik's](https://certik.io) code [audit](https://medium.com/certik/certik-audits-terras-new-cosmwasm-smart-contract-solution-67178e96188) on the CosmWasm code copied into the Terra core project.

### DIFF
--- a/x/wasm/internal/keeper/keeper.go
+++ b/x/wasm/internal/keeper/keeper.go
@@ -566,6 +566,10 @@ func (k Keeper) dispatchMessages(ctx sdk.Context, contractAddr sdk.AccAddress, m
 
 func gasForContract(ctx sdk.Context) uint64 {
 	meter := ctx.GasMeter()
+	// avoid integer overflow
+	if meter.IsOutOfGas() {
+		return 0
+	}
 	remaining := (meter.Limit() - meter.GasConsumed()) * GasMultiplier
 	if remaining > MaxGas {
 		return MaxGas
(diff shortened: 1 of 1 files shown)


x/wasm/internal/keeper/keeper.go
```diff
@@ -566,6 +566,10 @@ func (k Keeper) dispatchMessages(ctx sdk.Context, contractAddr sdk.AccAddress, m
 
 func gasForContract(ctx sdk.Context) uint64 {
 	meter := ctx.GasMeter()
+	// avoid integer overflow
```
ethanfrey, line 569: It avoids an underflow, when `meter.GasConsumed() > meter.Limit()`, comment should be corrected.

x/wasm/internal/keeper/keeper.go
```diff
@@ -566,6 +566,10 @@ func (k Keeper) dispatchMessages(ctx sdk.Context, contractAddr sdk.AccAddress, m
...
+	if meter.IsOutOfGas() {
```
ethanfrey, line 570: A better solution would be to use `meter.GasConsumedToLimit()` which is the proper API to avoid underflow: https://github.com/cosmos/cosmos-sdk/blob/v0.39.1/store/types/gas.go#L62-L67

However, looking at the code for the only two implementations of GasMeter, this will never fail. We cannot trust that I guess, but using `GasConsumedToLimit` uses an API that double-ensures we never return a value over `Limit`)

ethanfrey (commented):
Interesting point that can never actually happen.

If we do want a defensive measure here (fine), there is a cleaner way

alpe:
Good feedback. I dropped the comment as it did not made sense anymore and used `GasConsumedToLimit` ftw.

ethanfrey:
I was saying that you can remove the whole if block as you use the new "GasConsumedToLimit". But fine. We have tripple defense against overflow here (the implementation never allow it, and we check here twice)

PR #279, opened 2020-10-07
Improve query testing

webmaster128:
Test query tesults based on `[]byte`, which is the original response that can be `nil`.

Tests nil keys and adapt docs.


ethanfrey (approved):
Good test improvement

PR #281, opened 2020-10-07
Test WasmQuery::Raw with missing data

ethanfrey:
- [x] Add tests 
- [x] Fix (Go-)CosmWasm (https://github.com/CosmWasm/go-cosmwasm/pull/143 and https://github.com/CosmWasm/cosmwasm/pull/567)
- [x] Update wasmd to work

Error message:

```
Error Trace:    reflect_test.go:421
                Error:          Received unexpected error:
                                
                                github.com/CosmWasm/wasmd/x/wasm/internal/keeper.Keeper.QuerySmart
                                        /home/ethan/go/src/github.com/cosmwasm/wasmd/x/wasm/internal/keeper/keeper.go:419
                                github.com/CosmWasm/wasmd/x/wasm/internal/keeper.TestWasmRawQueryWithNil
                                        /home/ethan/go/src/github.com/cosmwasm/wasmd/x/wasm/internal/keeper/reflect_test.go:420
                                query wasm contract failed: Generic error: Querier system error: Cannot parse response: Parsing Go response: expected value at line 1 column 8 in: {"ok":{}}
                Test:           TestWasmRawQueryWithNil
```

This is what we got from the reflect contract: `Generic error: Querier system error: Cannot parse response: Parsing Go response: expected value at line 1 column 8 in: {"ok":{}}`, which seems to be a bit Go's "omitempty"

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/CosmWasm/wasmd
 go 1.14
 
 require (
-	github.com/CosmWasm/go-cosmwasm v0.11.0-alpha2
+	github.com/CosmWasm/go-cosmwasm v0.11.0-rc
 	github.com/cosmos/cosmos-sdk v0.39.1-0.20200727135228-9d00f712e334
 	github.com/golang/mock v1.4.3 // indirect
 	github.com/google/gofuzz v1.0.0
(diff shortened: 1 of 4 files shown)


ethanfrey:
Tracing this down, it happens in go-cosmwasm: 

```go
type QueryResponse struct {
	Ok  []byte `json:"ok,omitempty"`
	Err string `json:"error,omitempty"`
}
```

You can see how this is encoded: https://github.com/CosmWasm/go-cosmwasm/blob/c11572e00c50bbed6e75a3ea50bca09f5773e56c/types/queries.go#L21-L64

Basically, I creates:

```go
QuerierResult{
  Ok: &QueryResponse{
    Ok: nil
  }
}
```

Which then neither encodes "ok" or "err" variants

PR #293, opened 2020-10-27
Upgrade to Stargate

alpe:
Upgrade to cosmos-sdk `6e569e125571`

Replaces https://github.com/CosmWasm/wasmd/pull/252

PR #294, opened 2020-10-28
Update README

conr2d:
Update the command in README example for running with docker


alpe:
Thanks for the PR! The script name has changed, indeed. 💐 

There is a big chunk of work in https://github.com/CosmWasm/wasmd/pull/293 and I want to avoid another rebase due to updates in master. Please bare with us for some more days before we can merge this.

conr2d:
@alpe Sure, thank you. :)

ethanfrey (approved):
Good catch. Thank you.

PR #299, opened 2020-11-03
Upgrade to Stargate RC2

alpe:
Stargate RC2

Left to do:
- [x] manual smoke tests


ethanfrey (approved):
Looked at the diff and seems good.

PR #300, opened 2020-11-05
Support self calling contract on instantiation

### LINKED ISSUE #292: Instantiate contract process ordering
## Problem State
As you can see in https://github.com/terra-project/core/issues/399

If you store the instantiated contract after dispatching all messages, it will block the messages to use instantiated contract.

https://github.com/CosmWasm/wasmd/blob/ff37dc68f2a4ea86abf7b7f124c166c793d63e67/x/wasm/internal/keeper/keeper.go#L242-L250

## Proposed solution
So recommend to store the instantiated contract data and dispatch response messages later.

alpe:
Resolves #292 

While fixing the issue was trivial testing it was not with the current setup. This patch also introduces a new interface `WasmerEngine` that is implemented by `cosmwasm.Wasmer`.  This enables us to bypass the wasmer rust engine in tests for more flexibility and a faster feedback cycle.


x/wasm/internal/keeper/test_common.go
```diff
@@ -115,10 +117,12 @@ var TestingStakeParams = stakingtypes.Params{
 type TestKeepers struct {
 	AccountKeeper authkeeper.AccountKeeper
 	StakingKeeper stakingkeeper.Keeper
+	// deprecated use WasmKeeperRef instead
```
ethanfrey, line 120: Why keep both of these? This was my confusion with the setup above.

If needed for some commits to get to a stable state, makes sense. But let's remove it and replace all calls with `WasmKeeperRef` before merging.

x/wasm/internal/keeper/keeper_test.go
```diff
@@ -431,6 +433,43 @@ func TestInstantiateWithNonExistingCodeID(t *testing.T) {
...
+	keepers.WasmKeeper.wasmer = wasmerMock
+	example := StoreHackatomExampleContract(t, ctx, keepers)
+	_, err = keepers.WasmKeeper.Instantiate(ctx, example.CodeID, example.CreatorAddr, nil, nil, "test", nil)
+	require.NoError(t, err)
+	assert.True(t, wasmerMock.excuteCalled)
+}
+
+// mock to call itself on instantiation
+type selfCallingInstMockWasmer struct {
+	AlwaysFailMockWasmer
```
ethanfrey, line 452: That always failed when you override Instantiate and Execute is a bit confusing. I expect this to be an error, not actually do something. Mayb just provide a Query implementation (or extend NoopMockWasmer with a positive query operation).

x/wasm/internal/keeper/test_common.go
```diff
@@ -412,3 +418,37 @@ func keyPubAddr() (crypto.PrivKey, crypto.PubKey, sdk.AccAddress) {
...
+func (a AlwaysFailMockWasmer) Migrate(code cosmwasm.CodeID, env wasmTypes.Env, info wasmTypes.MessageInfo, migrateMsg []byte, store cosmwasm.KVStore, goapi cosmwasm.GoAPI, querier cosmwasm.Querier, gasMeter cosmwasm.GasMeter, gasLimit uint64) (*wasmTypes.MigrateResponse, uint64, error) {
+	return nil, 0, MockDefaultError
+}
+
+func (a AlwaysFailMockWasmer) GetCode(code cosmwasm.CodeID) (cosmwasm.WasmCode, error) {
+	return nil, MockDefaultError
+}
+
+func (a AlwaysFailMockWasmer) Cleanup() {
+	panic("implement me")
```
ethanfrey, line 453: Huh? That seems dangerous. Where is Cleanup called, just in test code?

ethanfrey: I would make this a no-op, I don't think we ever try to handle panics there.

x/wasm/internal/types/wasmer_engine.go
```diff
@@ -0,0 +1,102 @@
+package types
+
+import (
+	"github.com/CosmWasm/go-cosmwasm"
+	"github.com/CosmWasm/go-cosmwasm/types"
+)
+
+// WasmerEngine defines the WASM contract runtime engine.
+type WasmerEngine interface {
```
ethanfrey, line 9: Nice idea to abstract it, kind of like you did in the ibc spike.

ethanfrey (commented):
Nice abstraction and way to quickly iterate on tests of the go logic.

Added some comments. I think a bit more polish on it and it will be ready to merge.

x/wasm/internal/keeper/test_common.go
alpe: 👍  I tried to not touch too many things here but better go the full mile.

x/wasm/internal/keeper/keeper_test.go
alpe: The `AlwaysFailMockWasmer` implements all the methods to satisfy the WasmEngine interface. I overwrite only 3 methods here for the test so that I can create the proper data. A call to any other method should fail the test.
An always panic wasmer engine implementation would be more accurate though....
I will refactor this.

alpe:
Thanks for the feedback. I have revisited the test support types to apply the feedback. 
I would be happy to make another round if needed to get this right.

x/wasm/internal/keeper/test_common.go
alpe: This would have been called from tests only

x/wasm/internal/keeper/keeper_test.go
```diff
@@ -436,38 +436,33 @@ func TestInstantiateWithNonExistingCodeID(t *testing.T) {
 func TestInstantiateWithCallbackToContract(t *testing.T) {
 	ctx, keepers := CreateTestInput(t, false, SupportedFeatures, nil, nil)
 	var (
-		wasmerMock = &selfCallingInstMockWasmer{}
-		err        error
+		excuteCalled bool
+		err          error
 	)
-	keepers.WasmKeeperRef.wasmer = wasmerMock
+	wasmerMock := &MockWasmer{
```
ethanfrey, line 442: I like this constructor over extending `AlwaysFailMockWasmer`. But it does seem a bit odd to inline this.. Fine as now, as it is only used once, but maybe we can make it reusable for other tests. My rough thought:

```go
func selfCallingInstMockWasmer(executeCalled *bool) *MockWasmer {
    return &MockWasmer{

}	CreateFn: func(code cosmwasm.WasmCode) (cosmwasm.CodeID, error) {
			anyCodeID := bytes.Repeat([]byte{0x1}, 32)
			return anyCodeID, nil
		},
		InstantiateFn: func(code cosmwasm.CodeID, env wasmTypes.Env, info wasmTypes.MessageInfo, initMsg []byte, store cosmwasm.KVStore, goapi cosmwasm.GoAPI, querier cosmwasm.Querier, gasMeter cosmwasm.GasMeter, gasLimit uint64) (*wasmTypes.InitResponse, uint64, error) {
			return &wasmTypes.InitResponse{
				Messages: []wasmTypes.CosmosMsg{
					{Wasm: &wasmTypes.WasmMsg{Execute: &wasmTypes.ExecuteMsg{ContractAddr: env.Contract.Address, Msg: []byte(`{}`)}}},
				},
			}, 1, nil
		},
		ExecuteFn: func(code cosmwasm.CodeID, env wasmTypes.Env, info wasmTypes.MessageInfo, executeMsg []byte, store cosmwasm.KVStore, goapi cosmwasm.GoAPI, querier cosmwasm.Querier, gasMeter cosmwasm.GasMeter, gasLimit uint64) (*wasmTypes.HandleResponse, uint64, error) {
			excuteCalled = true
			return &wasmTypes.HandleResponse{}, 1, nil
		},
	}
}
```

And in this test we could do:

```go
executedCalled := false
wasmerMock = selfCallingInstMockWasmer(&executeCalled)
// ...
_, err := keepers.WasmKeeper.Instantiate(ctx, example.CodeID, example.CreatorAddr, nil, nil, "test", nil)
assert.True(t, excuteCalled)
```

Not a blocker and this can always be refactored later. I like the use of MockWasmer.

x/wasm/internal/keeper/test_common.go
```diff
@@ -117,12 +116,10 @@ var TestingStakeParams = stakingtypes.Params{
 type TestKeepers struct {
 	AccountKeeper authkeeper.AccountKeeper
 	StakingKeeper stakingkeeper.Keeper
-	// deprecated use WasmKeeperRef instead
-	WasmKeeper    Keeper
```
ethanfrey, line 118: :+1:

x/wasm/internal/keeper/test_common.go
```diff
@@ -419,36 +415,98 @@ func keyPubAddr() (crypto.PrivKey, crypto.PubKey, sdk.AccAddress) {
 	return key, pub, addr
 }
 
-var _ types.WasmerEngine = &AlwaysFailMockWasmer{}
+var _ types.WasmerEngine = &alwaysPanicMockWasmer{}
```
ethanfrey, line 418: Very clear name

ethanfrey: can't you also define

`var alwaysPanicMockWasmer = &MockWasmer{}`

And then just refer to that where needed?

x/wasm/internal/keeper/test_common.go
```diff
@@ -419,36 +415,98 @@ func keyPubAddr() (crypto.PrivKey, crypto.PubKey, sdk.AccAddress) {
...
-	return nil, 0, MockDefaultError
+func (a alwaysPanicMockWasmer) Cleanup() {
+	panic("not supposed to be called!")
 }
 
-func (a AlwaysFailMockWasmer) Query(code cosmwasm.CodeID, env wasmTypes.Env, queryMsg []byte, store cosmwasm.KVStore, goapi cosmwasm.GoAPI, querier cosmwasm.Querier, gasMeter cosmwasm.GasMeter, gasLimit uint64) ([]byte, uint64, error) {
-	return nil, 0, MockDefaultError
+var _ types.WasmerEngine = &MockWasmer{}
+
+// MockWasmer implements types.WasmerEngine for testing purpose. One or multiple messages can be stubbed.
```
ethanfrey, line 452: I like this for composition. Very nice!

ethanfrey (approved):
Thank you for addressing the feedback.
You can make another round if you find it useful and merge it when you are satisfied. They are just stylistic comments now.

PR #301, opened 2020-11-05
Reject wasm zips exceeding limit

### LINKED ISSUE #287: Return error if wasm to big
Currently it cuts of the wasm at 400kb, but doesn't return an error, leading to odd error messages.

We can raise the size limit a bit, maybe 600kb. But mainly, we need to return an error on the limit. Not sure how to do it with this interface.

https://github.com/CosmWasm/cosmwasm/issues/576#issuecomment-706766984

alpe:
Resolves #287


alpe:
Included in #302

PR #302, opened 2020-11-06
Reject wasm code exceeding limit 

### LINKED ISSUE #287: Return error if wasm to big

### LINKED ISSUE #289: Make Wasm maxSize a configuration option
Right now we have a constant Wasm size limit at https://github.com/CosmWasm/wasmd/blob/v0.11.1/x/wasm/internal/keeper/ioutil.go#L15-L16, and there is the desire to change its value.

I think this value should be configurable when using x/wasm and not be fixed.

Increasing the value to 600 KB in wasmd would be nice as well since the random beacon contract is about 500 KB big.

alpe:
* Resolves #287
* Resolves #289

Summary of changes:
* Returns ErrLimit when wasm code exceeds max size
* New default wasm code max size: 600 * 1024 bytes
* Max size configurable via parameters
* Limit also applied to uncompressed wasm code


x/wasm/internal/keeper/ioutil.go
```diff
@@ -28,6 +30,24 @@ func uncompress(src []byte) ([]byte, error) {
...
+type LimitedReader struct {
+	r *io.LimitedReader
+}
 
-	return ioutil.ReadAll(io.LimitReader(zr, maxSize))
+func (l *LimitedReader) Read(p []byte) (n int, err error) {
+	if l.r.N <= 0 {
+		return 0, types.ErrLimit
+	}
+	return l.r.Read(p)
```
ethanfrey, line 52: ReadAll keeps reading until one read returns `(0, nil)` or any return `(x, err)` correct?

Update: checked [ioutil](https://golang.org/pkg/io/ioutil/#ReadAll) and [bytes.Buffer](https://golang.org/pkg/bytes/#Buffer.ReadFrom). They only stop when it hits an error. If the error is EOF, then it stops and return data with no error, otherwise, it returns the error.

Seems fine, just note this returns an error if the input is the exact size of n.

x/wasm/internal/keeper/ioutil.go
```diff
@@ -28,6 +30,24 @@ func uncompress(src []byte) ([]byte, error) {
 		return nil, err
 	}
 	zr.Multistream(false)
+	defer zr.Close()
+	return ioutil.ReadAll(LimitReader(zr, int64(limit)))
```
ethanfrey, line 34: This means if the total size >= limit we will get the types.ErrLimit error. However, we want total_size > limit.
I would use `int64(limit + 1)` here, so if there is one byte more than the requested limit, it returns an error

x/wasm/internal/keeper/ioutil_test.go
```diff
@@ -57,31 +60,38 @@ func TestUncompress(t *testing.T) {
 			src:      wasmGzipped[:len(wasmGzipped)-5],
 			expError: io.ErrUnexpectedEOF,
 		},
+		"handle limit gzip output": {
```
ethanfrey, line 63: Huh, this passes? That means it can return `(n, EOF)` where `n > 0`?

The test cases (both limit and limit+1 gzips) pass. So I can approve it. Just seems there is some possible way something can slip through.

x/wasm/internal/keeper/ioutil.go
ethanfrey: Or maybe, so the read, then check if it overflowed? Or maybe I don't fully understand how LimitedReader works.

x/wasm/internal/keeper/keeper.go
```diff
@@ -110,6 +110,12 @@ func (k Keeper) getInstantiateAccessConfig(ctx sdk.Context) types.AccessType {
 	return a
 }
 
+func (k Keeper) GetMaxWasmCodeSize(ctx sdk.Context) uint64 {
```
ethanfrey, line 113: Okay, nice to add the parameter here - this is set in genesis, right?

x/wasm/internal/types/types.proto
```diff
@@ -32,6 +32,7 @@ message Params {
     option (gogoproto.goproto_stringer) = false;
     AccessConfig code_upload_access = 1 [(gogoproto.nullable) = false, (gogoproto.moretags) = "yaml:\"code_upload_access\""];
     AccessType instantiate_default_permission = 2 [(gogoproto.moretags) = "yaml:\"instantiate_default_permission\""];
+    uint64 max_wasm_code_size = 3 [(gogoproto.moretags) = "yaml:\"max_wasm_code_size\""];
```
ethanfrey, line 35: Yup. Nice addition here.

Reading the config option (1k gas) on upload (> 1 million gas) is virtually 0 overhead. Happy for a param here.

ethanfrey (approved):
Looks good. I am scratching my head how it is sure that limit passes and limit+1 fails. But tests pass for those cases and your understanding of the go stdlib is better than mine. 

You can merge as is or attempt to add some more defensive code there if you think my worry is justified.

x/wasm/internal/types/params.go
```diff
@@ -15,10 +15,13 @@ import (
 const (
 	// DefaultParamspace for params keeper
 	DefaultParamspace = ModuleName
+	// DefaultMaxWasmCodeSize limit max bytes read to prevent gzip bombs
```
ethanfrey, line 18: Very nice updates here.
You have thought more angles than I did.

:100:

alpe:
The underlying (gzip) reader returns `n, io.EOF` for the last read with a `size <=limit`. `n, EOF` are handled in `ioutil.ReadAll`
The limit reader comes into play only for `read > limit` where the last call before returned `n, nil`.
I have added a missing test case for uncompressed bytes.

x/wasm/internal/keeper/ioutil.go
```diff
@@ -5,20 +5,22 @@ import (
...
-
 // uncompress returns gzip uncompressed content or given src when not gzip.
-func uncompress(src []byte) ([]byte, error) {
-	if len(src) < 3 {
+func uncompress(src []byte, limit uint64) ([]byte, error) {
+	switch n := uint64(len(src)); {
+	case n < 3:
 		return src, nil
+	case n > limit:
+		return nil, types.ErrLimit
```
webmaster128, line 23: This check also rejects compressed data > limit, which is smaller than limit when uncompressed. Is this intended?

alpe: good edge case!
This can only happen when gzip does not compress the data but adds it header on top. I guess that is very unlikely to happen and the workaround would be to send uncompressed data.

webmaster128: Right. If found the code a bit confusing to review, because it is not very clear which parts deals with compressed and which with uncompressed data. Some checks operate on both.

PR #307, opened 2020-11-11
Use correct directories for commands

alpe:
`wasmd add-genesis-account` command was accessing the wrong folder for the key

Also added some tests from cosmos-sdk simapp


alpe:
Closing this in favour of https://github.com/CosmWasm/wasmd/issues/308

PR #311, opened 2020-11-12
Upgrade wasmvm aka go-cosmwasm

### LINKED ISSUE #309: Upgrade go-cowasmwasm to wasmvm 0.12
We now have a https://github.com/CosmWasm/wasmvm tagged with v0.12.0-alpha1.

Important changes since  go-cowasmwasm 0.11:
- Renaming of the repo and the dynamic libraries
- Wasmer was renamed to VM and is created using NewVM
- NewVM new takes two additional arguments documented here: https://github.com/CosmWasm/wasmvm/blob/v0.12.0-alpha1/lib.go#L41-L42

alpe:
Resolves #309
This got a bit bigger than expected:

* Rename go-cosmwasm to wasmvm
* Upgrade wasmvm to `v0.12.0-alpha1
* Upgrade Dockerfile for new `go-ext-builder`
* Fix `WasmConfig` parsing
* New ` --wasm.lru_cache_size` and `--wasm.query_gas_limit` wasmd start parameter
* Default "lru cache size" to 100 MiB
* Upgrade CircleCI build container to Go 1.15
* Stop tests on race conditions via `GORACE=halt_on_error=1`


x/wasm/internal/keeper/keeper.go
```diff
@@ -73,7 +73,9 @@ func NewKeeper(
 	customEncoders *MessageEncoders,
 	customPlugins *QueryPlugins,
 ) Keeper {
-	wasmer, err := wasm.NewWasmer(filepath.Join(homeDir, "wasm"), supportedFeatures)
+	debug := false
+	const cacheSize = 0 // 100 * 1 << 20 // 100 MiB
```
webmaster128, line 77: This value is in MiB, not bytes. So the `<< 20` seems wrong here

webmaster128 (commented):
Diff lgtm. Just one thing.

alpe:
Pushed a new version

.circleci/config.yml
```diff
@@ -71,6 +71,7 @@ jobs:
       - run:
           name: Run tests with coverage
           command: |
+            GORACE=halt_on_error=1
```
webmaster128, line 74: Is there no `export ` needed here? I don't really understand the difference to be honest.

x/wasm/README.md
```diff
@@ -10,9 +10,9 @@ You can add the following section to `config/app.toml`. Below is shown with defa
 [wasm]
 # This is the maximum sdk gas (wasm and storage) that we allow for any x/wasm "smart" queries
 query_gas_limit = 300000
-# This is the number of wasm vm instances we keep cached in memory for speed-up
-# Warning: this is currently unstable and may lead to crashes, best to keep for 0 unless testing locally
-lru_size = 0
+# This defines the memory size for wasm contract instances that we can keep cached for speed-up
```
webmaster128, line 13:
```suggestion
# This defines the memory size for Wasm modules that we can keep cached to speed-up instantiation
```

x/wasm/README.md
```diff
@@ -10,9 +10,9 @@ You can add the following section to `config/app.toml`. Below is shown with defa
...
+# The value is in MiB not bytes
+lru_cache_size = 0
```
webmaster128, line 15:
```suggestion
memory_cache_size = 300
```

The type of cache (lru) is an implementation detail that can change any time. Also we can now encourage using the cache by setting a few hundred MiB for it.

x/wasm/internal/types/types.go
```diff
@@ -1,13 +1,17 @@
...
-	wasmTypes "github.com/CosmWasm/go-cosmwasm/types"
+	wasmvmtypes "github.com/CosmWasm/wasmvm/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 )
 
-const defaultLRUCacheSize = uint64(0)
-const defaultQueryGasLimit = uint64(3000000)
+const (
+	DisabledLRUCacheSize     uint32 = 0
```
webmaster128, line 10: Is this constant needed? I don't think we need a special case for disabling the cache. Just set to 0 if you don't want to spend momre than 0 for the cache.

x/wasm/internal/types/types.go
```diff
@@ -1,13 +1,17 @@
...
+	defaultLRUCacheSize      uint32 = 100 // in MiB
```
webmaster128, line 11:
```suggestion
	defaultMemoryCacheSize      uint32 = 100 // in MiB
```

x/wasm/internal/types/types.go
```diff
@@ -203,14 +207,18 @@ func ParseEvents(logs []wasmTypes.EventAttribute, contractAddr sdk.AccAddress) s
 
 // WasmConfig is the extra config required for wasm
 type WasmConfig struct {
-	SmartQueryGasLimit uint64 `mapstructure:"query_gas_limit"`
-	CacheSize          uint64 `mapstructure:"lru_size"`
+	SmartQueryGasLimit uint64
+	// LRUCacheSize in MiB not bytes
+	LRUCacheSize uint32
```
webmaster128, line 212:
```suggestion
	MemoryCacheSize uint32
```

x/wasm/internal/types/types.go
```diff
@@ -203,14 +207,18 @@ func ParseEvents(logs []wasmTypes.EventAttribute, contractAddr sdk.AccAddress) s
...
+	// ContractDebugMode log what contract print
+	ContractDebugMode bool
 }
 
 // DefaultWasmConfig returns the default settings for WasmConfig
 func DefaultWasmConfig() WasmConfig {
 	return WasmConfig{
 		SmartQueryGasLimit: defaultQueryGasLimit,
-		CacheSize:          defaultLRUCacheSize,
+		LRUCacheSize:       defaultLRUCacheSize,
```
webmaster128, line 221:
```suggestion
		MemoryCacheSize:       defaultMemoryCacheSize,
```

x/wasm/module.go
```diff
@@ -26,6 +29,12 @@ var (
 	_ module.AppModuleBasic = AppModuleBasic{}
 )
 
+// Module init related flags
+const (
+	flagWasmLRUSize       = "wasm.lru_cache_size"
```
webmaster128, line 34:
```suggestion
	flagWasmMemoryCacheSize       = "wasm.memory_cache_size"
```

x/wasm/module.go
```diff
@@ -170,3 +179,35 @@ func (am AppModule) RegisterStoreDecoder(sdr sdk.StoreDecoderRegistry) {
 func (am AppModule) WeightedOperations(simState module.SimulationState) []simtypes.WeightedOperation {
 	return nil
 }
+
+//____________________________________________________________________________
+
+// AddModuleInitFlags implements servertypes.ModuleInitFlags interface.
+func AddModuleInitFlags(startCmd *cobra.Command) {
+	defaults := DefaultWasmConfig()
+	startCmd.Flags().Uint32(flagWasmLRUSize, defaults.LRUCacheSize, "Sets the size in MiB (NOT bytes) of an in-memory cache for wasm contract instances. Set to 0 to disable.")
```
webmaster128, line 188:
```suggestion
	startCmd.Flags().Uint32(flagWasmLRUSize, defaults.LRUCacheSize, "Sets the size in MiB (NOT bytes) of an in-memory cache for wasm modules. Set to 0 to disable.")
```

webmaster128 (commented):
A bit more 💅 but looks nice overall

x/wasm/internal/keeper/keeper.go
```diff
@@ -73,7 +73,7 @@ func NewKeeper(
 	customEncoders *MessageEncoders,
 	customPlugins *QueryPlugins,
 ) Keeper {
-	wasmer, err := wasm.NewWasmer(filepath.Join(homeDir, "wasm"), supportedFeatures)
+	wasmer, err := wasmvm.NewVM(filepath.Join(homeDir, "wasm"), supportedFeatures, wasmConfig.ContractDebugMode, wasmConfig.LRUCacheSize)
```
ethanfrey, line 76: Ah, cool, this allows us to debug contracts in a running chain with the proper app config. Cool stuff.

x/wasm/internal/types/types.go
```diff
@@ -203,14 +207,18 @@ func ParseEvents(logs []wasmTypes.EventAttribute, contractAddr sdk.AccAddress) s
 
 // WasmConfig is the extra config required for wasm
 type WasmConfig struct {
-	SmartQueryGasLimit uint64 `mapstructure:"query_gas_limit"`
-	CacheSize          uint64 `mapstructure:"lru_size"`
+	SmartQueryGasLimit uint64
```
ethanfrey, line 210: Looking to see where this is read/parsed, but surprised there are no explicit field naming here.

x/wasm/module.go
```diff
@@ -170,3 +179,35 @@ func (am AppModule) RegisterStoreDecoder(sdr sdk.StoreDecoderRegistry) {
...
+		if cfg.LRUCacheSize, err = cast.ToUint32E(v); err != nil {
+			return cfg, err
+		}
+	}
+	if v := opts.Get(flagWasmQueryGasLimit); v != nil {
+		if cfg.SmartQueryGasLimit, err = cast.ToUint64E(v); err != nil {
+			return cfg, err
+		}
+	}
+	// attach contract debugging to global "trace" flag
```
ethanfrey, line 206: Nice one - turn on all debuging metrics with one switch, not a complex series of hidden ones.

x/wasm/module.go
```diff
@@ -170,3 +179,35 @@ func (am AppModule) RegisterStoreDecoder(sdr sdk.StoreDecoderRegistry) {
 func (am AppModule) WeightedOperations(simState module.SimulationState) []simtypes.WeightedOperation {
 	return nil
 }
+
+//____________________________________________________________________________
+
+// AddModuleInitFlags implements servertypes.ModuleInitFlags interface.
+func AddModuleInitFlags(startCmd *cobra.Command) {
```
ethanfrey, line 186: Okay, so we no longer have `app.toml`, but rather pass in these as cli flags always. (Which explains why there is no codec name mapping in the struct). I guess using such a config file is no longer supported/recommended?

x/wasm/module_test.go
```diff
@@ -369,6 +370,59 @@ func TestHandleExecuteEscrow(t *testing.T) {
 	assert.Equal(t, sdk.Coins(nil), data.bankKeeper.GetAllBalances(data.ctx, contractAcct.GetAddress()))
 }
 
+func TestReadWasmConfig(t *testing.T) {
```
ethanfrey, line 373: :+1:

ethanfrey (approved):
Can you add these new command line flags to the README?

Otherwise, looks top.

x/wasm/module.go
alpe: The configuration in `app.toml` works as before. We don't need a struct and `mapstructure` tag as before. It is enough to register a flag with the `wasm.` prefix. So both ways do work.

cmd/wasmd/root.go
```diff
@@ -115,6 +116,7 @@ func initRootCmd(rootCmd *cobra.Command, encodingConfig app.EncodingConfig) {
 
 func addModuleInitFlags(startCmd *cobra.Command) {
 	crisis.AddModuleInitFlags(startCmd)
+	wasm.AddModuleInitFlags(startCmd)
```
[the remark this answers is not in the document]
alpe: ja, they started to push flags down to the modules.

.circleci/config.yml
alpe: > halt_on_error (default 0): Controls whether the program exits after reporting first data race.

https://golang.org/doc/articles/race_detector.html#Options

webmaster128: I was referring to the (missing?) `export` of `GORACE`, not the content of it.
See https://stackoverflow.com/a/1158231/2013738

x/wasm/internal/types/types.go
alpe: The .toml file is parsed with a viper instance that had the wasm flags registered. We are not parsing the file in our app package anymore

PR #312, opened 2020-11-13
Integrate wasmcli into wasmd

### LINKED ISSUE #308: Merge wasmd and wasmcli into a single binary
Like [gaiad](https://github.com/cosmos/gaia/tree/gaiav3.0/cmd) and [simd](https://github.com/cosmos/cosmos-sdk/tree/master/simapp/simd).

alpe:
Resolves #308 

Good bye to wasmcli, coral, gaiaflex clients


ethanfrey:
I had one thought on this last night.

`wasmd` requires the `lib_wasmvm.so` runtime. That is only built for Linux and OSX. Meaning, we cannot compile `wasmd` for Windows. If we completely remove `wasmcli`, we will not have a windows based cli now.

I see two options here:

* Present cosmjs as the only client interface for Windows
* Maintain wasmcli for windows, while `wasmd` also works as an all-in-one toolkit in platforms that support it.

I lean toward the second. If so, we should test this (once) to make sure it works (Abel uses Windows), and support it only for Windows. Then again, maybe we just ask people to install WSL (Windows Subsystem for Linux), and the CLI tools would just work. (Does anyone running vanilla Windows without WSL actually use CLI tools??)

Makefile
```diff
@@ -54,7 +54,6 @@ build_tags_comma_sep := $(subst $(empty),$(comma),$(build_tags))
 
 ldflags = -X github.com/cosmos/cosmos-sdk/version.Name=wasm \
 		  -X github.com/cosmos/cosmos-sdk/version.AppName=wasmd \
-		  -X github.com/CosmWasm/wasmd/cmd/wasmcli/version.ClientName=wasmcli \
```
ethanfrey, line 57: As mentioned in the comment, maybe we don't remove the CLI stuff, just keep it for Windows?

cmd/wasmd/root.go
```diff
@@ -119,6 +124,73 @@ func addModuleInitFlags(startCmd *cobra.Command) {
 	wasm.AddModuleInitFlags(startCmd)
 }
 
+func queryCommand() *cobra.Command {
```
ethanfrey, line 127: This all came from wasmcli, right?

contrib/local/02-contracts.sh
```diff
@@ -5,80 +5,80 @@ DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
...
 echo "## List code"
-wasmcli query wasm list-code --node=http://localhost:26657 --chain-id=testing | jq
+wasmd query wasm list-code --node=http://localhost:26657 --chain-id=testing | jq
 
 echo "-----------------------"
 echo "## Create new contract instance"
-INIT="{\"verifier\":\"$(wasmcli keys show validator -a)\", \"beneficiary\":\"$(wasmcli keys show fred -a)\"}"
-wasmcli tx wasm instantiate "$CODE_ID" "$INIT" --admin=$(wasmcli keys show validator -a) \
+INIT="{\"verifier\":\"$(wasmd keys show validator -a)\", \"beneficiary\":\"$(wasmd keys show fred -a)\"}"
+wasmd tx wasm instantiate "$CODE_ID" "$INIT" --admin=$(wasmd keys show validator -a) \
```
ethanfrey, line 24: Yes, `wasmd` here everywhere, even if we keep the cli for windows

ethanfrey (approved):
The merge looks good.

I wonder again about providing some WIndows CLI support. But unless anyone actually uses this without WSL, maybe we don't need to care.

@webmaster128 do you know if wasmer-reborn will compile on windows? (Which would mean we can eventually add full windows support for `wasmd`)

cmd/wasmd/root.go
alpe: ja, they were in the cli only

webmaster128:
> @webmaster128 do you know if wasmer-reborn will compile on windows? (Which would mean we can eventually add full windows support for `wasmd`)

I will try to find out. Let's track this issue in https://github.com/CosmWasm/wasmvm/issues/28. For the near future, we don't have Windows support.

PR #314, opened 2020-11-17
Replace sdk.AccAddress with bech32 string

### LINKED ISSUE #306: Use string type for AccAddresses in proto
Context: https://github.com/cosmos/cosmos-sdk/issues/7195

alpe:
Resolves #306


x/wasm/internal/keeper/test_common.go
```diff
@@ -319,9 +335,9 @@ func handleExecute(ctx sdk.Context, k *Keeper, msg *types.MsgExecuteContract) (*
 	return res, nil
 }
 
-func AnyAccAddress(_ *testing.T) sdk.AccAddress {
+func AnyAccAddress(_ *testing.T) string {
```
webmaster128, line 338: If `AccAddress` is a binary address, isn't the funtion name misleading now? What about `func RandomAddress`?

webmaster128 (commented):
Beautiful

x/wasm/internal/keeper/test_common.go
alpe: Ja, it is a random account address in bech32 format. So `RandomBech32AccountAddress`

webmaster128 (approved):

PR #316, opened 2020-11-17
Set bech32 prefix

### LINKED ISSUE #313: Change bech32 prefix for wasmd
Too many software in this ecosystem blindly inherits the Cosmos Hub SLIP44 coin index and the `cosmos` address prefix. To ensure tooling is compatible with different prefixes, let us use a default that is not `cosmos`.

We want to use the wasmd binary for a bunch of upcoming testsnets and avoid the overhead of creating a new binary every time. We can use the same address prefix for all of them. I propose `wasm`.

alpe:
Resolves #313 

* Default "wasm" in app 
* Set prefix in build (not really needed with app default but consistent with building other artifacts)

No covered are cosmos prefixes in unit tests


webmaster128 (approved):
thanks

PR #317, opened 2020-11-18
Add pagination to grpc queries

### LINKED ISSUE #242: Stargate: Add pagination support for queries
At the moment we return all elements.

With stargate we can use the `"cosmos/query/pagination.proto"`in the rpc endpoint.

This includes upgrading the cli to use the grpc query server

alpe:
Resolves #242 

New:
* New second index for contracts by code id and creation marker
* Contract code history elements stored into separate prefix


x/wasm/internal/types/query.proto
```diff
@@ -53,52 +53,70 @@ message QueryContractInfoRequest {
...
+
     // address is the address of the contract
     string address = 1;
     ContractInfo contract_info = 2 [(gogoproto.embed) = true, (gogoproto.jsontag) = ""];
 }
 
 // QueryContractHistoryRequest is the request type for the Query/ContractHistory RPC method
 message QueryContractHistoryRequest {
-    // address is the address of the contract to query
+    // address is the address of the contract to query´
```
webmaster128, line 65:
```suggestion
    // address is the address of the contract to query
```

x/wasm/internal/keeper/test_common.go
```diff
@@ -340,28 +340,36 @@ func RandomBech32AccountAddress(_ *testing.T) string {
...
+	wasmCode, err := ioutil.ReadFile(wasmFile)
 	require.NoError(t, err)
 
 	codeID, err := keepers.WasmKeeper.Create(ctx, creatorAddr, wasmCode, "", "", nil)
 	require.NoError(t, err)
-	return HackatomExampleContract{anyAmount, creator, creatorAddr, codeID}
+	return ExampleContract{anyAmount, creator, creatorAddr, codeID}
 }
 
 type HackatomExampleInstance struct {
```
webmaster128, line 371:
```suggestion
type ExampleContractInstance struct {
```

x/wasm/internal/types/keys.go
```diff
@@ -60,3 +55,44 @@ func GetContractAddressKey(addr sdk.AccAddress) []byte {
 func GetContractStorePrefixKey(addr sdk.AccAddress) []byte {
 	return append(ContractStorePrefix, addr...)
 }
+
+// GetContractByCreatedSecondaryIndexKey returns the key for the secondary index:
+// `<prefix><codeID><created><contractAddr>`
+func GetContractByCreatedSecondaryIndexKey(contractAddr sdk.AccAddress, c *ContractInfo) []byte {
+	prefix := GetContractByCodeIDSecondaryIndexPrefix(c.CodeID)
```
webmaster128, line 62: This is a bit tricky to follow because the term "prefix" does not match the one in the comment above. Seems correct but any naming improvement would be appreciated.

x/wasm/internal/types/query.proto
```diff
@@ -53,52 +53,70 @@ message QueryContractInfoRequest {
...
+    // pagination defines the pagination in the response.
+    cosmos.base.query.v1beta1.PageResponse pagination = 2;
 }
 
 // QueryRawContractStateRequest is the request type for the Query/RawContractState RPC method
 message QueryRawContractStateRequest {
     // address is the address of the contract
     string address = 1;
     bytes query_data = 2;
+    ;
```
webmaster128, line 119: This looks stray

x/wasm/internal/types/types.go
```diff
@@ -130,6 +128,24 @@ func (c *ContractInfo) AdminAddr() sdk.AccAddress {
 	return admin
 }
 
+// NewAbsoluteTxPosition gets a block position from the context
+func NewAbsoluteTxPosition(ctx sdk.Context) *AbsoluteTxPosition {
+	// we muxst safely handle nil gas meters
```
webmaster128, line 133:
```suggestion
	// we must safely handle nil gas meters
```

webmaster128: Why do we need to do this? This leads to multiple transactions at `TxIndex` of `0` which destroys al guarantees of unique positioning

x/wasm/internal/types/types.go
```diff
@@ -141,18 +157,18 @@ func (a *AbsoluteTxPosition) LessThan(b *AbsoluteTxPosition) bool {
...
-	if meter != nil {
-		index = meter.GasConsumed()
-	}
-	return &AbsoluteTxPosition{
-		BlockHeight: ctx.BlockHeight(),
-		TxIndex:     index,
+// AbsoluteTxPositionLen number of elements in byte representation
+const AbsoluteTxPositionLen = 16
+
+// Bytes encodes the object into a 16 byte representation with big endian block height adn tx index.
```
webmaster128, line 163:
```suggestion
// Bytes encodes the object into a 16 byte representation with big endian block height and tx index.
```

x/wasm/internal/types/types.proto
```diff
@@ -100,10 +97,10 @@ message ContractCodeHistoryEntry {
 
 }
 
-// AbsoluteTxPosition can be used to sort contracts
+// AbsoluteTxPosition is a block position that can be used to sort contracts
 message AbsoluteTxPosition {
     // BlockHeight is the block the contract was created at
-    int64 block_height = 1;
+    uint64 block_height = 1;
```
webmaster128, line 103: 🤟

webmaster128 (commented):
Nice stuff; some thoughs

x/wasm/internal/keeper/test_common.go
alpe: This is not "any" contract instance but the Hackatom example. Some tests rely on this specific setup so that I do not want to mix this.

webmaster128: I see, right. I did not read the code well enough.

x/wasm/internal/types/keys.go
alpe: This is a bit tricky indeed. This one generates a full key while the one above was the prefix used with the store.  To be more correct, I will rename the `GetContractStorePrefixKey` method to `GetContractStorePrefix` so that it is matching the other method signatures. Good 👁️

webmaster128: Could we make the name in `<prefix><codeID><created><contractAddr>` more specific? Something like `<bucket_prefix>` or however you call those now? Then we free "prefix" as a term for any kind of prefix that changes depending on the context.

x/wasm/internal/types/types.go
alpe: An empty block gas meter can only exist in tests afaik. I am not sure about the genesis block though.
But the sorting in the new GRPC endpoint take also the contract address into account. It uses the new secondary index `<prefix><codeID><created><contractAddr>` so that we have an order.

The legacy rest endpoint was not updated though to use the second index. The old code only looks at `Created` for sorting. I will compare for address when equal `created`.

PR #318, opened 2020-11-19
Fix chain id setup in contrib/local/setup_wasmd.sh

orkunkl:
The script did not use `CHAIN_ID` variable.


alpe:
Good 👁️ . 
The `$CHAIN_ID` is not used in the other scripts in that folder but hard coded to `testing`. The only requirement so far was that it has to work on my box :-) 
Do you have a use case where you want this different than `testing` ? I would suggest that this PR either removes the variable or also touches the other scripts. WDTY?

orkunkl:
I was using to it set up a local chain named `musselnet`. not using docker. I found this bug and thought would be good to fix this.

PR #320, opened 2020-11-20
Exclude generate proto code files in coverage

### LINKED ISSUE #298: Check codecov report
It shows only `17%` covered which is too low.
https://codecov.io/gh/cosmwasm/wasmd

alpe:
Resolves #298 

With this patch we just ignore the Go files generated by protobuf and some test files.

Most of this is copied from https://github.com/cosmos/cosmos-sdk/blob/master/.codecov.yml . Thanks 💐


orkunkl (approved):
LGTM

PR #332, opened 2020-12-04
Replace outdated changelog

### LINKED ISSUE #331: Restart changelog
The `changelog.md` was not maintained well and contains misleading version numbers. We will reset it and start with `v0.12.0` as the baseline.

alpe:
Resolves #331 

Generated with
https://github.com/github-changelog-generator/github-changelog-generator

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,162 +1,87 @@
-<!--
-Guiding Principles:
-
-Changelogs are for humans, not machines.
-There should be an entry for every single version.
-The same types of changes should be grouped.
-Versions and sections should be linkable.
-The latest version comes first.
-The release date of each version is displayed.
-Mention whether you follow Semantic Versioning.
-
-Usage:
-
-Change log entries are to be added to the Unreleased section under the
-appropriate stanza (see below). Each entry should ideally include a tag and
-the Github issue reference in the following format:
-
-* (<tag>) \#<issue-number> message
-
-The issue numbers will later be link-ified during the release process so you do
-not have to worry about including a link manually, but you can if you wish.
-
-Types of changes (Stanzas):
-
-"Features" for new features.
-"Improvements" for changes in existing functionality.
-"Deprecated" for soon-to-be removed features.
-"Bug Fixes" for any bug fixes.
-"Client Breaking" for breaking CLI commands and REST routes.
-"State Machine Breaking" for breaking the AppState
-
-Ref: https://keepachangelog.com/en/1.0.0/
--->
-
 # Changelog
 
-## [Unreleased]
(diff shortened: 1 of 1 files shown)


ethanfrey (approved):
Lgtm for auto gen.

Two comments:
1. We need to be extra careful with the names of issues or prs, so the title makes sense without more context.
2. Some items (eg closed issues as "will not do") should not appear here. Maybe you can generate the log with the tool, and then do a bit of hand sanitizing, to remove noise.

alpe:
Agree that we should be more sensible with titles. Hand sanitizing would not work as it always generates the full changelog and not only the diff to the last commit. But with a better use of labels we can have this covered as well:

> Automatically exclude specific issues that are irrelevant to your changelog (by default, any issue labeled question, duplicate, invalid, or wontfix) ✂️

https://github.com/github-changelog-generator/github-changelog-generator#features-and-advantages-of-this-project

There is one behaviour that we need to consider in the workflow: the v0.13.0 section can only be generated when the tag was set. In order to have a proper changelog,  we need to generate it at the latest before we tag, then set the tag and generate again after the tag. The PR to close an issue will not contain the changelog for it.... 🤔 
Nevertheless with a generator we will not miss an issue anymore and also we can be more calm on merge conflicts. Any mistake would be fixed with the next run....

alpe:
I have gone through the closed issues without a PR and assigned exclusion labels before regenerating.

ethanfrey:
Thank you for the explanation and it sounds good.

> There is one behaviour that we need to consider in the workflow: the v0.13.0 section can only be generated when the tag was set. In order to have a proper changelog, we need to generate it at the latest before we tag, then set the tag and generate again after the tag. The PR to close an issue will not contain the changelog for it.... 🤔

Looks like we need some time traveling circular dependencies here. 👍 

What about.
* prepare 0.13 pr
* locally tag v0.13.0 (do not push tag)
* generate changelog, commit and push
* merge pr
* pull latest master, `git tag -f v0.13.0` and push that tag

alpe:
🤔  the generated doc seems not to be deterministic. I got a different file for every run now on the same codebase.
I will have another look for a generator.

alpe:
It worked better with an docker image: ` docker run -it --rm -v "$(pwd)":/usr/local/src/wasmd ferrarimarco/github-changelog-generator --USER=CosmWasm --Project=wasmd --TOKEN=`

alpe:
Cropped everything before v0.12.0

I think some best practice to use labels for better semantic makes sense. But as discussed on internal discord, I would go with a manual changelog in the future and use the generator before a release to collect the data and double check.

PR #338, opened 2020-12-11
Fix param usage issue in genesis import

### LINKED ISSUE #335: Errors in `InitGenesis`
There exist errors in `InitGenesis` of the latest code.

1. Using params before they are initialized.
   - https://github.com/CosmWasm/wasmd/blob/master/x/wasm/internal/keeper/genesis.go#L17
      - https://github.com/CosmWasm/wasmd/blob/master/x/wasm/internal/keeper/keeper.go#L166
         - https://github.com/CosmWasm/wasmd/blob/master/x/wasm/internal/keeper/keeper.go#L117
   - This will cause a nil error.

2. The `LastCodeID` is always zero after importing codes
   - In `importCode`, It inserts codes but does not increase lastCodeId
   - So It always steps into the error case from the below code.
      - https://github.com/CosmWasm/wasmd/blob/master/x/wasm/internal/keeper/genesis.go#L47-L49

alpe:
Resolves #335  (first point only)

The genesis test setup was modified in `setupKeeper ` to not have params set there. The reported issue was confirmed by the tests then.

### DIFF
--- a/x/wasm/internal/keeper/genesis.go
+++ b/x/wasm/internal/keeper/genesis.go
@@ -12,6 +12,8 @@ import (
 //
 // CONTRACT: all types of accounts must have been already initialized/created
 func InitGenesis(ctx sdk.Context, keeper *Keeper, data types.GenesisState) error {
+	keeper.setParams(ctx, data.Params)
+
 	var maxCodeID uint64
 	for i, code := range data.Codes {
 		err := keeper.importCode(ctx, code.CodeID, code.CodeInfo, code.CodeBytes)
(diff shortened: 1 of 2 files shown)


x/wasm/internal/keeper/genesis_test.go
```diff
@@ -482,12 +482,11 @@ func TestImportContractWithCodeHistoryReset(t *testing.T) {
...
 }
 
-func setupKeeper(t *testing.T) (*Keeper, sdk.Context, []sdk.StoreKey, func()) {
+func setupKeeper(t *testing.T) (*Keeper, sdk.Context, []sdk.StoreKey) {
 	t.Helper()
 	tempDir, err := ioutil.TempDir("", "wasm")
 	require.NoError(t, err)
-	cleanup := func() { os.RemoveAll(tempDir) }
-	//t.Cleanup(cleanup) todo: add with Go 1.14
+	t.Cleanup(func() { os.RemoveAll(tempDir) })
```
ethanfrey, line 489: Such a nice trick, cleaning up a lot

x/wasm/internal/keeper/genesis.go
```diff
@@ -12,6 +12,8 @@ import (
 //
 // CONTRACT: all types of accounts must have been already initialized/created
 func InitGenesis(ctx sdk.Context, keeper *Keeper, data types.GenesisState) error {
+	keeper.setParams(ctx, data.Params)
```
ethanfrey, line 15: Much better

ethanfrey (approved):
ack

PR #346, opened 2020-12-16
Improve Contract initialization in genesis and tooling

### LINKED ISSUE #326: Make it easy to initialize contracts in genesis file
Use Case:

I want to launch a chain and create a number of multisig accounts in genesis that control a number of genesis tokens. This is doable for the cosmos sdk fixed multisig, but I would rather use a cosmwasm contract. This should be easy to do, and easy to inspect when reviewing the genesis.

Current State:

We can currently create initial `x/wasm` state by "dump state and restart" style. However, the format is the raw data and not very human readable. 

Proposal:

A much more friendly version would be to allow a number of `x/wasm` messages to be included in the genesis. We don't need these to be signed by 3rd parties (like gentx), but rather just allow setting a "sender" in the messages themselves. At minimumum, we would want to support `MsgStoreCode`, `MsgInstantiateContract` and `MsgExecuteContract`. This would let us flexibly set the code, contracts, and then make any adjustments (like setting some initial allowances on `cw1-subkeys`).

To make this usable, we also need a utility to calculate the contract address a priori. This can be calculated deterministically by the code id and the contract sequence. I propose adding a simple cli command to `x/wasm` and `wasmd` that take a code id and contract sequence (you can determine this via genesis), and return what the contract address would be. This address then can be used when setting bank balances, or referred to by other contracts.

Alternatives:

Only add support for `MsgInstantiateContract` and `MsgExecuteContract` and use the existing format to initialize the code blobs.

alpe:
Resolves #326 

- Extend wasm genesis with `gen_mgs` field for messages
- handle messages in genesis init process

```
> wasmd add-wasm-genesis-message -h

genesis_contract_msg_326 f014acb
Wasm genesis subcommands

Usage:
  wasmd add-wasm-genesis-message [flags]
  wasmd add-wasm-genesis-message [command]

Available Commands:
  execute              Execute a command on a wasm contract
  instantiate-contract Instantiate a wasm contract
  list-codes           Lists all codes from genesis contract dump and queued messages
  list-contracts       Lists all contracts from genesis contract dump and queued messages
  store                Upload a wasm binary

Flags:
  -h, --help   help for add-wasm-genesis-message

Global Flags:
      --home string         directory for config and data (default "/Users/alex/.wasmd")
      --log_format string   The logging format (json|plain) (default "plain")
      --log_level string    The logging level (trace|debug|info|warn|error|fatal|panic) (default "info")
      --trace               print out full stack trace on errors

```


app/app.go
```diff
@@ -409,7 +409,7 @@ func NewWasmApp(logger log.Logger, db dbm.DB, traceStore io.Writer, loadLatest b
 		distr.NewAppModule(appCodec, app.distrKeeper, app.accountKeeper, app.bankKeeper, app.stakingKeeper),
 		staking.NewAppModule(appCodec, app.stakingKeeper, app.accountKeeper, app.bankKeeper),
 		upgrade.NewAppModule(app.upgradeKeeper),
-		wasm.NewAppModule(&app.wasmKeeper),
+		wasm.NewAppModule(&app.wasmKeeper, app.stakingKeeper),
```
ethanfrey, line 412: Oh nice, so we can even call staking from init contracts. And eg. set up staking derivatives in Genesis

doc/proto.md
```diff
@@ -116,6 +117,25 @@ GenesisState - genesis state of x/wasm
...
+
+
+
+
+
+
+<a name="cosmwasm.wasm.v1beta1.GenesisState.GenMsgs"></a>
+
+### GenesisState.GenMsgs
+GenMsgs define the messages that can be executed during genesis phase.
```
ethanfrey, line 130: A mention they are executed in order, and how to reference contract addresses? Or I guess that goes in `x/wasm`

x/wasm/internal/keeper/genesis.go
```diff
@@ -29,30 +35,56 @@ func InitGenesis(ctx sdk.Context, keeper *Keeper, data types.GenesisState) error
...
 	}
 	if keeper.peekAutoIncrementID(ctx, types.KeyLastInstanceID) <= uint64(maxContractID) {
-		return sdkerrors.Wrapf(types.ErrInvalid, "seq %s must be greater %d ", string(types.KeyLastInstanceID), maxContractID)
+		return nil, sdkerrors.Wrapf(types.ErrInvalid, "seq %s must be greater %d ", string(types.KeyLastInstanceID), maxContractID)
+	}
+
+	if len(data.GenMsgs) == 0 {
+		return nil, nil
+	}
+	for _, genTx := range data.GenMsgs {
```
ethanfrey, line 65: Nice and simple.

x/wasm/internal/keeper/genesis_test.go
```diff
@@ -482,6 +483,42 @@ func TestImportContractWithCodeHistoryReset(t *testing.T) {
 	assert.Equal(t, expHistory, keeper.GetContractHistory(ctx, contractAddr))
 }
 
+func TestImportWithGenMsg(t *testing.T) {
```
ethanfrey, line 486: We'll definitely need a test of setting up staking (validator set changes) in the init messages.

As well as executing a newly created contract (and giving it a bank balance).

I'm sure you have these on your TODO, just wanted to mention them

x/wasm/internal/keeper/test_common.go
```diff
@@ -282,21 +282,37 @@ func CreateTestInput(t *testing.T, isCheckTx bool, supportedFeatures string, enc
 func TestHandler(k *Keeper) sdk.Handler {
 	return func(ctx sdk.Context, msg sdk.Msg) (*sdk.Result, error) {
 		ctx = ctx.WithEventManager(sdk.NewEventManager())
-
 		switch msg := msg.(type) {
+		case *types.MsgStoreCode:
```
ethanfrey, line 286: nice

x/wasm/internal/types/genesis.proto
```diff
@@ -12,6 +13,18 @@ message GenesisState {
     repeated Code codes = 2 [(gogoproto.nullable) = false, (gogoproto.jsontag) = "codes,omitempty"];
     repeated Contract contracts = 3 [(gogoproto.nullable) = false, (gogoproto.jsontag) = "contracts,omitempty"];
     repeated Sequence sequences = 4 [(gogoproto.nullable) = false, (gogoproto.jsontag) = "sequences,omitempty"];
+    repeated GenMsgs gen_msgs = 5 [(gogoproto.nullable) = false, (gogoproto.jsontag) = "gen_msgs,omitempty"];
+
+    // GenMsgs define the messages that can be executed during genesis phase.
+    // The intention is to have more human readable data that is auditable.
+    message GenMsgs {
```
ethanfrey, line 20: nice

ethanfrey (commented):
Nice start. I like the approach and implementation so far

doc/proto.md
alpe: I have added `list-contracts` and `list-code` commands that also display the queued message data. So it should be easy to find the contract address for any instantiation.

ethanfrey: That is very cool. So I write an init one, and then I can list them to see what the address was and use that result for future tx (or bank accounts).

Nice tooling

x/wasm/internal/keeper/genesis_test.go
alpe: My test strategy here is unit based. Which means the new functionality is based upon existing components and tests so that I need to cover the changes only. 
This means concrete for the `InitGenesis` method that the `msgHandler` and `stakingKeeper` are called with the right params as well as return values handled as expected. I can verify this with a mock more efficient.
What's not covered by tests is the wiring in the app package. We don't have automated tests for this, yet. I run `contrib/local` scripts on my box to verify 

To setup staking via genesis message or to execute a newly created contract is using existing logic in the keeper. Do you think we need an additional integration test for this? What makes the genesis phase different to a normal TX execution?

ethanfrey: I see all the detail logic is unit tested. And it is super hard to write automatic go tests for this level of interaction.

Just saying you ran one test locally full stack is good.

x/wasm/client/cli/gov_tx.go
```diff
@@ -23,27 +24,44 @@ func ProposalStoreCodeCmd() *cobra.Command {
 				return err
 			}
 
-			src, err := parseStoreCodeArgs(args, clientCtx)
+			src, err := parseStoreCodeArgs(args[0], clientCtx.FromAddress, cmd.Flags())
```
ethanfrey, line 27: This change is due to the genesis changes?
Or a more general update/refactor to the new stargate proposals?

Looks fine, just trying to understand why this is here

x/wasm/client/cli/tx.go
```diff
@@ -76,8 +76,8 @@ func StoreCodeCmd() *cobra.Command {
 	return cmd
 }
 
-func parseStoreCodeArgs(args []string, cliCtx client.Context) (types.MsgStoreCode, error) {
-	wasm, err := ioutil.ReadFile(args[0])
+func parseStoreCodeArgs(file string, sender sdk.AccAddress, flags *flag.FlagSet) (types.MsgStoreCode, error) {
```
ethanfrey, line 79: Cleaner than passing in full context

x/wasm/client/cli/tx.go
```diff
@@ -94,23 +94,48 @@ func parseStoreCodeArgs(args []string, cliCtx client.Context) (types.MsgStoreCod
 	}
 
 	var perm *types.AccessConfig
-	if onlyAddrStr := viper.GetString(flagInstantiateByAddress); onlyAddrStr != "" {
+	onlyAddrStr, err := flags.GetString(flagInstantiateByAddress)
```
ethanfrey, line 97: This seems much more complex than using viper.

I guess it differentiates between not set (error) and set to empty sting as two different cases we handled the same before? Or is there any reason `flag.GetString()` would return an error besides being unset, which we actually want to allow, right? Huh

x/wasm/internal/keeper/genesis.go
```diff
@@ -4,21 +4,25 @@ import (
...
+// ValidatorSetSource is a subset of the staking keeper
+type ValidatorSetSource interface {
+	ApplyAndReturnValidatorSetUpdates(sdk.Context) (updates []abci.ValidatorUpdate, err error)
+}
+
 // InitGenesis sets supply information for genesis.
 //
 // CONTRACT: all types of accounts must have been already initialized/created
-func InitGenesis(ctx sdk.Context, keeper *Keeper, data types.GenesisState) error {
+func InitGenesis(ctx sdk.Context, keeper *Keeper, data types.GenesisState, stakingKeeper ValidatorSetSource, msgHandler sdk.Handler) ([]abci.ValidatorUpdate, error) {
```
ethanfrey, line 18: Nice abstraction for the staking keeper, showing exactly why is it used :+1:

x/wasm/internal/keeper/genesis_test.go
```diff
@@ -355,19 +360,51 @@ func TestFailFastImport(t *testing.T) {
 				Params: types.DefaultParams(),
 			},
 		},
+		"validator set update called for any genesis messages": {
```
ethanfrey, line 363: Nice test with mocks

x/wasm/module.go
```diff
@@ -131,10 +133,11 @@ func (AppModule) QuerierRoute() string {
 func (am AppModule) InitGenesis(ctx sdk.Context, cdc codec.JSONMarshaler, data json.RawMessage) []abci.ValidatorUpdate {
 	var genesisState GenesisState
 	cdc.MustUnmarshalJSON(data, &genesisState)
-	if err := InitGenesis(ctx, am.keeper, genesisState); err != nil {
+	validators, err := InitGenesis(ctx, am.keeper, genesisState, am.validatorSetSource, am.Route().Handler())
+	if err != nil {
 		panic(err)
 	}
-	return []abci.ValidatorUpdate{}
+	return validators
```
ethanfrey, line 140: Cool, I didn't realize we could return this - thus surprised by the return value. We were always just returning nil.

Nice that you use this :+1:

ethanfrey (approved):
Very nice. I think this is ready to merge. I was just a bit confused by some of the changes to cli/proposal parsing, but the genesis stuff looks solid.

One comment, which could be added later, is to add a file (REAMDE or other) that shows how to add a few items to the genesis file via cli and how to get the contract addresses that would be created. I guess `contrib/local/00-genesis.sh` is a good reference, but a bit of test in the README (linking to that example) would make onboarding easier

x/wasm/client/cli/tx.go
alpe: Viper can not be used as before. The previous call was using a static instance but this has changed in the sdk. There is a viper instance behind the scenes but `flags` is also populated from the `app.toml` file with defaults.
`viper.GetString` simply ignored the error.

x/wasm/client/cli/gov_tx.go
alpe: Ja, I am also using this parseStoreCodeArgs for the genesis message command

alpe:
After #358

alpe:
Rebased

x/wasm/client/cli/genesis_msg.go
```diff
@@ -293,6 +284,52 @@ func getAllCodes(state *types.GenesisState) ([]CodeMeta, error) {
...
+					Label:   msg.Label,
+				},
+			})
+			seq++
+		}
+	}
+	return all
+}
+
+func hasAccountBalance(cmd *cobra.Command, a map[string]json.RawMessage, sender sdk.AccAddress, coins sdk.Coins) (bool, error) {
```
ethanfrey, line 319: variable `a` is a bit confusing.

But I would note this check is only required if `! coins.IsEmpty()`. I guess it could make sense that funded only genesis accounts send messages, but if we just want to prevent obvious errors, we can short circuit this is we are not sending any balance, so we let accounts with no balances create contracts as long as they don't interact with funds.

ethanfrey (approved):
Looks good.

I will merge, but please address my comment in a follow-up PR (or let me know it is better how it is). I can now at least do last integration tests

PR #351, opened 2020-12-30
Register upgrade plan for musselnet-v2

orkunkl:
This change is required to succesfully executing a cosmovisor assisted upgrade. Reference: https://godoc.org/github.com/cosmos/cosmos-sdk/x/upgrade#hdr-Performing_Upgrades

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -307,7 +307,7 @@ func NewWasmApp(logger log.Logger, db dbm.DB, traceStore io.Writer, loadLatest b
 		app.getSubspace(crisistypes.ModuleName), invCheckPeriod, app.bankKeeper, authtypes.FeeCollectorName,
 	)
 	app.upgradeKeeper = upgradekeeper.NewKeeper(skipUpgradeHeights, keys[upgradetypes.StoreKey], appCodec, homePath)
-
+	app.registerUpgradePlans()
 	// register the staking hooks
 	// NOTE: stakingKeeper above is passed by reference, so that it will contain these hooks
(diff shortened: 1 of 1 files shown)


alpe:
@orkunkl I have modified the target branch as discussed. Please update the "upgrade name". Otherwise looks good 👍

alpe (approved):

app/app.go
```diff
@@ -645,3 +645,9 @@ func initParamsKeeper(appCodec codec.BinaryMarshaler, legacyAmino *codec.LegacyA
 
 	return paramsKeeper
 }
+
+func (app *WasmApp) registerUpgradePlans() {
+	app.upgradeKeeper.SetUpgradeHandler("v0.13.1", func(ctx sdk.Context, plan upgradetypes.Plan) {
```
ethanfrey, line 650: Looks good. I would just name this "musselnet-v2" or something like that.

ethanfrey (approved):

ethanfrey:
I heard some upgrade issues with -rc6 (maybe earlier). It would be good to do some local testing.

Once #354 is merged, please rebase on/merge from that one. And it would be good to locally test a (one node) upgrade from the current musselnet version (0.12.0?) to this commit.

I would be happy to report any issues to the cosmos sdk team. Or report that this worked without problems as a positive test.

ethanfrey:
We will not perform this.
I will close the PR, but the branch is a good example of an upgrade script. Let's not delete the branch

PR #354, opened 2021-01-05
Upgrade to cosmos-sdk v0.40.0-rc6

alpe:
Upgrade to [`cosmos-sdk v0.40.0-rc6`](https://github.com/cosmos/cosmos-sdk/releases/tag/v0.40.0-rc6)

There is a known 🐛 [#8251](https://github.com/cosmos/cosmos-sdk/issues/8251) with this SDK version in the gentx CLI command. The workaround is documented in the issue as well:
```sh
# pass the staked amount as argument AND as parameter
wasmd gentx validator "250000000$STAKE" --chain-id="$CHAIN_ID" --amount="250000000$STAKE"
```


contrib/local/setup_wasmd.sh
```diff
@@ -22,5 +22,5 @@ for addr in "$@"; do
   wasmd add-genesis-account "$addr" "1000000000$STAKE,1000000000$FEE"
 done
 # submit a genesis validator tx
-(echo "$PASSWORD"; echo "$PASSWORD"; echo "$PASSWORD") | wasmd gentx validator --chain-id="$CHAIN_ID" --amount "250000000$STAKE"
+(echo "$PASSWORD"; echo "$PASSWORD"; echo "$PASSWORD") | wasmd gentx validator "250000000$STAKE" --chain-id="$CHAIN_ID"
```
ethanfrey, line 25: Oh, this is important part for the cosmjs tests.
I don't know how best to flag it, so @willclarktech @webmaster128 Note this when we test cosmjs against 0.14.0 wasmd

go.mod
```diff
@@ -4,8 +4,8 @@ go 1.15
 
 require (
 	github.com/CosmWasm/wasmvm v0.12.0
-	github.com/cosmos/cosmos-sdk v0.40.0-rc5
-	github.com/cosmos/iavl v0.15.0-rc5
+	github.com/cosmos/cosmos-sdk v0.40.0-rc6
+	github.com/cosmos/iavl v0.15.3
```
ethanfrey, line 8: Ohh... one less rc dependency :smile: 

Faster than dependabot :wink:

x/wasm/internal/keeper/staking_test.go
```diff
@@ -616,7 +616,7 @@ func addValidator(t *testing.T, ctx sdk.Context, stakingKeeper stakingkeeper.Kee
 	pubKey := privKey.PubKey()
 	addr := sdk.ValAddress(pubKey.Address())
 
-	pkAny, err := codectypes.PackAny(pubKey)
+	pkAny, err := codectypes.NewAnyWithValue(pubKey)
```
ethanfrey, line 619: This is the only needed code change. Cool.

ethanfrey (approved):
Good stuff.
Not sure how to document the client changes (diff staking args).

Maybe a little changelog entry?

contrib/local/setup_wasmd.sh
webmaster128: Does this change anything for the caller of this script? We should get the change when bundled as part of the docker container.

alpe: yes, this is part of the docker script as well.
Please note: due to the 🐛  in the sdk the workaround is to pass the staked tokens twice. As argument and as flag.

alpe:
I had added https://github.com/CosmWasm/docs/issues/128 to update the docs

PR #358, opened 2021-01-07
Upgrade WasmVM to v0.13.0

alpe:
Upgrade WasmVM to [v0.13.0](https://github.com/CosmWasm/wasmvm/releases/tag/v0.13.0)

### DIFF
--- a/x/wasm/internal/keeper/keeper.go
+++ b/x/wasm/internal/keeper/keeper.go
@@ -42,6 +42,10 @@ const InstanceCost uint64 = 40_000
 // CompileCost is how much SDK gas we charge *per byte* for compiling WASM code.
 const CompileCost uint64 = 2
 
+// contractMemoryLimit is the memory limit of each contract execution (in MiB)
+// constant value so all nodes run with the same limit.
+const contractMemoryLimit = 32
+
 // Keeper will have a reference to Wasmer with it's own data directory.
 type Keeper struct {
 	storeKey      sdk.StoreKey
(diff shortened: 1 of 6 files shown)


x/wasm/internal/keeper/keeper.go
```diff
@@ -42,6 +42,10 @@ const InstanceCost uint64 = 40_000
 // CompileCost is how much SDK gas we charge *per byte* for compiling WASM code.
 const CompileCost uint64 = 2
 
+// contractMemoryLimit is the memory limit of each contract execution (in MiB)
+// constant value so all nodes run with the same limit.
+const contractMemoryLimit = 32
```
ethanfrey, line 47: Let's make this a param in a later PR.
If not too much work, we can add that as a param (with default 32 MB) in a separate PR, but include that in the 0.14 release.

I am happy to merge this first and test it manually before that addition. (It can come later as well)

ethanfrey (approved):
Looks good, as soon as broken CI test fixed.
One idea for a follow up

x/wasm/internal/keeper/keeper.go
ethanfrey: After reflection, that params issue is part of #202 
Let's just do that with everything else

webmaster128 (commented):
This PR updates the Go code, but not `libwasmvm_muslc.a`. `Dockerfile` needs to be updated as well.

ethanfrey:
Ah, I will check the Dockerfile. Thanks for the review

ethanfrey:
Interesting the .a file doubled in size from 0.12 to 0.13 from ~30MB to ~60MB

Not a major issue, but interesting if you know why

```
$ ls -l
total 99228
-rw-rw-r-- 1 ethan ethan 38236768 ene  7 13:40 012.a
-rw-rw-r-- 1 ethan ethan 63363772 ene  7 13:38 013.a
```

webmaster128:
> Interesting the .a file doubled in size from 0.12 to 0.13 from ~30MB to ~60MB

One part of it is probably that we have the (unused) cranelift compiler in our dependencies now. This is a consequence of how Wasmer packages and features are organized now.

ethanfrey:
Ah, that makes sense.

Final docker image grows from 224MB to 237MB, so no real issue there. Just an oddity I wanted to flag.
Including cranelift makes sense.

PR #363, opened 2021-01-07
Document upgrade path

ethanfrey:
Added some detailed docs (and step-by-step guide) for using `x/upgrade` to upgrade the chain.

This failed for musselnet due to state changed (in block sigs I think) between sdk 0.40.0-rc3 and 0.40.0-rc6. Looks like we will need a dump state and restart for this. Or just make a new testnet. 

Check out the new [UPGRADING.md](https://github.com/CosmWasm/wasmd/blob/document-upgrade/UPGRADING.md) or skip directly to the [panic on starting the new version](https://github.com/CosmWasm/wasmd/blob/document-upgrade/UPGRADING.md#swap-out-binaries)

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,17 +119,12 @@ docker volume rm -f wasmd_data
 docker run --rm -it \
     -e PASSWORD=xxxxxxxxx \
     --mount type=volume,source=wasmd_data,target=/root \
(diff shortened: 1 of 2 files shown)


alpe:
Looks good! Untested ACK

alpe (approved):

PR #366, opened 2021-01-08
Refactor to GRPC message server

### LINKED ISSUE #328: Register GRPC message service
We support the GRPC query handler but there is also a message handler implementation in other modules like:

```
// RegisterServices registers module services.
func (am AppModule) RegisterServices(cfg module.Configurator) {
	types.RegisterMsgServer(cfg.MsgServer(), keeper.NewMsgServerImpl(am.keeper))
	querier := keeper.Querier{Keeper: am.keeper}
	types.RegisterQueryServer(cfg.QueryServer(), querier)
}

```

alpe:
Resolves #328

This PR introduces support for the sdk grpc server to handle messages.
It introduces a breaking change to the response data that is returned. Instead of our custom data it is marshalled protobuf now. See `tx.proto` for service and type definitions

Left to do:
- [ ] Tests for keeper/message server
- [ ] Refactor handler tests to use a MsgServer mock


x/wasm/genesis_test.go
```diff
@@ -30,7 +30,10 @@ func TestInitGenesis(t *testing.T) {
 
 	res, err := h(data.ctx, &msg)
 	require.NoError(t, err)
-	require.Equal(t, res.Data, []byte("1"))
+
+	var pStoreResp MsgStoreCodeResponse
```
ethanfrey, line 34: If we do similar tests a lot (eg. in handler tests), it may be useful to have helper methods for them

x/wasm/handler.go
```diff
@@ -2,215 +2,70 @@ package wasm
...
-func handleInstantiate(ctx sdk.Context, k *Keeper, msg *MsgInstantiateContract) (*sdk.Result, error) {
-	senderAddr, err := sdk.AccAddressFromBech32(msg.Sender)
-	if err != nil {
-		return nil, sdkerrors.Wrap(err, "sender")
-	}
-	var adminAddr sdk.AccAddress
-	if msg.Admin != "" {
-		if adminAddr, err = sdk.AccAddressFromBech32(msg.Admin); err != nil {
-			return nil, sdkerrors.Wrap(err, "admin")
+func hasWasmModuleAttribute(attrs []abci.EventAttribute) bool {
```
ethanfrey, line 63: Why do we need this? I am a bit unsure and need to investigate the event/attribute system a bit more.

There were multiple sub-messages merging into each other before causing confusion. I will try some tests in a branch

x/wasm/internal/keeper/msg_server.go
```diff
@@ -0,0 +1,185 @@
+package keeper
```
ethanfrey, line 1: Huh, isn't this supposed to be auto-gen code?

x/wasm/internal/types/tx.proto
```diff
@@ -0,0 +1,126 @@
...
+  uint64 code_id = 3 [(gogoproto.customname) = "CodeID"];
+  // Label is optional metadata to be stored with a contract instance.
+  string label = 4;
+  // InitMsg json encoded message to be passed to the contract on instantiation
+  bytes init_msg = 5 [(gogoproto.casttype) = "encoding/json.RawMessage"];
+  // InitFunds coins that are transferred to the contract on instantiation
+  repeated cosmos.base.v1beta1.Coin init_funds = 6 [(gogoproto.nullable) = false, (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coins"];
+}
+// MsgInstantiateContractResponse return instantiation result data
+message MsgInstantiateContractResponse {
```
ethanfrey, line 63: I do like these clear types for return value

x/wasm/internal/types/tx.proto
```diff
@@ -0,0 +1,126 @@
...
+  string source = 3;
+  // Builder is a valid docker image name with tag, optional
+  string builder = 4;
+  // InstantiatePermission access control to apply on contract creation, optional
+  AccessConfig instantiate_permission = 5;
+}
+// MsgStoreCodeResponse returns store result data.
+message MsgStoreCodeResponse{
+  // CodeID is the reference to the stored WASM code
+  string code_id = 1  [(gogoproto.customname) = "CodeID"];
```
ethanfrey, line 44: If we get actual types, let's use int64 here

x/wasm/internal/types/tx.proto
```diff
@@ -0,0 +1,126 @@
...
+  string contract = 2;
+  // CodeID references the new WASM code
+  uint64 code_id = 3 [(gogoproto.customname) = "CodeID"];
+  // MigrateMsg json encoded message to be passed to the contract on migration
+  bytes migrate_msg = 4 [(gogoproto.casttype) = "encoding/json.RawMessage"];
+}
+
+// MsgMigrateContractResponse returns contract migration result data.
+message MsgMigrateContractResponse {
+  // Data contains base64-encoded bytes returned from the wasm contract.
```
ethanfrey, line 100: Why are they base64-encoded? We can return the raw bytes there, right? And let the proto-buf code handle base64 or other encoding

x/wasm/module_test.go
```diff
@@ -154,7 +154,9 @@ func TestHandleInstantiate(t *testing.T) {
 	}
 	res, err := h(data.ctx, msg)
 	require.NoError(t, err)
-	require.Equal(t, res.Data, []byte("1"))
+	var pStoreResp MsgStoreCodeResponse
```
ethanfrey, line 157: Yup, a few helpers would be nice for `assertStoreCodeResponse([]byte, uint64)` or such

ethanfrey (commented):
Nice work. A few comments on my pass through it.
I do think this is an improvement, and we can shove it into 0.14 so cosmjs works the same for 0.14 and 0.15

x/wasm/internal/keeper/msg_server.go
ethanfrey: I guess not: https://github.com/cosmos/cosmos-sdk/blob/master/x/bank/keeper/msg_server.go

x/wasm/genesis_test.go
ethanfrey: Added

x/wasm/internal/types/tx.proto
ethanfrey: I made this change

x/wasm/internal/types/tx.proto
ethanfrey: It actually returns the same raw bytes, updating the comment

x/wasm/module_test.go
ethanfrey: added

ethanfrey:
I am good to merge this. I need to investigate the how events are handled better in the baseapp and tendermint levels.

x/wasm/handler.go
ethanfrey: Saved follow up for #368 but not for 0.14.0 release

PR #386, opened 2021-01-24
Fix ParseEvents to always include wasm.contract_address

assafmo:


x/wasm/internal/types/types.go
```diff
@@ -219,20 +219,21 @@ func NewWasmCoins(cosmosCoins sdk.Coins) (wasmCoins []wasmvmtypes.Coin) {
...
+	// append attributes from wasm to the sdk.Event
+	for _, l := range wasmOutputAttrs {
 		// and reserve the contract_address key for our use (not contract)
 		if l.Key != AttributeKeyContractAddr {
 			attr := sdk.NewAttribute(l.Key, l.Value)
 			attrs = append(attrs, attr)
 		}
 	}
+	
+	// each wasm invokation always returns one sdk.Event
```
ethanfrey, line 236: This then seems like a TODO to return `sdk.Event` here.

ethanfrey (approved):
In general, this seems like a fine change.

I wonder about changing the return type to `sdk.Event` (1) rather than `[]sdk.Event` (0 or 1) as now

x/wasm/internal/types/types.go
assafmo: It's semantics IMO, because we'd have to wrap it within an array right after it returns:
- https://github.com/CosmWasm/wasmd/blob/b0b407d5f839cfa1c8f0bc7fce06032a3ca5c774/x/wasm/internal/keeper/keeper.go#L254-L255
- https://github.com/CosmWasm/wasmd/blob/b0b407d5f839cfa1c8f0bc7fce06032a3ca5c774/x/wasm/internal/keeper/keeper.go#L310-L311

ethanfrey: There is an `EmitEvent` call. But fair enough, leaving it as is for now.

PR #390, opened 2021-01-26
Update to Cosmos SDK 0.41

ethanfrey:
This will require a 0.15.0 release, as it breaks consensus (cannot be mixed with nodes running wasmd 0.14.1)


alpe:
👍  tested on my box

PR #395, opened 2021-01-28
Implement relayer integration tests

orkunkl:
This PR basically installs relayer on CI level, then executes the bash tests which located at `/integration`. These setup scripts are copied from relayer tests. For now the integration test setups up two wasm chains and sends token from one to another, then checks the balance. I made some changes on CI build. Not sure it will work as intended in the first try. Open for suggestions.


ethanfrey:
Hmmm... this assumes the relayer works?

If it were stable, I think this would be a very good test. As relayer is still somewhat buggy and not at 1.0, I would wait. This branch is a good start, but let's make sure the other side is stable before we test our code against it.

orkunkl:
@ethanfrey Actually relayer works with recently launched two local networks. So tests will pass

alpe:
I appreciate the effort ❤️    but I have mixed feelings about adding the relayer into automated CI tests.
One thing is adding and relying on a new dependency but under the hood what is it really testing that we have not covered by integration tests in our project or cosmos-sdk?
It is definitively a smoke test on the compiled binary but mostly the relayer I afraid! 
But before working with 2 chains we should have automated system tests for a single instance covering core wasmd functionality. There are some scripts in `contrib/local` for example addressing basic code upload/instantiation/execution/ queries... 

How to move forward? The scripts are adding value to the project and make it easier to get started for anybody testing out IBC. Running them on CI is a bit controversy. I would suggest to move them to `contrib/relayer-demo` or a similar path. WDYT?

orkunkl:
@alpe I agree these tests are bit shallow and insufficient. But at some point we will need IBC integration tests. We can leave the branch as it is, or exclude the tests from CI.

orkunkl:
Moved tests to `contrib`

alpe (approved):
Looks good.:tulip: 
A nice to have would be a README in the relayer-tests folder with thank you to the original authors and link to their repo.

PR #402, opened 2021-01-29
Add custom message handler option

### LINKED ISSUE #327: Support custom MessageHandler in wasm
An `interface` would be a great extension point to the system.

We provide a default implementation but people may want to use a different implementation in their system or for testing.
This is only a small change in [keeper](https://github.com/CosmWasm/wasmd/blob/master/x/wasm/internal/keeper/keeper.go#L54)

alpe:
Resolves #327


ethanfrey (approved):
Nice one. Just exposing this with the options. Such a great design

PR #403, opened 2021-02-03
Register v0.15.1 upgrade plan

orkunkl:
We need to register this upgrade plan to upgrade musselnet live. After this is merged we need to cut a new release. the tag naming can be different than `v0.15.1`. Let me know if you want to change it.

Re: tag and upgrade name will be `musselnet-3`


alpe (approved):
Looks good 👍 
fyi: I have merged master into the `musselnet` branch and modified this PR to point to the `musselnet` branch

PR #404, opened 2021-02-05
Remove redundant wasm bytes limit in rest

secret2830:
There exists a constant used for wasm bytes limit in rest, which is 400K.
 In fact, the size validation will be performed in `ValidateBasic` for `MsgStoreCode`, where the limit is 500K.

 I think the inconsistency is confusing.


alpe:
Good 👁️ . Thank you for the fix 🌻

PR #427, opened 2021-03-03
Add data field to Instantiate response

### LINKED ISSUE #385: Add MsgInstantiateContractResponse.data
[MsgExecuteContractResponse](https://github.com/CosmWasm/wasmd/blob/v0.14.1/x/wasm/internal/types/tx.proto#L80-L84) and [MsgMigrateContractResponse](https://github.com/CosmWasm/wasmd/blob/v0.14.1/x/wasm/internal/types/tx.proto#L98-L103) have a contract defined `data` field. [MsgInstantiateContractResponse](https://github.com/CosmWasm/wasmd/blob/v0.14.1/x/wasm/internal/types/tx.proto#L62-L66) does not have this for historic reasons (we returned the raw contract address before).

Now that we can easily return structured proto messages, it would be good to add `MsgInstantiateContractResponse.data` for consistency.

alpe:
Resolves #385 

New field in `MsgInstantiateContractResponse` and minor refactoring to testing framework.


x/wasm/internal/types/tx.proto
```diff
@@ -63,6 +63,8 @@ message MsgInstantiateContract {
 message MsgInstantiateContractResponse {
   // Address is the bech32 address of the new contract instance.
   string address = 1;
+  // Data contains base64-encoded bytes to returned from the contract
+  bytes data = 3;
```
ethanfrey, line 67: Why 3? and not 2?

ethanfrey (approved):
Looks good. One little nit.

I think you have a good point of the data not being returned for any of the calls via governance. I would address that in another PR, as that is not just about instantiate.

alpe:
I have created https://github.com/CosmWasm/wasmd/issues/431 to follow up on the events

alpe:
Sorry, had to push -f after merge conflicts.

ethanfrey:
> I have created #431 to follow up on the events

Great. and for 0.17 :smile:  (Actually events are non-breaking, so that could be 0.16.1).

Feel free to merge when CI finally runs

PR #429, opened 2021-03-03
Rename to funds in msgs

### LINKED ISSUE #423: Rename MsgInstantiateContract.init_funds/MsgExecuteContract.sent_funds to just funds
Both fields end up in the same structure in wasmvm, where the field name has been simplified to just `funds` over the last weeks. We should do that here too for consistentcy. This would also better reflect that both fields do the same thing.

- [MsgInstantiateContract.init_funds](https://github.com/CosmWasm/wasmd/blob/v0.15.1/x/wasm/internal/types/tx.proto#L59-L60)
- [MsgExecuteContract.sent_funds](https://github.com/CosmWasm/wasmd/blob/v0.15.1/x/wasm/internal/types/tx.proto#L76-L77)

alpe:
Resolves #423 

Note: To be consistent with the renaming in messages the `init_funds` in `InstantiateProposal` was renamed as well.


ethanfrey (approved):
Nice

PR #430, opened 2021-03-03
Use Label from contract's instantiate message

### LINKED ISSUE #376: Use label from WasmMsg:::Init if present
See the TODO at 
https://github.com/CosmWasm/wasmd/blob/v0.14.0/x/wasm/internal/keeper/handler_plugin.go#L187-L188

alpe:
Resolves #376 

This also removes the default label. As labels are required the contracts need to set them proper


ethanfrey (approved):
After all that discussion, such a simple fix.

Looks good

Issue #400, opened 2021-01-29
Upgrade to wasmvm 0.14 when released


ethanfrey:
we currently point to the `0.14-dev` branch. When this is finalized, we need to point to a tagged release on master.

Also, we need to update all test contracts to the proper 0.14 CosmWasm release

PR #432, opened 2021-03-03
Wasmvm 0.14 alpha2

ethanfrey:
Part of #400 

Upgraded wasmvm and the test contracts, fixed compile issues and failing tests. Still need to enable the new callbacks (in separate issues).

This exposed https://github.com/CosmWasm/wasmvm/issues/200 which is worked around now. Until then, you cannot use `supportedFeatures = ""`


alpe (approved):
🌷 Awesome

PR #433, opened 2021-03-03
Sudo entry point

### LINKED ISSUE #420: Expose new "system" entry point for contracts
This is optional and has the same function signature as Migrate. It is intended as an entry point for arbitrary priviledged calls, which is an extension point to be used by blockchain apps (not used in wasmd outside of tests)

ethanfrey:
Closes #420 

Merge after #432 (this is based on that)

Note the entry point name changed from system -> sudo after some discussions in CosmWasm

### DIFF
--- a/x/wasm/internal/keeper/keeper_test.go
+++ b/x/wasm/internal/keeper/keeper_test.go
@@ -1025,6 +1025,66 @@ func TestMigrateWithDispatchedMessage(t *testing.T) {
 	assert.Equal(t, deposit, balance)
 }
 
+type sudoMsg struct {
+	StealFunds stealFundsMsg `json:"StealFunds"`
+}
+
+type stealFundsMsg struct {
+	Recipient string            `json:"recipient"`
+	Amount    wasmvmtypes.Coins `json:"amount"`
+}
+
+func TestSudo(t *testing.T) {
+	ctx, keepers := CreateTestInput(t, false, SupportedFeatures, nil, nil)
+	accKeeper, keeper, bankKeeper := keepers.AccountKeeper, keepers.WasmKeeper, keepers.BankKeeper
+
+	deposit := sdk.NewCoins(sdk.NewInt64Coin("denom", 100000))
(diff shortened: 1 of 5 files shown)


x/wasm/internal/keeper/keeper_test.go
```diff
@@ -1025,6 +1025,66 @@ func TestMigrateWithDispatchedMessage(t *testing.T) {
...
+	require.Equal(t, "cosmos18vd8fpwxzck93qlwghaj6arh4p7c5n89uzcee5", addr.String())
+
+	// the community is broke
+	_, _, community := keyPubAddr()
+	comAcct := accKeeper.GetAccount(ctx, community)
+	require.Nil(t, comAcct)
+
+	// now the community wants to get paid via sudo
+	msg := sudoMsg{
+		StealFunds: stealFundsMsg{
```
alpe, line 1069: Not sure if this is the best use case of this function 🤣 

👮👮👮 But more serious, people tend to not understand jokes very well when it comes to money or security. Maybe some additional context can help tot explain that this is not about stealing money but calling a privileged function in the contract that would otherwise not be available. This example would not work with random contract but only with the hackatom example that had this `stealFundsMsg` function added for this particular demo case. We do not encourage people to build backdoors into their contracts... 👮 👮👮

🤣 🤣 🤣

alpe (approved):
Very nice 👍 . Only minor nit on the test example

x/wasm/internal/keeper/keeper_test.go
ethanfrey: I will add a comment. I added it in CosmWasm as a funny example.
Our first tutorial of cosmwasm was to modify the escrow contract to add a backdoor. And then said this is why it is essential you verify the actual rust code.

ethanfrey: Rebased and added a comment in https://github.com/CosmWasm/wasmd/pull/433/commits/311be37e90f8932d8754f53612732e87f9bb18ba

PR #435, opened 2021-03-04
Stargate msg and query

### LINKED ISSUE #388: Add support for Stargate CosmosMsg/QueryRequest
In master branch (v0.14-dev) of CosmWasm, we add a type for dispatching a protobuf-encoded "stargate message", or making a protobuf-encoded query to allow quick use of any mdoule that doesn't have a custom handler.

This is exposed in `wasmvm` go types and the most recent `reflect` contract, but we need to add dispatch support to `wasmd` and test it.

ethanfrey:
Closes #388 

- [x] Implement StargateMsgEncoder
- [x] Implement StargateQueryHandler
- [x] Test message encoder
- [x] Test query handler

### DIFF
--- a/x/wasm/internal/keeper/handler_plugin.go
+++ b/x/wasm/internal/keeper/handler_plugin.go
@@ -2,8 +2,10 @@ package keeper
 
 import (
 	"encoding/json"
+	"fmt"
 	"github.com/CosmWasm/wasmd/x/wasm/internal/types"
 	wasmvmtypes "github.com/CosmWasm/wasmvm/types"
+	codectypes "github.com/cosmos/cosmos-sdk/codec/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
@@ -20,8 +22,8 @@ type DefaultMessageHandler struct {
 	encoders MessageEncoders
 }
 
-func NewDefaultMessageHandler(router sdk.Router, channelKeeper types.ChannelKeeper, capabilityKeeper types.CapabilityKeeper, customEncoders *MessageEncoders) DefaultMessageHandler {
-	encoders := DefaultEncoders(channelKeeper, capabilityKeeper).Merge(customEncoders)
+func NewDefaultMessageHandler(router sdk.Router, channelKeeper types.ChannelKeeper, capabilityKeeper types.CapabilityKeeper, unpacker codectypes.AnyUnpacker, customEncoders *MessageEncoders) DefaultMessageHandler {
+	encoders := DefaultEncoders(channelKeeper, capabilityKeeper, unpacker).Merge(customEncoders)
 	return DefaultMessageHandler{
 		router:   router,
 		encoders: encoders,
@@ -44,13 +46,13 @@ type MessageEncoders struct {
 	Wasm     WasmEncoder
 }
 
-func DefaultEncoders(channelKeeper types.ChannelKeeper, capabilityKeeper types.CapabilityKeeper) MessageEncoders {
+func DefaultEncoders(channelKeeper types.ChannelKeeper, capabilityKeeper types.CapabilityKeeper, unpacker codectypes.AnyUnpacker) MessageEncoders {
 	return MessageEncoders{
 		Bank:     EncodeBankMsg,
 		Custom:   NoCustomMsg,
 		IBC:      EncodeIBCMsg(channelKeeper, capabilityKeeper),
 		Staking:  EncodeStakingMsg,
-		Stargate: EncodeStargateMsg,
+		Stargate: EncodeStargateMsg(unpacker),
 		Wasm:     EncodeWasmMsg,
 	}
 }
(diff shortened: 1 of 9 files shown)


x/wasm/internal/keeper/handler_plugin.go
```diff
@@ -20,8 +22,8 @@ type DefaultMessageHandler struct {
 	encoders MessageEncoders
 }
 
-func NewDefaultMessageHandler(router sdk.Router, channelKeeper types.ChannelKeeper, capabilityKeeper types.CapabilityKeeper, customEncoders *MessageEncoders) DefaultMessageHandler {
-	encoders := DefaultEncoders(channelKeeper, capabilityKeeper).Merge(customEncoders)
+func NewDefaultMessageHandler(router sdk.Router, channelKeeper types.ChannelKeeper, capabilityKeeper types.CapabilityKeeper, unpacker codectypes.AnyUnpacker, customEncoders *MessageEncoders) DefaultMessageHandler {
```
alpe, line 25: good use of the minimal interface

x/wasm/internal/keeper/handler_plugin.go
```diff
@@ -170,6 +180,20 @@ func EncodeStakingMsg(sender sdk.AccAddress, msg *wasmvmtypes.StakingMsg) ([]sdk
...
+	return func(sender sdk.AccAddress, msg *wasmvmtypes.StargateMsg) ([]sdk.Msg, error) {
+		any := codectypes.Any{
+			TypeUrl: msg.TypeURL,
+			Value:   msg.Value,
+		}
+		var sdkMsg sdk.Msg
+		if err := unpacker.UnpackAny(&any, &sdkMsg); err != nil {
+			return nil, sdkerrors.Wrap(types.ErrInvalidMsg, fmt.Sprintf("Cannot unpack proto message with type URL: %s", msg.TypeURL))
+		}
+		return []sdk.Msg{sdkMsg}, nil
```
alpe, line 193: The `sdkMsg` can contain `Any` types as well that need to be unpacked. This is not done recursively (https://github.com/cosmos/cosmos-sdk/blob/v0.41.4/codec/types/compat.go#L197).
You need another call to `codectypes.UnpackInterfaces(sdkMsg, unpacker)`

ethanfrey: I did this here: https://github.com/CosmWasm/wasmd/pull/435/commits/c33b3694acef890e5754e2e0e75ac0bf9a701a14

Can you check that was correct?

alpe: ja, looks good!
btw. you can use `sdkerrors.Wrapf` to save you some `fmt.Sprintf`

x/wasm/internal/keeper/handler_plugin_test.go
```diff
@@ -276,6 +288,27 @@ func TestEncoding(t *testing.T) {
 				},
 			},
 		},
+		// TODO: alpe? can you add an example with sub-interfaces (where the UnpackInterfaces call would be needed)
```
ethanfrey, line 291: I am not sure how to create such a case. (I also think we only have bank, staking, and distribution registered here in the tests).

If you could add such a test to trigger this that would be great. Or if it is a pain, please open up an issue on it and explain more or less what is needed. I can put that down lower on the priority list

ethanfrey (commented):
@alpe this should be ready now. I'd like a review, especially on the proto part and how I wire it with the app

x/wasm/internal/keeper/handler_plugin_test.go
alpe: I found `Any` types in sdk (master). 
* MsgGrantAuthorizationRequest
* MsgGrantFeeAllowance
* MsgSubmitProposal
* MsgCreateValidator
I guess `MsgSubmitProposal` would be a good candidate. I can add a test.

x/wasm/internal/keeper/reflect_test.go
```diff
@@ -296,6 +297,79 @@ func TestMaskReflectCustomQuery(t *testing.T) {
...
+	mustParse(t, simpleChain.Data, &simpleBalance)
+	require.Equal(t, len(expectedBalance), len(simpleBalance.Amount))
+	assert.Equal(t, simpleBalance.Amount[0].Amount, expectedBalance[0].Amount.String())
+	assert.Equal(t, simpleBalance.Amount[0].Denom, expectedBalance[0].Denom)
+
+	// now, try to build a protobuf query
+	protoQuery := banktypes.QueryAllBalancesRequest{
+		Address: creator.String(),
+	}
+	protoQueryBin, err := proto.Marshal(&protoQuery)
```
alpe, line 348: Clever to test with a different golang protobuf marshaller.  🌻

x/wasm/internal/keeper/reflect_test.go
```diff
@@ -296,6 +297,79 @@ func TestMaskReflectCustomQuery(t *testing.T) {
...
+	})
+	require.NoError(t, err)
+	simpleRes, err := keeper.QuerySmart(ctx, contractAddr, simpleQueryBz)
+	require.NoError(t, err)
+	var simpleChain ChainResponse
+	mustParse(t, simpleRes, &simpleChain)
+	var simpleBalance wasmvmtypes.AllBalancesResponse
+	mustParse(t, simpleChain.Data, &simpleBalance)
+	require.Equal(t, len(expectedBalance), len(simpleBalance.Amount))
+	assert.Equal(t, simpleBalance.Amount[0].Amount, expectedBalance[0].Amount.String())
```
alpe, line 341: nit: no need to change: wrong order of arguments. Here and others (t, expect, got)  (to be fair it is not consistent within `testify` like assert.Len(t, got, exp)` )

alpe (approved):
Great work. This will add a very valuable feature to the system. 🚀

ethanfrey:
Thanks. If you could add a follow up pr sometime with submit proposal, that would be great 

Now contracts can vote

PR #436, opened 2021-03-04
Start Pin/Unpin contract

### LINKED ISSUE #401: Expose contract pinning
In current CosmWasm master, wasmvm `0.14-dev`, they expose an API to pin/unpin contracts by Checksum. If a contract is pinned, it is guaranteed to be in memory cache and we can reduce the instantiation cost significantly:

* Expose keeper API to pin/unpin contracts
* Store pinned contracts in KVStore and re-pin on restart (can we just lazily pin everytime we run them, assuming that this is near zero cost if they are already pinned?)
* Lower fix gas cost of calling a pinned contract from 40k to 1k (much faster on all machines)
* Write some benchmarks that show calling pinned contracts is much faster (ensuring we are using APIs correctly). They can be Keeper level, and we can set the LRU cache size to 0 to ensure no other caching is going on.

alpe:
Resolves #401 

* Add Pin/Unpin/IsPinned methods
* Persist pinned code id state as second index
* Export/ Import pinned flag via genesis
* Extend engineMock
* Gov proposal types
* Benchmarks


x/wasm/internal/types/wasmer_engine.go
```diff
@@ -197,4 +197,15 @@ type WasmerEngine interface {
 		gasMeter wasmvm.GasMeter,
 		gasLimit uint64,
 	) (*wasmvmtypes.IBCBasicResponse, uint64, error)
+
+	// Pin pins a code to an in-memory cache, such that is
+	// always loaded quickly when executed.
+	// Pin is idempotent.
+	Pin(checksum wasmvm.Checksum) error
```
ethanfrey, line 204: :+1:

x/wasm/internal/keeper/wasmtesting/mock_engine.go
```diff
@@ -138,6 +140,21 @@ func (m *MockWasmer) Cleanup() {
 	m.CleanupFn()
 }
 
+func (m *MockWasmer) Pin(checksum wasmvm.Checksum) error {
```
ethanfrey, line 143: Let's just make these noops? Or counters?
I guess this is consistent, but feel it is easier to do nothing here

x/wasm/internal/keeper/keeper.go
```diff
@@ -693,6 +694,54 @@ func (k Keeper) dispatchMessages(ctx sdk.Context, contractAddr sdk.AccAddress, i
...
+	}
+
+	store := ctx.KVStore(k.storeKey)
+	store.Delete(types.GetPinnedContractIndexPrefix(contractAddr))
+	return nil
+}
+
+func (k Keeper) IsPinnedContract(ctx sdk.Context, contractAddr sdk.AccAddress) bool {
+	store := ctx.KVStore(k.storeKey)
+	return store.Has(types.GetPinnedContractIndexPrefix(contractAddr))
```
ethanfrey, line 726: Okay, this is read from store everytime, right? (My proposal) I thought you said there was an in-memory cache

x/wasm/internal/keeper/keeper.go
```diff
@@ -693,6 +694,54 @@ func (k Keeper) dispatchMessages(ctx sdk.Context, contractAddr sdk.AccAddress, i
...
+}
+
+// InitializePinnedContracts updates wasmvm to pin to cache all contracts marked as pinned
+func (k Keeper) InitializePinnedContracts(ctx sdk.Context) error {
```
ethanfrey, line 730: Ah, this is the difference, you don't need to call `Pin` everytime, but you do read the store to check for gas.

Yeah, this works well. And we can try to cache the state for `IsPinnedContract` query in 0.17

ethanfrey (commented):
Nice start, a few comments

x/wasm/internal/keeper/keeper.go
alpe: I had planned to do benchmarks first before adding the in memory store.

x/wasm/internal/keeper/wasmtesting/mock_engine.go
alpe: No test to touch for this (yet)

x/wasm/internal/keeper/keeper.go
```diff
@@ -536,12 +543,14 @@ func (k Keeper) GetContractHistory(ctx sdk.Context, contractAddr sdk.AccAddress)
...
 // QuerySmart queries the smart contract itself.
 func (k Keeper) QuerySmart(ctx sdk.Context, contractAddr sdk.AccAddress, req []byte) ([]byte, error) {
-	ctx.GasMeter().ConsumeGas(InstanceCost, "Loading CosmWasm module: query")
-
-	_, codeInfo, prefixStore, err := k.contractInstance(ctx, contractAddr)
+	contractInfo, codeInfo, prefixStore, err := k.contractInstance(ctx, contractAddr)
 	if err != nil {
 		return nil, err
 	}
+	if !k.IsPinnedCode(ctx, contractInfo.CodeID) {
```
ethanfrey, line 550: nice. you got them all

x/wasm/internal/keeper/keeper.go
```diff
@@ -693,6 +702,58 @@ func (k Keeper) dispatchMessages(ctx sdk.Context, contractAddr sdk.AccAddress, i
...
+	codeInfo := k.GetCodeInfo(ctx, codeID)
+	if codeInfo == nil {
+		return sdkerrors.Wrap(types.ErrNotFound, "code info")
+	}
+
+	if err := k.wasmer.Pin(codeInfo.CodeHash); err != nil {
+		return sdkerrors.Wrap(types.ErrPinContractFailed, err.Error())
+	}
+	store := ctx.KVStore(k.storeKey)
+	store.Set(types.GetPinnedCodeIndexPrefix(codeID), []byte{})
```
ethanfrey, line 716: Oof, the handling of `[]byte{}` and `nil` inside `store.Set()` is a bit of a headache. Has *should* return true, but Get will return nil. 

Please just make this `[]byte{1}` to avoid potential issues later

x/wasm/internal/keeper/keeper.go
```diff
@@ -693,6 +702,58 @@ func (k Keeper) dispatchMessages(ctx sdk.Context, contractAddr sdk.AccAddress, i
...
+
+	store := ctx.KVStore(k.storeKey)
+	store.Delete(types.GetPinnedCodeIndexPrefix(codeID))
+	return nil
+}
+
+// IsPinnedCode returns true when codeID is pinned in wasmvm cache
+func (k Keeper) IsPinnedCode(ctx sdk.Context, codeID uint64) bool {
+	store := ctx.KVStore(k.storeKey)
+	return store.Has(types.GetPinnedCodeIndexPrefix(codeID))
```
ethanfrey, line 738: Okay, this should be fine with `[]byte{}`, but I don't like to test fate.
Lots of memories of KVStore debugging on this one

x/wasm/internal/keeper/keeper.go
```diff
@@ -693,6 +702,58 @@ func (k Keeper) dispatchMessages(ctx sdk.Context, contractAddr sdk.AccAddress, i
...
+}
+
+// InitializePinnedCodes updates wasmvm to pin to cache all contracts marked as pinned
+func (k Keeper) InitializePinnedCodes(ctx sdk.Context) error {
+	store := prefix.NewStore(ctx.KVStore(k.storeKey), types.PinnedCodeIndexPrefix)
+	iter := store.Iterator(nil, nil)
+	for ; iter.Valid(); iter.Next() {
+		codeInfo := k.GetCodeInfo(ctx, sdk.BigEndianToUint64(iter.Value()))
```
ethanfrey, line 746: This is the inverse of `types.GetPinnedCodeIndexPrefix()`, right?
Maybe you can declare a function `types.ParsePinnedCodeIndexPrefix()` or something, so have the encode/decode steps next to each other

x/wasm/internal/types/keys.go
```diff
@@ -96,3 +97,12 @@ func GetContractCodeHistoryElementPrefix(contractAddr sdk.AccAddress) []byte {
 	copy(r[prefixLen:], contractAddr)
 	return r
 }
+
+// GetPinnedCodeIndexPrefix returns the key prefix for a code id pinned into the wasmvm cache
+func GetPinnedCodeIndexPrefix(codeID uint64) []byte {
+	prefixLen := len(PinnedCodeIndexPrefix)
+	r := make([]byte, prefixLen+sdk.AddrLen)
```
ethanfrey, line 104: Uh, this should be prefixLen + 8, right? (64 bytes)

ethanfrey (commented):
Nice stuff.

I would add a few more tests, but only a few stylistic nits in the code.

x/wasm/internal/types/keys.go
alpe: Ups, right. Leftover from previous version 🙈

x/wasm/internal/keeper/bench_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+import (
+	"github.com/CosmWasm/wasmd/x/wasm/internal/types"
+	"github.com/stretchr/testify/require"
+	"github.com/syndtr/goleveldb/leveldb/opt"
+	dbm "github.com/tendermint/tm-db"
+	"testing"
+)
+
+func BenchmarkExecution(b *testing.B) {
```
ethanfrey, line 11: I'd love to see the numbers from this on, say, wasmd 0.12 (which had the old wasmer engine), so we can compare the speedup for uncached over the last 3-4 months, plus the huge speedup for pinned contracts

x/wasm/internal/keeper/proposal_handler.go
```diff
@@ -171,3 +187,49 @@ func handleClearAdminProposal(ctx sdk.Context, k Keeper, p types.ClearAdminPropo
 	ctx.EventManager().EmitEvent(ourEvent)
 	return nil
 }
+
+func handlePinCodesProposal(ctx sdk.Context, k governing, p types.PinCodesProposal) error {
+	if err := p.ValidateBasic(); err != nil {
+		return err
+	}
+	for _, v := range p.CodeIDs {
```
ethanfrey, line 195: nice, i like the list of code ids

x/wasm/internal/keeper/proposal_integration_test.go
```diff
@@ -346,3 +349,178 @@ func TestUpdateParamsProposal(t *testing.T) {
...
+
+			// when stored
+			storedProposal, gotErr := govKeeper.SubmitProposal(ctx, &proposal)
+			if spec.expErr {
+				require.Error(t, gotErr)
+				return
+			}
+			require.NoError(t, gotErr)
+
+			// and proposal execute
```
ethanfrey, line 427: nice shortcut to mock out the whole voting process

x/wasm/internal/keeper/test_common.go
```diff
@@ -388,12 +405,12 @@ func handleExecute(ctx sdk.Context, k *Keeper, msg *types.MsgExecuteContract) (*
 	return res, nil
 }
 
-func RandomAccountAddress(_ *testing.T) sdk.AccAddress {
+func RandomAccountAddress(_ TestingT) sdk.AccAddress {
```
ethanfrey, line 408: There is a `testing.TB` interface I have seen before that probably is what you want here - one interface that lets you use the code both in unit tests as well as benchmarks.

ethanfrey (approved):
Great work.

I added one minor comment you may want to integrate later (testing.TB), but this looks great. I'm going to merge it in and start rebasing other PRs on top.

PR #437, opened 2021-03-04
Encode WasmMsg::Migrate and test

### LINKED ISSUE #422: Add support for WasmMsg::Migrate
One contract (eg multisig) can now request to migrate another contract. This must meet the same ContractInfo.Admin check as any public transaction (not bypass it like governance). Otherwise, this is just one more message to parse and dispatch

ethanfrey:
Closes #422


alpe (approved):
Small and 🚀

PR #439, opened 2021-03-04
Ibc query support

### LINKED ISSUE #434: Add IBCQuery support
While working on #388 I saw we do not handle QueryRequest.IBC variants. We need to add that to fully enable IBC contracts.

ethanfrey:
Closes #434

Code "complete", but needs testing

The `IbcQuery.Channels` variant doesn't seem to match what we can get from the IBC Keeper (we cannot filter by port). This will be very expensive to call.

Let us reconsider what we want to do here. We can still modify the queries we expose to the contracts (remove or adjust them). CosmWasm is still in `-alpha2`, so we can change types.

### DIFF
--- a/x/wasm/internal/keeper/query_plugins.go
+++ b/x/wasm/internal/keeper/query_plugins.go
@@ -173,15 +173,24 @@ func IBCQuerier(wasm *Keeper, channelKeeper types.ChannelKeeper) func(ctx sdk.Co
 			}
 			return json.Marshal(res)
 		}
+		// I just discovered we cannot query by port (how odd since they store it like this)
+		// Anyway, I will allow Port as an optional filter
+		// TODO: let's revisit this query (if needed, if we should change it)
 		if request.ListChannels != nil {
 			portID := request.ListChannels.PortID
-			if portID == "" {
-				contractInfo := wasm.GetContractInfo(ctx, caller)
-				portID = contractInfo.IBCPortID
-			}
-			// TODO: query the channels for this port
+			var channels wasmvmtypes.IBCEndpoints
+			channelKeeper.IterateChannels(ctx, func(ch types.IdentifiedChannel) bool {
+				if portID == "" || portID == ch.PortId {
+					newChan := wasmvmtypes.IBCEndpoint{
+						PortID:    ch.PortId,
+						ChannelID: ch.ChannelId,
+					}
+					channels = append(channels, newChan)
+				}
+				return false
+			})
 			res := wasmvmtypes.ListChannelsResponse{
-				Channels: wasmvmtypes.IBCEndpoints{},
+				Channels: channels,
 			}
 			return json.Marshal(res)
 		}
--- a/x/wasm/internal/keeper/query_plugins.go
+++ b/x/wasm/internal/keeper/query_plugins.go
@@ -3,6 +3,7 @@ package keeper
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/CosmWasm/wasmd/x/wasm/internal/types"
 
 	wasmvmtypes "github.com/CosmWasm/wasmvm/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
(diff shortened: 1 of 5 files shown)


x/wasm/internal/keeper/query_plugins.go
```diff
@@ -146,6 +164,71 @@ func NoCustomQuerier(sdk.Context, json.RawMessage) ([]byte, error) {
...
+		if request.PortID != nil {
+			contractInfo := wasm.GetContractInfo(ctx, caller)
+			res := wasmvmtypes.PortIDResponse{
+				PortID: contractInfo.IBCPortID,
+			}
+			return json.Marshal(res)
+		}
+		// I just discovered we cannot query by port (how odd since they store it like this)
+		// Anyway, I will allow Port as an optional filter
+		// TODO: let's revisit this query (if needed, if we should change it)
```
alpe, line 178: The code looks good. Is gas cost your concern with the iterator?
I wonder if we can abort the iterator early when > portID ?

Let's avoid todos in code and have an issue instead.

x/wasm/internal/keeper/query_plugins.go
```diff
@@ -146,6 +164,71 @@ func NoCustomQuerier(sdk.Context, json.RawMessage) ([]byte, error) {
...
+				return false
+			})
+			res := wasmvmtypes.ListChannelsResponse{
+				Channels: channels,
+			}
+			return json.Marshal(res)
+		}
+		if request.Channel != nil {
+			channelID := request.Channel.ChannelID
+			_ = channelID
```
alpe, line 199: nit: `_` unused

x/wasm/internal/keeper/query_plugins.go
```diff
@@ -146,6 +164,71 @@ func NoCustomQuerier(sdk.Context, json.RawMessage) ([]byte, error) {
...
+						PortID:    portID,
+						ChannelID: channelID,
+					},
+					CounterpartyEndpoint: wasmvmtypes.IBCEndpoint{
+						PortID:    got.Counterparty.PortId,
+						ChannelID: got.Counterparty.ChannelId,
+					},
+					Order:               got.Ordering.String(),
+					Version:             got.Version,
+					CounterpartyVersion: got.Version, // FIXME: maybe we change the type? This data is not stored AFAIK
```
alpe, line 219: ja, it is not stored with channel anymore. Too bad. We need the field during IBC handshake only. Otherwise it is empty.

alpe (approved):
Minor comments only. Looks good 🌻

x/wasm/internal/keeper/query_plugins.go
ethanfrey: Gas cost is my issue. This could become expensive and is always O(total channels in the system). 

My main note was that our query does not align with what the sdk exposed, so I'd rather redefine this to never take a port (since we cannot filter by port) and then it is just as inefficient as the description is. But I wanted feedback before changing behavior.

x/wasm/internal/keeper/query_plugins.go
ethanfrey: Good point, left from a WIP version so it would compile

x/wasm/internal/keeper/query_plugins.go
ethanfrey: Shall I remove this from CosmWasm return value?

ethanfrey:
I'd like to make a few changes to the query types as mentioned above. Do you think those are good?

* Remove the port argument to Channels query (as we basically ignore it, filtering could be done just as easy in wasm).
* Include full channel info in the list query (we have it all)
* Remove the Counterparty Version info from the ChannelInfo

x/wasm/internal/keeper/query_plugins.go
alpe: If we can not abort the iterator early then I agree that there is no benefit to filter. Let's go for the simplest solution.

x/wasm/internal/keeper/query_plugins.go
alpe: `CounterpartyVersion` should be empty string here to be consistent with what we do during IBC handshake when the data is not available. The field has value during the handshake.

alpe:
> I'd like to make a few changes to the query types as mentioned above. Do you think those are good?
> 
> * ~Remove the port argument to Channels query (as we basically ignore it, filtering could be done just as easy in wasm).~ (Keep it there and fake sdk side, future versions may provide better queries)
👌 

> * Include full channel info in the list query (we have it all)
👌 Can save further queries

> * Remove the Counterparty Version info from the ChannelInfo
🙅 We need it during IBC handshake. Maybe we make it a pointer type to communicate that it is not always set.

PR #441, opened 2021-03-05
Submsg and replies

### LINKED ISSUE #421: Add support for submessages/callback
This is based on CosmWasm issue https://github.com/CosmWasm/cosmwasm/issues/691 and PR https://github.com/CosmWasm/cosmwasm/pull/796. 

Note: this is intended for v0.16.0. It is also a large change and we could wait until v0.17.0 to implement it and for now just throw "unimplemented" error if any contract returns subcall. Let's evaluate this one well.

We need to add a new `subcall_response` entry point similar to #420, however, this entry point is only used internally in the keeper and not exported to any other module.

We also need to handle submessages in the Response. These have the following semantics:

- Store the parent context
- Wrap the current context/storage in a new context (with CacheWrap()) 
- If a gas limit is set, and limit is less than current gas used, create a new gas meter with smaller limit and update context
- Call the message in this new context (just as we handle messages)
- If we set a custom gas limit:
  - If we get an OutOfGas error, charge the parent context the full gas allowance of this subcommand and return error
  - Otherwise, charge the parent context the actual gas used
  - If either cause an OutOfGas on the parent context, we don't catch that and allow it to abort the full connection
- Check result and handle storage:
  - If success, Commit the CacheWrap to the parent storage
  - If error, "rollback" (drop) the CacheWrap
- Return the result to the calling contract via new `subcall_response` entry point
  - If error, return a string error result
  - If success, return all events and data set in the submessage
    - Also add these events to the parent transaction (to return to original caller, just as normal messages)
    - We **do not** charge the instantiation cost (40.000 gas) on this `subcall_response` as the parent is assumed to be cached.
- Process `subcall_response` result just like any other result
  - note that `subcall_response` from message 1 is processed before message 2
  - note all submessages (and their responses) are processed before all normal messages

You can test this with the reflect contract, with has a new `HandleMsg::ReflectSubCall` variant to test this, and just stores the `subcall_response` in the storage for later queries (and test assertions)

ethanfrey:
Closes #421 

- [x] Add private method to call reply entry point
- [x] Wrote logic to dispatch submessages
- [x] Dispatch submessages from all (non-ibc) entrypoints (see note below)
- [x] Test submessage handling via reflect contract
  - [x] no gas limit, success
  - [x] no gas limit, failure
  - [x] no gas limit, out-of-gas panic
  - [x] gas limit, success
  - [x] gas limit, failure
  - [x] gas limit, consume all gas
  - [x] Instantiate Contract, get address from reply.Data

Note: we will need to update again after https://github.com/CosmWasm/cosmwasm/issues/822 to add support for submessages in the ibc callbacks

### DIFF
--- a/x/wasm/internal/keeper/keeper.go
+++ b/x/wasm/internal/keeper/keeper.go
@@ -482,12 +482,14 @@ func (k Keeper) Sudo(ctx sdk.Context, contractAddress sdk.AccAddress, msg []byte
 // reply is only called from keeper internal functions (dispatchSubmessages) after processing the submessage
 // it
 func (k Keeper) reply(ctx sdk.Context, contractAddress sdk.AccAddress, reply wasmvmtypes.Reply) (*sdk.Result, error) {
-	// we can assume this is still in cache and not charge again for returning the reply. correct?
-	//ctx.GasMeter().ConsumeGas(InstanceCost, "Loading CosmWasm module: reply")
-
 	contractInfo, codeInfo, prefixStore, err := k.contractInstance(ctx, contractAddress)
 	if err != nil {
 		return nil, err
+	}
+
+	// current thought is to charge gas like a fresh run, we can revisit whether to give it a discount later
+	if !k.IsPinnedCode(ctx, contractInfo.CodeID) {
+		ctx.GasMeter().ConsumeGas(InstanceCost, "Loading CosmWasm module: reply")
 	}
 
 	env := types.NewEnv(ctx, contractAddress)
--- a/x/wasm/internal/keeper/keeper.go
+++ b/x/wasm/internal/keeper/keeper.go
@@ -798,6 +798,31 @@ func (k Keeper) dispatchMessages(ctx sdk.Context, contractAddr sdk.AccAddress, i
 	return nil
 }
 
+func (k Keeper) dispatchMsgWithGasLimit(ctx sdk.Context, contractAddr sdk.AccAddress, ibcPort string, msg wasmvmtypes.CosmosMsg, gasLimit uint64) (events []sdk.Event, data [][]byte, err error) {
+	// FIXME: do I need to wrap the BlockGasMeter in the same way?
+	limitedMeter := sdk.NewGasMeter(gasLimit)
+	subCtx := ctx.WithGasMeter(limitedMeter)
+
+	// catch out of gas panic and just charge the entire gas limit
+	defer func(){
+		fmt.Println("Got gas limit defer")
+		if r := recover(); r != nil {
+			fmt.Println("Recovering...")
+			// TODO: check for out of gas, not any panic
+			//ctx.GasMeter().ConsumeGas(gasLimit, "Sub-Message OutOfGas panic")
+			err = sdkerrors.Wrap(sdkerrors.ErrorInvalidGasAdjustment, "SubMsg hit gas limit")
+			fmt.Println("Finished recovering...")
+		}
+	}()
+	events, data, err = k.messenger.DispatchMsg(ctx, contractAddr, ibcPort, msg)
+
+	// make sure we charge the parent what was spent
+	spent := subCtx.GasMeter().GasConsumed()
+	ctx.GasMeter().ConsumeGas(spent, "From limited Sub-Message")
+
+	return events, data, err
+}
+
 // dispatchSubmessages builds a sandbox to execute these messages and returns the execution result to the contract
 // that dispatched them, both on success as well as failure
 func (k Keeper) dispatchSubmessages(ctx sdk.Context, contractAddr sdk.AccAddress, ibcPort string, msgs []wasmvmtypes.SubMsg) error {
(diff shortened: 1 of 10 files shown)


x/wasm/internal/keeper/keeper.go
```diff
@@ -469,7 +473,48 @@ func (k Keeper) Sudo(ctx sdk.Context, contractAddress sdk.AccAddress, msg []byte
...
+
+	return &sdk.Result{
+		Data: res.Data,
+	}, nil
+}
+
+// reply is only called from keeper internal functions (dispatchSubmessages) after processing the submessage
+// it
+func (k Keeper) reply(ctx sdk.Context, contractAddress sdk.AccAddress, reply wasmvmtypes.Reply) (*sdk.Result, error) {
+	// we can assume this is still in cache and not charge again for returning the reply. correct?
```
alpe, line 490: The LRU cache might be disabled via `--wasm.memory_cache_size` start parameter.
I have mixed feelings not charging anything or charging the full instantiation fee (again). Let's discuss in a bigger group.

x/wasm/internal/keeper/keeper.go
```diff
@@ -684,15 +729,134 @@ func (k Keeper) GetByteCode(ctx sdk.Context, codeID uint64) ([]byte, error) {
...
+			return err
+		}
+		// redispatch all events, (type sdk.EventTypeMessage will be filtered out in the handler)
+		ctx.EventManager().EmitEvents(events)
+	}
+	return nil
+}
+
+func (k Keeper) dispatchMsgWithGasLimit(ctx sdk.Context, contractAddr sdk.AccAddress, ibcPort string, msg wasmvmtypes.CosmosMsg, gasLimit uint64) (events []sdk.Event, data [][]byte, err error) {
+	// FIXME: do I need to wrap the BlockGasMeter in the same way?
```
alpe, line 755: IMHO the block gas meter should not be wrapped here. If we consume more than available for the block then it would fail later anyway: https://docs.cosmos.network/master/basics/gas-fees.html#block-gas-meter

x/wasm/internal/keeper/keeper.go
```diff
@@ -684,15 +729,134 @@ func (k Keeper) GetByteCode(ctx sdk.Context, codeID uint64) ([]byte, error) {
...
+	limitedMeter := sdk.NewGasMeter(gasLimit)
+	subCtx := ctx.WithGasMeter(limitedMeter)
+
+	// catch out of gas panic and just charge the entire gas limit
+	defer func() {
+		if r := recover(); r != nil {
+			// if it's not an OutOfGas error, raise it again
+			// FIXME: do I need to do something special to get the proper stack trace here?
```
alpe, line 763: We have similar logic in the querier https://github.com/CosmWasm/wasmd/blob/master/x/wasm/internal/keeper/querier.go#L157-L175

alpe (commented):
Some quick feedback on the questions. No time for a proper review yet.

x/wasm/internal/keeper/keeper.go
```diff
@@ -684,15 +729,134 @@ func (k Keeper) GetByteCode(ctx sdk.Context, codeID uint64) ([]byte, error) {
...
+		}
+		// on failure, revert state from sandbox, and ignore events (just skip doing the above)
+
+		var result wasmvmtypes.SubcallResult
+		if err == nil {
+			result = wasmvmtypes.SubcallResult{
+				Ok: &wasmvmtypes.SubcallResponse{
+					Events: sdkEventsToWasmVmEvents(events),
+					// just take the first one for now if there are multiple sub-sdk messages
+					Data: data[0],
```
alpe, line 813: Req: `data` can be nil with `EncodeIBCMsg` for example 💣 
To be correct we should change the type to `[][]byte` for data. Another option would be to let the messenger "merge" the result data and return only a single element there. With both options I can see a communication issue for developers to use the data field proper without an understanding of the messenger implementation.

x/wasm/internal/keeper/submsg_test.go
```diff
@@ -0,0 +1,375 @@
...
+
+	sender := sub.Events[1]
+	assert.Equal(t, "message", sender.Type)
+	assert.Equal(t, wasmvmtypes.EventAttribute{
+		Key:   "sender",
+		Value: contractAddr.String(),
+	}, sender.Attributes[0])
+
+	// where does this come from?
+	module := sub.Events[2]
```
alpe, line 112: The event-filter is built into the [handler](https://github.com/CosmWasm/wasmd/blob/master/x/wasm/handler.go#L51) currently. It guess we need it in the keeper now.

x/wasm/internal/keeper/submsg_test.go
```diff
@@ -0,0 +1,375 @@
...
+	_, err = keeper.Execute(ctx, contractAddr, creator, reflectSendBz, nil)
+	require.NoError(t, err)
+
+	// fred got coins
+	checkAccount(t, ctx, accKeeper, bankKeeper, fred, sdk.NewCoins(sdk.NewInt64Coin("denom", 15000)))
+	// contract lost them
+	checkAccount(t, ctx, accKeeper, bankKeeper, contractAddr, sdk.NewCoins(sdk.NewInt64Coin("denom", 25000)))
+	checkAccount(t, ctx, accKeeper, bankKeeper, creator, creatorBalance)
+
+	// TODO: query the reflect state to ensure the result was stored
```
alpe, line 78: Any todo left?

x/wasm/internal/keeper/submsg_test.go
```diff
@@ -0,0 +1,375 @@
...
+		"out of gas caught with gas limit": {
+			id:          17,
+			msg:         infiniteLoop,
+			subMsgError: true,
+			gasLimit:    &subGasLimit,
+			// uses all the subGasLimit, plus the 50k or so for the main contract
+			resultAssertions: []assertion{assertGasUsed(subGasLimit+51000, subGasLimit+53000), assertErrorString("out of gas")},
+		},
+
+		"instantiate contract gets address in data and events": {
```
alpe, line 302: good test cases 🌷

alpe (changes requested):
Good 🎸 
Only the data payload may become an issue.

x/wasm/internal/keeper/keeper.go
ethanfrey: Decision was to charge the same fee here (which is 0 if pinned).
We can consider some discount later

x/wasm/internal/keeper/keeper.go
ethanfrey: Thanks, that was what I thought, but was unsure

x/wasm/internal/keeper/keeper.go
ethanfrey: Okay, that wants to catch all panics, I want to re-raise any other panic (just handle out of gas here).

x/wasm/internal/keeper/submsg_test.go
ethanfrey: We don't really need it there, I guess it would just be cleaner?

Let's make another issue for event filtering. Or just investigate more during https://github.com/CosmWasm/wasmd/issues/445

x/wasm/internal/keeper/keeper.go
ethanfrey: Nice find.

Tested and reproduced in https://github.com/CosmWasm/wasmd/pull/441/commits/c1d31ecca2b3a79429da9dc564c2984fd7e2f14d

Fixed in https://github.com/CosmWasm/wasmd/pull/441/commits/aba0596d720e6d2a818f528de0e6420422601491

ethanfrey:
Okay, I addressed all the comments and added a bug fix and test.

Rebased on master (to include pinned contract metering) and please feel free to merge if it looks better now.

alpe:
Awesome feature, Thanks 💐

x/wasm/internal/keeper/submsg_test.go
alpe: I will add a new issue for this.

PR #453, opened 2021-03-10
Allow subsecond block times

ethanfrey:
We need to adjust the genesis file in docker setup script to allow sub-second block times.

Solution from here: https://github.com/tendermint/tendermint/issues/6217#issuecomment-795589506 and manually tested on simapp in https://github.com/confio/ts-relayer/pull/100.

Let's get this in before we tag a wasmd release.


alpe (approved):
👍

PR #459, opened 2021-03-15
Handle nil in grpc querier

### LINKED ISSUE #416: GRPC querier impl should handle `nil` request objects
Add some `nil` check for a better error message
```
	if req == nil {
		return nil, status.Error(codes.InvalidArgument, "empty request")
	}
```

alpe:
Resolves #416


ethanfrey (approved):
:+1:

PR #460, opened 2021-03-16
Refactor wasm message handlers and query plugin

alpe:
🚧  WIP - not all tested

This is a refactoring within the wasmvm message handling components and queries.

Design goal was to 
* separation of message handling and encoding - IBC packets were hacked into encoders
* use of interfaces for decoupling
* drop of `nil` instantiation parameters for default plugins

New `Options` as extension points:
* `WithMessageHandler` - replace default message handler
* `WithQueryHandler` - replace default query handler
* `WithMessageEncoders` - replace default encoders
* `WithQueryPlugins` - replace default query plugins


x/wasm/internal/keeper/handler_plugin_encoders.go
```diff
@@ -0,0 +1,292 @@
...
+			SourcePort:       portID,
+			SourceChannel:    msg.Transfer.ChannelID,
+			Token:            amount,
+			Sender:           sender.String(),
+			Receiver:         msg.Transfer.ToAddress,
+			TimeoutHeight:    convertWasmIBCTimeoutHeightToCosmosHeight(msg.Transfer.TimeoutBlock),
+			TimeoutTimestamp: convertWasmIBCTimeoutTimestampToCosmosTimestamp(msg.Transfer.TimeoutTimestamp),
+		}
+		return []sdk.Msg{msg}, nil
+	default:
```
ethanfrey, line 252: Very cool, so the ones that can be easily translated into send messages are taken care of here.
And the one we cannot gets `ErrUnknownMsg`, so the IbcMsgHandler can take it on (and directly call the keeper)

x/wasm/internal/keeper/keeper.go
```diff
@@ -95,8 +102,6 @@ func NewKeeper(
 	homeDir string,
 	wasmConfig types.WasmConfig,
 	supportedFeatures string,
-	customEncoders *MessageEncoders,
```
ethanfrey, line 98: And I assume these are turned into `Option`s ?

x/wasm/internal/keeper/options.go
```diff
@@ -1,28 +1,73 @@
...
+		})
+		if !ok {
+			panic(fmt.Sprintf("Unsupported query handler type: %T", k.wasmVMQueryHandler))
+		}
+		k.wasmVMQueryHandler = q.Merge(x)
+	})
+}
+
+// WithMessageEncoders is an optional constructor parameter to pass custom message encoder to the default wasm message handler.
+func WithMessageEncoders(x *MessageEncoders) Option {
```
ethanfrey, line 50: Very cool to do this as options

x/wasm/internal/keeper/options.go
```diff
@@ -1,28 +1,73 @@
...
+	return optsFn(func(k *Keeper) {
+		q, ok := k.messenger.(*MessageHandlerChain)
+		if !ok {
+			panic(fmt.Sprintf("Unsupported message handler type: %T", k.messenger))
+		}
+		s, ok := q.handlers[0].(SDKMessageHandler)
```
ethanfrey, line 56: But these multiple forced casts to concrete types makes me a bit nervous.

I guess you do call `NewDefaultMessageHandler`, so this will only be false if someone tinkered with the private member before, which is only in test code most likely.

I doubt you could get this flexibility and still have compile time guarantees, but it would be good to have a note why this never should fail if you only use the public interfaces.

x/wasm/internal/keeper/options.go
```diff
@@ -1,28 +1,73 @@
...
+		if !ok {
+			panic(fmt.Sprintf("Unexpected message handler type: %T", q.handlers[0]))
+		}
+		e, ok := s.encoders.(interface {
```
ethanfrey, line 60: Maybe we just cast this to the concrete type we set up in `NewDefaultMessageHandler`. This is after 2x casting interfaces to concrete types, so if some funky modifications on the types happened already, you have no guarantees.

ethanfrey (approved):
Looks nice. And cool trick with the chaining and `ErrUnknownMsg`

x/wasm/internal/keeper/handler_plugin_encoders.go
alpe: ja, anything that can not be translated into a sdk msg type can be handled by a new handler impl. No need to mess with the encoders as before. 😄

x/wasm/internal/keeper/options.go
alpe: ja, I am not super happy with the casting but it would only fail when `WithMessageHandler` option was used before (or tests with different state in this package)

ethanfrey: You added a note above:

> This option should not be combined with Option `WithMessageEncoders`.

Looks good to me.

x/wasm/internal/keeper/options.go
```diff
@@ -1,28 +1,73 @@
...
 // given one.
 func WithWasmEngine(x types.WasmerEngine) Option {
 	return optsFn(func(k *Keeper) {
-		k.wasmer = x
+		k.wasmVM = x
 	})
 }
 
-// WithMessageHandler is an optional constructor parameter to set a custom message handler.
+// WithMessageHandler is an optional constructor parameter to set a custom handler for wasmVM messages.
```
ethanfrey, line 22: Not a blocker, but an idea for the future.

Now that we chain the handlers, most customizations (outside of tests) would likely be prepending or appending something to the default chain. This should be easy to do. 

Right now, making a new chain like the default is exposed as public constructors

```go
return NewMessageHandlerChain(
		NewSDKMessageHandler(router, encoders),
		NewIBCRawPacketHandler(channelKeeper, capabilityKeeper),
	)
```

But it does require the router, and various keepers to construct it. Those are not exposed on the keeper, we could only extend the default if we do that in `app.go` where we have the keepers, or some other code that wraps the constuctor of the keeper. Maybe that is the only place it would make sense to customize the handler, but just thinking....

If this makes sense to you, please open a future issue to allow appending/prepending to the default chain. If you think this is do-able with the current API, then I am fine.

Use case: if there is something I cannot express as an Encoder and I want to add support for that Custom message, I want to be able to extend this.

PR #463, opened 2021-03-18
Add telemetry

### LINKED ISSUE #387: Collect Contract Metrics
Placeholder for different kind of metrics we can collect for a contract.
Can be:
- Instantiation time
- Execution time
- Migration time
- IBC operations
- ...

alpe:
Resolves #387

### DIFF
--- a/contrib/prometheus/README.md
+++ b/contrib/prometheus/README.md
@@ -1,0 +1,47 @@
+# Setup
+Enable prometheus metrics in wasmd:
+
+* Edit `$HOME/config/app.toml`
+```toml
+[telemetry]
+
+# Enabled enables the application telemetry functionality. When enabled,
+# an in-memory sink is also enabled by default. Operators may also enabled
+# other sinks such as Prometheus.
+enabled =true
+# ...
+
+# PrometheusRetentionTime, when positive, enables a Prometheus metrics sink.
+prometheus-retention-time = 15
+```
+
+`retention-time` must be >0 (see prometheus scrape config)
+
+
+
+* Edit `$HOME/config/config.toml`
+```toml
+[instrumentation]
+
+# When true, Prometheus metrics are served under /metrics on
+# PrometheusListenAddr.
+# Check out the documentation for the list of available metrics.
+prometheus = true
+```
+# Local testing
+## Run Prometheus
+```sh
+# port 9090 is used by wasmd already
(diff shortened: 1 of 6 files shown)


contrib/prometheus/README.md
```diff
@@ -0,0 +1,47 @@
...
+```toml
+[instrumentation]
+
+# When true, Prometheus metrics are served under /metrics on
+# PrometheusListenAddr.
+# Check out the documentation for the list of available metrics.
+prometheus = true
+```
+# Local testing
+## Run Prometheus
```
ethanfrey, line 32: nice to document how to connect these parts, and even better with docker (no install)

x/wasm/internal/keeper/keeper.go
```diff
@@ -443,6 +447,7 @@ func (k Keeper) migrate(ctx sdk.Context, contractAddress sdk.AccAddress, caller
 // another native Go module directly. Thus, the keeper doesn't place any access controls on it, that is the
 // responsibility or the app developer (who passes the wasm.Keeper in app.go)
 func (k Keeper) Sudo(ctx sdk.Context, contractAddress sdk.AccAddress, msg []byte) (*sdk.Result, error) {
+	defer telemetry.MeasureSince(time.Now(), "wasm", "contract", "sudo")
```
ethanfrey, line 450: Nice, so we get a count and time used for all the keeper calls.

If we have recursive contract calls (A calls B calls C), we will get 3 prometheus events, right? (A + B + C, B + C, and C). Is it possible to annotate that these are nested somehow?

ethanfrey (approved):
Looks good. Open question about how the metrics work with recursive calls.
This probably needs manual testing to determine and can be improved in another PR. Adding this at all is already a great improvement.

x/wasm/internal/keeper/keeper.go
alpe: These are very broad metrics that measure an operation but not link them with the contract itself. With recursive calls we would get 3 events. I had chosen to not add contract labels as this would be a growing dataset. But I agree that some sort of tracing makes a lot of sense. We should look into https://opentelemetry.io/ and friends instead.

alpe: `pinned` could be a good category label in a new iteration

PR #465, opened 2021-03-18
Actions docker

faddat:
This is a github action that will build wasmd in amd64 + arm64

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2.1
 executors:
   golang:
     docker:
-      - image: circleci/golang:1.15
+      - image: circleci/golang:1.16
     working_directory: /go/src/github.com/cosmwasm/wasmd
 
 commands:
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -5,8 +5,7 @@ on:
   pull_request:
   push:
     branches:
-      - develop
-      - master
+      - '*'  
     tags:
       - "v[0-9]+.[0-9]+.[0-9]+" # Push events to matching v*, i.e. v1.0, v20.15.10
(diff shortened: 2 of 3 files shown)


.circleci/config.yml
```diff
@@ -3,7 +3,7 @@ version: 2.1
 executors:
   golang:
     docker:
-      - image: circleci/golang:1.15
+      - image: circleci/golang:1.16
```
alpe, line 6: Here and in go.mod: I think we can upgrade safely but without further tests I would like to keep this out of this PR

.github/workflows/docker.yml
```diff
@@ -0,0 +1,64 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+        with:
+          fetch-depth: 0
+
+      - name: Prepare
+        id: prep
+        run: |
+          DOCKER_IMAGE=faddat/wasmd
```
alpe, line 24: We publish our "official" wasmd image to https://hub.docker.com/r/cosmwasm/wasmd/tags?page=1&ordering=last_updated

alpe (commented):
Thank you very much for your contribution to the project 💐 

I was wondering if you have seen that we [build and upload](https://github.com/CosmWasm/wasmd/blob/master/.circleci/config.yml#L139-L144) a docker image (for testing) via Circle-CI. Artifacts are published to [docker-hub](https://hub.docker.com/r/cosmwasm/wasmd/tags?page=1&ordering=last_updated) automatically already.

I like github actions for new projects but can you elaborate why this is better than what we do with Circle-CI already? 
Maybe a github issue is a good start so that we discuss the idea and not the solution.  Open discussions are very much appreciated.

alpe:
Ping @faddat

alpe:
Closing this for now. Feel free to open the PR again when you have time. You can also find us on [discord](https://discord.com/invite/cPjEnPd) to follow up on the discussion. Thanks 🌻

PR #466, opened 2021-03-19
Restructure protobuf files

### LINKED ISSUE #342: Follow proto dir best practice (in cosmos eco)
I found more and more projects that store the protobuf files into
* `proto/` - for their own files
* `third_party/proto` - for the dependencies

This will make it easier to maintain multiple versions in the future but also works more nicely with `buf` linter.
On the downside these folders are not resolved by the Golang IDE automatically for dependencies.

alpe:
Resolves #342


proto/cosmwasm/wasm/v1beta1/proposal.proto
```diff
@@ -0,0 +1,121 @@
+syntax = "proto3";
+package cosmwasm.wasm.v1beta1;
```
ethanfrey, line 2: These are just moved around, right?

Seems like cleaner path naming

Makefile
```diff
@@ -201,21 +202,29 @@ format:
 ###############################################################################
 ###                                Protobuf                                 ###
 ###############################################################################
+PROTO_BUILDER_IMAGE=tendermintdev/sdk-proto-gen@sha256:372dce7be2f465123e26459973ca798fc489ff2c75aeecd814c0ca8ced24faca
```
ethanfrey, line 205: Cool, they provide these standard docker images now

ethanfrey (approved):
Looks good, I think we should stick with v1beta1 until we tag wasmd 1.0

proto/cosmwasm/wasm/v1beta1/proposal.proto
alpe: ja, files were just moved und updates on the path for imports.

Makefile
alpe: ja, I dropped our image for this.

scripts/protocgen.sh
```diff
@@ -1,19 +1,38 @@
...
+  $(find "${dir}" -maxdepth 1 -name '*.proto')
+
+done
+#
+## command to generate docs using protoc-gen-doc
+buf protoc \
+-I "proto" \
+-I "third_party/proto" \
+--doc_out=./docs/proto \
+--doc_opt=./docs/proto/protodoc-markdown.tmpl,proto-docs.md \
```
webmaster128, line 33: In this diff, `proto.md` was renamed to `proto-docs.md` but `proto.md` was not removed

PR #471, opened 2021-03-22
Introduce new interfaces for extendability

alpe:
Refactoring towards a more modular design.

* New interfaces in  `x/wasm/types/exported_keepers.go`
* `PermissionedKeeper` decorates the `keeper` with authorization policies and provides public methods to provision or interact with a contract
* Minor bugfix for registering proposal types with amino
* Emit multiple events for pin/unpin instead of list attribute


x/wasm/handler.go
```diff
@@ -12,7 +12,7 @@ import (
 )
 
 // NewHandler returns a handler for "bank" type messages.
-func NewHandler(k *Keeper) sdk.Handler {
+func NewHandler(k types.ContractOpsKeeper) sdk.Handler {
```
ethanfrey, line 15: Comment above this is incorrect I believe, can you update?

x/wasm/ibc.go
```diff
@@ -15,12 +15,12 @@ import (
...
 
 type IBCHandler struct {
-	keeper        Keeper
-	channelKeeper wasmTypes.ChannelKeeper
+	keeper        types.IBCContractKeeper
+	channelKeeper types.ChannelKeeper
 }
 
-func NewIBCHandler(keeper Keeper) IBCHandler {
-	return IBCHandler{keeper: keeper, channelKeeper: keeper.ChannelKeeper}
```
ethanfrey, line 23: Makes sense to add one arg to the constructor to not depend on the internal data structure

x/wasm/keeper/contract_keeper.go
```diff
@@ -0,0 +1,68 @@
...
+	unpinCode(ctx sdk.Context, codeID uint64) error
+	execute(ctx sdk.Context, contractAddress sdk.AccAddress, caller sdk.AccAddress, msg []byte, coins sdk.Coins) (*sdk.Result, error)
+}
+
+type PermissionedKeeper struct {
+	authZPolicy AuthorizationPolicy
+	nested      decoratedKeeper
+}
+
+func NewPermissionedKeeper(nested decoratedKeeper, authZPolicy AuthorizationPolicy) *PermissionedKeeper {
```
ethanfrey, line 26: Interesting way of combining these into interfaces. Allows two levels of interfaces for testing

x/wasm/types/exported_keepers.go
```diff
@@ -0,0 +1,89 @@
+package types
+
+import (
+	types2 "github.com/CosmWasm/wasmvm/types"
+	"github.com/cosmos/cosmos-sdk/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	capabilitytypes "github.com/cosmos/cosmos-sdk/x/capability/types"
+)
+
+// ViewKeeper provides read only operations
```
ethanfrey, line 10: Nice to make these external interfaces explicit

x/wasm/types/proposal.go
```diff
@@ -62,13 +62,13 @@ func init() { // register new content types with the sdk
...
 	govtypes.RegisterProposalType(string(ProposalTypePinCodes))
 	govtypes.RegisterProposalType(string(ProposalTypeUnpinCodes))
-	govtypes.RegisterProposalTypeCodec(StoreCodeProposal{}, "wasm/StoreCodeProposal")
-	govtypes.RegisterProposalTypeCodec(InstantiateContractProposal{}, "wasm/InstantiateContractProposal")
-	govtypes.RegisterProposalTypeCodec(MigrateContractProposal{}, "wasm/MigrateContractProposal")
-	govtypes.RegisterProposalTypeCodec(UpdateAdminProposal{}, "wasm/UpdateAdminProposal")
-	govtypes.RegisterProposalTypeCodec(ClearAdminProposal{}, "wasm/ClearAdminProposal")
-	govtypes.RegisterProposalTypeCodec(PinCodesProposal{}, "wasm/PinCodesProposal")
-	govtypes.RegisterProposalTypeCodec(UnpinCodesProposal{}, "wasm/UnpinCodesProposal")
+	govtypes.RegisterProposalTypeCodec(&StoreCodeProposal{}, "wasm/StoreCodeProposal")
```
ethanfrey, line 65: I wondered about this (pointer or struct) issue before when you created it.
My suggestion is to register both, as you cannot be fully sure what will come down the pipe and I have been hit by panics when only one type was registered before (*Msg but not Msg). Maybe that was only with amino, but in general, seems safer to just keep the old registration and add these to it (or is it illegal to have 2 types with the same type name?)

ethanfrey (approved):
Looks good. A few minor comments, but does make extensibility points much clearer.

x/wasm/types/proposal.go
alpe: You can not register both but it is best practice in the sdk to register the pointers not structs anymore. For example:
* [params](https://github.com/cosmos/cosmos-sdk/blob/v0.42.3/x/params/types/proposal/codec.go#L11)
* [upgrade](https://github.com/cosmos/cosmos-sdk/blob/v0.42.3/x/upgrade/types/codec.go#L12-L13)

PR #477, opened 2021-03-30
Add extension points to the CLI

alpe:
CLI was not on my list before for extension points but this may become handy for people working with CLI genesis messages:
```go
// GenesisReader reads genesis data. Extension point for custom genesis state readers.
type GenesisReader interface {
	ReadWasmGenesis(cmd *cobra.Command) (*GenesisData, error)
}

// GenesisMutator extension point to modify the wasm module genesis state
type GenesisMutator interface {
	// AlterModuleState loads the genesis from the default or set home dir,
	// unmarshalls the wasm module section into the object representation
	// calls the callback function to modify it
	// and marshals the modified state back into the genesis file
	AlterWasmModuleState(cmd *cobra.Command, callback func(state *types.GenesisState, appState map[string]json.RawMessage) error) error
}

```

### DIFF
--- a/x/wasm/client/cli/genesis_msg.go
+++ b/x/wasm/client/cli/genesis_msg.go
@@ -24,9 +24,23 @@ import (
 	tmtypes "github.com/tendermint/tendermint/types"
 )
 
+// GenesisReader reads genesis data. Extension point for custom genesis state readers.
+type GenesisReader interface {
+	ReadWasmGenesis(cmd *cobra.Command) (*GenesisData, error)
+}
+
+// GenesisMutator extension point to modify the wasm module genesis state
+type GenesisMutator interface {
+	// AlterModuleState loads the genesis from the default or set home dir,
+	// unmarshalls the wasm module section into the object representation
+	// calls the callback function to modify it
+	// and marshals the modified state back into the genesis file
+	AlterWasmModuleState(cmd *cobra.Command, callback func(state *types.GenesisState, appState map[string]json.RawMessage) error) error
+}
+
 // GenesisStoreCodeCmd cli command to add a `MsgStoreCode` to the wasm section of the genesis
 // that is executed on block 0.
-func GenesisStoreCodeCmd(defaultNodeHome string) *cobra.Command {
+func GenesisStoreCodeCmd(defaultNodeHome string, genesisMutator GenesisMutator) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "store [wasm file] --source [source] --builder [builder] --run-as [owner_address_or_key_name]\",",
 		Short: "Upload a wasm binary",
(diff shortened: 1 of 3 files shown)


x/wasm/client/cli/genesis_msg.go
```diff
@@ -377,44 +397,52 @@ func readGenesis(cmd *cobra.Command) (*genesisData, error) {
...
-	}, nil
+	return NewGenesisData(
+		genFile,
+		genDoc,
+		appState,
+		&wasmGenesisState,
+	), nil
+}
+
+type DefaultGenesisIO struct {
```
ethanfrey, line 408: Why is this here? Why not return `DefaultGenesisReader`?

I see that a project extending this would want to wrap it as you did below, but I don't see the need for it.

x/wasm/client/cli/genesis_msg.go
```diff
@@ -377,44 +397,52 @@ func readGenesis(cmd *cobra.Command) (*genesisData, error) {
...
+	GenesisReader
+}
+
+func NewDefaultGenesisIO() *DefaultGenesisIO {
+	return &DefaultGenesisIO{GenesisReader: DefaultGenesisReader{}}
```
ethanfrey, line 413: Proposal: `return &DefaultGenesisReader{}`

ethanfrey (approved):
Looks good. Confused by one level of indirection which may be unneeded. Otherwise great

x/wasm/client/cli/genesis_msg.go
alpe: `DefaultGenesisIO` implements `GenesisReader` and `GenesisMutator`. 
My motivation for the mutator was to give more flexibility for the genesis structure. In wasmd it is:
```json
{
"app_state":{
  "wasm":{
       "genesis":"data"
    }
}
```
With a different mutator implementation the structure can be different.

x/wasm/client/cli/genesis_msg.go
alpe: Will add some doc

x/wasm/client/cli/genesis_msg.go
ethanfrey: Okay, so someone could inherit the `GenesisReader` but provide a custom `GenesisMutator`.

I agree on the two interfaces, it was just a thought that `&DefaultGenesisIO{GenesisReader: DefaultGenesisReader{}}` just delegates to the DefaultGenesisReader, so why not just return that. Another project could do this wrapping you do here.

Anyway, this is a super minor point. Feel free to merge as is if there is a reason I don't understand.

alpe: Now I got it, sorry. Agree with you. :-)

PR #487, opened 2021-04-12
Upgrade to Wasmvm 0.14.0 beta3

### LINKED ISSUE #479: Update wasmvm to v0.14.0-beta2
As soon as https://github.com/CosmWasm/wasmvm/pull/208 is merged.

Besides updating contracts and gas and some error message, the only major change is adding `ReplyOn` to SubMsg. We need to check for Always (current behavior), Success (only reply if success, return errors directly) or Error (only reply if error, continue as normal on success).

ethanfrey:
Closes #479 

- [x] Update deps
- [x] Upgrade contracts
- [x] Add `ReplyOn` field to all SubMsg tests
- [x] Fix state tests to expect bech32 string, not base64 sdk.AccAddress
- [x] Update gas prices
- [x] Handle `ReplyOn` other than `Always`
- [x] Add support for BankMsg::Burn

@alpe This is working. Can you add support for BankMsg::Burn on this PR or merge this PR (if it looks good) and make that in another PR? I think you are much faster with your new handler design, I am not sure exactly where to add it properly.


x/wasm/keeper/submsg_test.go
```diff
@@ -450,3 +453,132 @@ func TestDispatchSubMsgEncodeToNoSdkMsg(t *testing.T) {
...
+	for name, tc := range cases {
+		id++
+		t.Run(name, func(t *testing.T) {
+			subMsg := wasmvmtypes.SubMsg{
+				ID:      id,
+				Msg:     tc.msg,
+				ReplyOn: wasmvmtypes.ReplySuccess,
+			}
+			if !tc.replyOnSuccess {
+				subMsg.ReplyOn = wasmvmtypes.ReplyError
```
alpe, line 546: nice to have: you can avoid the `if` and increase readability if you replace the `replyOnSuccess` bool by the `replyOn` value. But no need to change this.

x/wasm/keeper/submsg_test.go
```diff
@@ -450,3 +453,132 @@ func TestDispatchSubMsgEncodeToNoSdkMsg(t *testing.T) {
...
+				ToAddress: fred.String(),
+				Amount: []wasmvmtypes.Coin{{
+					Denom:  "no-such-token",
+					Amount: "777777",
+				}},
+			},
+		},
+	}
+
+	cases := map[string]struct {
```
alpe, line 501: nice test 👏

x/wasm/keeper/keeper.go
```diff
@@ -831,6 +831,15 @@ func (k Keeper) dispatchSubmessages(ctx sdk.Context, contractAddr sdk.AccAddress
 		}
 		// on failure, revert state from sandbox, and ignore events (just skip doing the above)
 
+		// we only callback if requested. Short-circuit here the two cases we don't want to
+		if msg.ReplyOn == wasmvmtypes.ReplySuccess && err != nil {
+			return err
+		}
+		if msg.ReplyOn == wasmvmtypes.ReplyError && err == nil {
+			return nil
+		}
```
alpe, line 840: 👍

go.sum
```diff
@@ -19,6 +19,10 @@ github.com/ChainSafe/go-schnorrkel v0.0.0-20200405005733-88cbf1b4c40d h1:nalkkPQ
 github.com/ChainSafe/go-schnorrkel v0.0.0-20200405005733-88cbf1b4c40d/go.mod h1:URdX5+vg25ts3aCh8H5IFZybJYKWhJHYMTnf+ULtoC4=
 github.com/CosmWasm/wasmvm v0.14.0-beta1 h1:ASqXB/2D8CEBmAI/uljRw6eEMbeKXPQtL/wZzKXZGGA=
 github.com/CosmWasm/wasmvm v0.14.0-beta1/go.mod h1:Id107qllDJyJjVQQsKMOy2YYF98sqPJ2t+jX1QES40A=
+github.com/CosmWasm/wasmvm v0.14.0-beta2 h1:8bLLNaxj796qUwu6UuQDbqpMV9zPpxG3lNZZbswNdOo=
+github.com/CosmWasm/wasmvm v0.14.0-beta2/go.mod h1:Id107qllDJyJjVQQsKMOy2YYF98sqPJ2t+jX1QES40A=
```
alpe, line 23: next `go mod tidy` run will drop the beta1 +2.

go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/CosmWasm/wasmd
 go 1.15
 
 require (
-	github.com/CosmWasm/wasmvm v0.14.0-beta1
+	github.com/CosmWasm/wasmvm v0.14.0-beta3
```
alpe, line 6: There is a reference in the [Dockerfile](https://github.com/CosmWasm/wasmd/blob/master/Dockerfile#L18) that we need to update. 🕵️

alpe (approved):
Very cool. 🏄  Please update the wasmvm version in the Dockerfile, too to be in sync.

alpe:
Ah, I will do a PR to this branch for `BankMsg::Burn` including the Dockerfile update

x/wasm/keeper/submsg_test.go
ethanfrey: I wanted to, but it requires me to use a private type in the struct... hmmmm

I made the type private to keep it a closed enum (there are only 3 public instances of it), rather than allow anyone to create new `ReplyOn` types.

go.mod
ethanfrey: Will do that and `go mod tidy`

PR #492, opened 2021-04-16
Extend ContractInfo for custom data

alpe:
Features:
* Adds new  `google.protobuf.Any extension` field to `Contractinfo` as extension point
* Replaces "VanillaJsonEncoder" in CLI with custom Proto-Json marshaler to handle `Any` types proper. This is a breaking change to the CLI json/ yaml output

This PR touched more files as the new `Any` field required some helper methods and tests for storage and queries.

To use the `Extension` you would need to register an implementation to the interface `types.ContractInfoExtension`
```go
interfaceRegistry.RegisterImplementations(
		(*types.ContractInfoExtension)(nil),
		&YourProtobufType{},
               )
```


cmd/wasmd/root.go
```diff
@@ -48,7 +49,7 @@ func NewRootCmd() (*cobra.Command, app.EncodingConfig) {
 	config.Seal()
 
 	initClientCtx := client.Context{}.
-		WithJSONMarshaler(encodingConfig.Marshaler).
+		WithJSONMarshaler(clientcodec.NewProtoCodec(encodingConfig.Marshaler, encodingConfig.InterfaceRegistry)).
```
ethanfrey, line 52: Ah, now I see the global encoder. The "omit empty" choice makes sense to me.

x/wasm/client/codec/marshaller.go
```diff
@@ -0,0 +1,76 @@
...
+	"fmt"
+	"github.com/cosmos/cosmos-sdk/codec"
+	"github.com/cosmos/cosmos-sdk/codec/types"
+	"github.com/gogo/protobuf/jsonpb"
+	"github.com/gogo/protobuf/proto"
+)
+
+var _ codec.Marshaler = (*ProtoCodec)(nil)
+
+// ProtoCodec that omits empty values
```
ethanfrey, line 14: Maybe comment that this is registered globally for all cli commands

x/wasm/keeper/genesis_test.go
```diff
@@ -642,6 +653,14 @@ func setupKeeper(t *testing.T) (*Keeper, sdk.Context, []sdk.StoreKey) {
 	}, false, log.NewNopLogger())
 
 	encodingConfig := MakeEncodingConfig(t)
+	// register an example extension. must be protobuf
+	encodingConfig.InterfaceRegistry.RegisterImplementations(
+		(*types.ContractInfoExtension)(nil),
+		&govtypes.Proposal{},
+	)
+	// also registering gov interfaces for nested Any type
+	govtypes.RegisterInterfaces(encodingConfig.InterfaceRegistry)
```
ethanfrey, line 662: Interesting example - and that this may be a general interface not just one type.

x/wasm/keeper/genesis_test.go
```diff
@@ -48,24 +48,35 @@ func TestGenesisExportImport(t *testing.T) {
...
+		f.Fuzz(&contractExtension)
+
 		creatorAddr, err := sdk.AccAddressFromBech32(codeInfo.Creator)
 		require.NoError(t, err)
 		codeID, err := contractKeeper.Create(srcCtx, creatorAddr, wasmCode, codeInfo.Source, codeInfo.Builder, &codeInfo.InstantiateConfig)
 		require.NoError(t, err)
 		if pinned {
 			contractKeeper.PinCode(srcCtx, codeID)
 		}
+		if contractExtension {
```
ethanfrey, line 72: cool that you fuzz this

x/wasm/keeper/querier_test.go
```diff
@@ -523,6 +524,77 @@ func TestQueryCodeList(t *testing.T) {
...
+			src:    &types.QueryContractInfoRequest{Address: contractAddr.String()},
+			stored: types.ContractInfoFixture(),
+			expRsp: &types.QueryContractInfoResponse{
+				Address: contractAddr.String(),
+				ContractInfo: types.ContractInfoFixture(func(info *types.ContractInfo) {
+					info.Created = nil // not returned on queries
+				}),
+			},
+		},
+		"with extension": {
```
ethanfrey, line 565: :+1:

x/wasm/types/types.go
```diff
@@ -83,6 +90,60 @@ func (c *ContractInfo) ValidateBasic() error {
 	if err := validateLabel(c.Label); err != nil {
 		return sdkerrors.Wrap(err, "label")
 	}
+	if c.Extension == nil {
+		return nil
+	}
+	e, ok := c.Extension.GetCachedValue().(validatable)
```
ethanfrey, line 96: This should definitely be documented (maybe below?)
not just how to register an Any type for the extension, but all variants of this type must implement `ValidateBasic()`

x/wasm/types/types.go
```diff
@@ -83,6 +90,60 @@ func (c *ContractInfo) ValidateBasic() error {
...
+	}
+	if err := e.ValidateBasic(); err != nil {
+		return sdkerrors.Wrap(err, "extension")
+	}
+	return nil
+}
+
+// SetExtension set new extension data
+func (c *ContractInfo) SetExtension(ext ContractInfoExtension) error {
+	if e, ok := ext.(validatable); ok {
```
ethanfrey, line 108: But here, you allow them to set one that doesn't implement `ValidateBasic` (treating that as "always valid").

Please unify this with the above function and document - either `ValidateBasic` is optional, and it is always valid if missing, or `ValidateBasic` is required and it is always invalid if missing. In any case, `ValidateBasic` is run on the extension when setting it and when validating ContractInfo

x/wasm/types/types.go
```diff
@@ -83,6 +90,60 @@ func (c *ContractInfo) ValidateBasic() error {
...
+		if err := e.ValidateBasic(); err != nil {
+			return err
+		}
+	}
+	if ext == nil {
```
ethanfrey, line 113: Shouldn't this be the first line of this function? (quick handling of nil, as with ValidateBasic above)

ethanfrey (approved):
Looks good.

I would just ask for some docs on usage, as well as clarifying how `ValidateBasic` works on the ContractExtension (it is handled slightly different in two places, this must be consistent and explicit - either implementation is fine with me, just make it clear)

x/wasm/keeper/genesis_test.go
alpe: yes, there can be multiple implementations registered.

x/wasm/types/types.go
alpe: uh, that was not intended. I had the interface implementation mandatory but made it optional later. Good to have you covering me. 😅

PR #493, opened 2021-04-16
Fix InitializePinnedCodes

alpe:
Fixes a 🐛  in `InitializePinnedCodes`


ethanfrey (approved):
Yup, that :bug: slipped in. Good fix.

Issue #496, opened 2021-04-23
Add cli alias


alpe:
My common typos :-)
* `list-code`, `codes` -> `list-codes`
* `list-contracts-by-code`, `contracts` -> `list-contract-by-code`

How about some short aliases for 
* `c` -> `contract`
* `lco` -> `list-codes`
* `lca` -> `list-contract-by-code`
* `ch` -> `contract-history`
* `cs` -> `contract-state`

ethanfrey:
I like more aliases. And optional short names.

PR #497, opened 2021-04-23
Improve list contracts by code query

### LINKED ISSUE #323: Correct order for migrated contracts
Currently we use the `ContractInfo.Created` marker as source for sorting. When a contract is migrate to a new code id the `list contract by code id` query should use the migration date instead.

With the `GetContractByCreatedSecondaryIndexKey` second index this can be implemented easily

alpe:
Resolves #323

Breaking change:
List contracts by code returns now the addresses without other data from CodeInfo
This was a gas intensive query before. Especially with the legacy rest endpoint


contrib/local/02-contracts.sh
```diff
@@ -25,7 +25,7 @@ wasmd tx wasm instantiate "$CODE_ID" "$INIT" --admin=$(wasmd keys show validator
   --from validator --amount="100ustake" --label "local0.1.0" \
   --gas 1000000 -y --chain-id=testing -b block | jq
 
-CONTRACT=$(wasmd query wasm list-contract-by-code "$CODE_ID" -o json | jq -r '.contract_infos[-1].address')
+CONTRACT=$(wasmd query wasm list-contract-by-code "$CODE_ID" -o json | jq -r '.contracts[-1]')
```
webmaster128, line 28: Shouldn't this be plural?

```suggestion
CONTRACT=$(wasmd query wasm list-contracts-by-code "$CODE_ID" -o json | jq -r '.contracts[-1]')
```

x/wasm/keeper/querier_test.go
```diff
@@ -283,13 +283,13 @@ func TestQueryContractListByCodeOrdering(t *testing.T) {
...
 	// create 10 contracts with real block/gas setup
-	for i := range [10]int{} {
+	for i := 0; i < 10; i++ {
 		// 3 tx per block, so we ensure both comparisons work
 		if i%3 == 0 {
 			ctx = setBlock(ctx, h)
 			h++
 		}
-		_, _, err = keepers.ContractKeeper.Instantiate(ctx, codeID, creator, nil, initMsgBz, fmt.Sprintf("contract %d", i), topUp)
+		_, _, err = keepers.ContractKeeper.Instantiate(ctx, codeID, creator, nil, initMsgBz, fmt.Sprintf("contractAddr %d", i), topUp)
```
webmaster128, line 292: Is this diff intended? Why is `i` a contract address?

x/wasm/keeper/keeper.go
```diff
@@ -507,8 +510,26 @@ func (k Keeper) reply(ctx sdk.Context, contractAddress sdk.AccAddress, reply was
...
+	store.Set(types.GetContractByCreatedSecondaryIndexKey(contractAddress, entry), []byte{})
+}
+
+// removeFromContractCodeSecondaryIndex removes element to the index for contracts-by-codeid queries
+func (k Keeper) removeFromContractCodeSecondaryIndex(ctx sdk.Context, contractAddress sdk.AccAddress, entry types.ContractCodeHistoryEntry) {
+	ctx.KVStore(k.storeKey).Delete(types.GetContractByCreatedSecondaryIndexKey(contractAddress, entry))
+}
+
+// IterateContractsByCode iterates over all contracts with given codeID ASC on code update time.
+func (k Keeper) IterateContractsByCode(ctx sdk.Context, codeID uint64, cb func(address sdk.AccAddress) bool) {
```
ethanfrey, line 525: Nice job optimizing this with secondary index

x/wasm/keeper/keeper.go
```diff
@@ -552,6 +573,19 @@ func (k Keeper) GetContractHistory(ctx sdk.Context, contractAddr sdk.AccAddress)
 	return r
 }
 
+// getLastContractHistoryEntry returns the last element from history. To be used internally only as it panics when none exists
+func (k Keeper) getLastContractHistoryEntry(ctx sdk.Context, contractAddr sdk.AccAddress) types.ContractCodeHistoryEntry {
```
ethanfrey, line 577: :+1:

x/wasm/types/keys_test.go
```diff
@@ -29,12 +29,12 @@ func TestGetContractByCodeIDSecondaryIndexPrefix(t *testing.T) {
 }
 
 func TestGetContractByCreatedSecondaryIndexKey(t *testing.T) {
-	c := &ContractInfo{
+	e := ContractCodeHistoryEntry{
 		CodeID:  1,
-		Created: &AbsoluteTxPosition{2 + 1<<(8*7), 3 + 1<<(8*7)},
```
ethanfrey, line 34: I would keep a different `Created` here just to prove it takes one and ignores the other.

x/wasm/keeper/keeper_test.go
```diff
@@ -1046,6 +1055,47 @@ func TestMigrateWithDispatchedMessage(t *testing.T) {
 	assert.Equal(t, deposit, balance)
 }
 
+func TestIterateContractsByCode(t *testing.T) {
```
ethanfrey, line 1058: I would extend this one (or make another one) that has a migrate - where a few contracts have different orders.

Migrating contract 1 to the same hackatom code (which we support) would change the order for example. 
Actually, better a second test - this is a great one to show creation usage - which must work for non-migrated contracts - we need another one for ordering of migrated contracts

ethanfrey (commented):
Good job with the code index, and looks like a good improvement. But I don't see any migration tests here. 

Can you add one that explicitly shows a different ordering based on migration vs creation?

contrib/local/02-contracts.sh
alpe: yes, and it is always 😕  
I have opened https://github.com/CosmWasm/wasmd/issues/496 already to address this via an alias so that old docs/tutorials still make sense

webmaster128: Ah, nice ticket and solution. Then `list-contracts-by-code` can become the primary command without braking.

x/wasm/keeper/querier_test.go
alpe: It is a label not an address. good 👁️

x/wasm/types/keys_test.go
alpe: I have switched from `ContractInfo` to `ContractCodeHistoryEntry` element. Any instantiation/migration results in a new `ContractCodeHistoryEntry` so that it covers both.

x/wasm/keeper/keeper_test.go
alpe: 👌 I will add a new one.

webmaster128 (approved):

x/wasm/keeper/keeper_test.go
```diff
@@ -1096,6 +1096,33 @@ func TestIterateContractsByCode(t *testing.T) {
...
+	example1 := InstantiateHackatomExampleContract(t, ctx, keepers)
+	ctx = ctx.WithBlockHeight(ctx.BlockHeight() + 1)
+	example2 := InstantiateIBCReflectContract(t, ctx, keepers)
+	ctx = ctx.WithBlockHeight(ctx.BlockHeight() + 1)
+	_, err := c.Migrate(ctx, example1.Contract, example1.CreatorAddr, example2.CodeID, []byte("{}"))
+	require.NoError(t, err)
+
+	// when
+	var gotAddr []sdk.AccAddress
+	k.IterateContractsByCode(ctx, example2.CodeID, func(address sdk.AccAddress) bool {
```
ethanfrey, line 1116: Very nice test case

ethanfrey (approved):
Looking good. Thanks for that test

Issue #486, opened 2021-04-12
Update to cosmwasm 0.14


ethanfrey:

webmaster128:
wasmvm v0.14.0-beta4 is released. Most important changes are the split of StakingMsg::Withdraw, the renamings of the all validators query and the new single validator query. You can also try the `GetMetrics(cache)` call if you are interested.

ethanfrey:
There will be some ibc-related changes coming in `beta5`. 
So far, we modified all the `timeout` fields on the messages (but left them as-is for the packets)

webmaster128:
This is done and released as 0.16.0

PR #499, opened 2021-04-26
Update to wasmvm 0.14.0-beta4

### LINKED ISSUE #399: Simplify staking reward query
See https://github.com/CosmWasm/wasmd/blob/v0.14.0/x/wasm/internal/keeper/query_plugins.go#L269-L292

https://github.com/cosmos/cosmos-sdk/issues/7466 was never closed but the relevant code was made public in stargate, so we can now use these lower level calls, rather than going through the public query interface.

ethanfrey:
Part of #486

This makes some changes to the staking/distribution messages and queries.


alpe (approved):
Very nice! 🌻

PR #502, opened 2021-04-27
Submessage reply can overwrite caller response

### LINKED ISSUE #495: Reply response on submessages can overwrite "caller" result
Currently we use the result from the call to `execute` and other methods as result of the operation. This makes it complicated to return the contract address for example that is created by a `submessage` instantiation call.

To resolve this issue the Data in [Response](https://github.com/CosmWasm/cosmwasm/blob/main/packages/std/src/results/response.rs#L86-L87) can optionally be overwritten by the `reply` Data.

if Data value is `nil` then it is ignored and the Data from the original `execute` call is used.
When empty or filled with binary data then it overwrites original response.
Attributes are *appended* to the original ones coming from `execute`.

More details:
https://github.com/CosmWasm/cosmwasm/pull/898/files#diff-dd0d039bdd2406cd229097a9fcb061900fd900c51e75931b59ac705ecfe421adR280

alpe:
Resolves #495

Although there a good tests for submessages that work with wasm contracts I found it very hard to to test the reply workflow without mocks. As a result I extracted:
* `MessageDispatcher` to handle the message/submessage flow 
* `DefaultWasmVMContractResponseHandler` to coordinate the dispatching order and result overwrite. This may become a nice extension point

Not related to the task but "refactored as needed" : rename of `WASMVMQueryHandler`

### DIFF
--- a/x/wasm/keeper/keeper.go
+++ b/x/wasm/keeper/keeper.go
@@ -14,7 +14,6 @@ import (
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 	authkeeper "github.com/cosmos/cosmos-sdk/x/auth/keeper"
 	paramtypes "github.com/cosmos/cosmos-sdk/x/params/types"
-	abci "github.com/tendermint/tendermint/abci/types"
 	"github.com/tendermint/tendermint/crypto"
 	"github.com/tendermint/tendermint/libs/log"
 	"path/filepath"
@@ -56,28 +55,36 @@ type WasmVMQueryHandler interface {
 	HandleQuery(ctx sdk.Context, caller sdk.AccAddress, request wasmvmtypes.QueryRequest) ([]byte, error)
 }
 
-// Messenger is an extension point for custom wasmVM message handling
-type Messenger interface {
-	// DispatchMsg encodes the wasmVM message and dispatches it.
-	DispatchMsg(ctx sdk.Context, contractAddr sdk.AccAddress, contractIBCPortID string, msg wasmvmtypes.CosmosMsg) (events []sdk.Event, data [][]byte, err error)
-}
-
 type CoinTransferrer interface {
 	// TransferCoins sends the coin amounts from the source to the destination with rules applied.
 	TransferCoins(ctx sdk.Context, fromAddr sdk.AccAddress, toAddr sdk.AccAddress, amt sdk.Coins) error
 }
 
+// WasmVMResponseHandler is an extension point to handles the response data returned by a contract call.
+type WasmVMResponseHandler interface {
+	// Handle processes the data returned by a contract invocation.
+	Handle(
+		ctx sdk.Context,
+		contractAddr sdk.AccAddress,
+		ibcPort string,
+		submessages []wasmvmtypes.SubMsg,
+		messages []wasmvmtypes.CosmosMsg,
+		origRspData []byte,
+	) ([]byte, error)
+}
+
 // Keeper will have a reference to Wasmer with it's own data directory.
 type Keeper struct {
-	storeKey           sdk.StoreKey
-	cdc                codec.Marshaler
-	accountKeeper      types.AccountKeeper
-	bank               CoinTransferrer
-	portKeeper         types.PortKeeper
-	capabilityKeeper   types.CapabilityKeeper
-	wasmVM             types.WasmerEngine
-	wasmVMQueryHandler WasmVMQueryHandler
-	messenger          Messenger
+	storeKey              sdk.StoreKey
+	cdc                   codec.Marshaler
+	accountKeeper         types.AccountKeeper
+	bank                  CoinTransferrer
+	portKeeper            types.PortKeeper
+	capabilityKeeper      types.CapabilityKeeper
+	wasmVM                types.WasmerEngine
+	wasmVMQueryHandler    WasmVMQueryHandler
+	wasmVMResponseHandler WasmVMResponseHandler
+	messenger             Messenger
 	// queryGasLimit is the max wasmvm gas that can be spent on executing a query with a contract
 	queryGasLimit uint64
 	paramSpace    paramtypes.Subspace
@@ -113,7 +120,7 @@ func NewKeeper(
 		paramSpace = paramSpace.WithKeyTable(types.ParamKeyTable())
 	}
 
-	keeper := Keeper{
+	keeper := &Keeper{
 		storeKey:         storeKey,
 		cdc:              cdc,
 		wasmVM:           wasmer,
@@ -125,11 +132,14 @@ func NewKeeper(
 		queryGasLimit:    wasmConfig.SmartQueryGasLimit,
 		paramSpace:       paramSpace,
 	}
-	keeper.wasmVMQueryHandler = DefaultQueryPlugins(bankKeeper, stakingKeeper, distKeeper, channelKeeper, queryRouter, &keeper)
+
+	keeper.wasmVMQueryHandler = DefaultQueryPlugins(bankKeeper, stakingKeeper, distKeeper, channelKeeper, queryRouter, keeper)
 	for _, o := range opts {
-		o.apply(&keeper)
+		o.apply(keeper)
 	}
-	return keeper
+	// not updateable, yet
+	keeper.wasmVMResponseHandler = NewDefaultWasmVMContractResponseHandler(NewMessageDispatcher(keeper.messenger, keeper))
+	return *keeper
 }
 
 func (k Keeper) getUploadAccessConfig(ctx sdk.Context) types.AccessConfig {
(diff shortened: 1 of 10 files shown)


x/wasm/keeper/keeper.go
```diff
@@ -113,7 +120,7 @@ func NewKeeper(
 		paramSpace = paramSpace.WithKeyTable(types.ParamKeyTable())
 	}
 
-	keeper := Keeper{
+	keeper := &Keeper{
```
ethanfrey, line 123: simpler

x/wasm/keeper/keeper_test.go
```diff
@@ -1309,3 +1309,90 @@ func TestInitializePinnedCodes(t *testing.T) {
...
+				m.DispatchMessagesFn = noopDMsgs
+			},
+			expErr:  false,
+			expData: []byte("mySubMsgData"),
+		},
+		"submessage overwrites result when empty": {
+			srcData: []byte("otherData"),
+			setup: func(m *wasmtesting.MockMsgDispatcher) {
+				m.DispatchSubmessagesFn = func(ctx sdk.Context, contractAddr sdk.AccAddress, ibcPort string, msgs []wasmvmtypes.SubMsg) ([]byte, error) {
+					return []byte(""), nil
```
ethanfrey, line 1339: We need to look closely at the FFI boundary and see if we have that distinction between `[]byte("")` and `nil` available

x/wasm/keeper/keeper_test.go
```diff
@@ -1309,3 +1309,90 @@ func TestInitializePinnedCodes(t *testing.T) {
 		assert.Equal(t, exp, capturedChecksums[i])
 	}
 }
+
+func TestNewDefaultWasmVMContractResponseHandler(t *testing.T) {
+	noopDMsgs := func(ctx sdk.Context, contractAddr sdk.AccAddress, ibcPort string, msgs []wasmvmtypes.CosmosMsg) error {
+		return nil
+	}
+
+	specs := map[string]struct {
```
ethanfrey, line 1318: Nice tests

x/wasm/keeper/msg_dispatcher.go
```diff
@@ -0,0 +1,174 @@
...
+	keeper    replyer
+}
+
+// NewMessageDispatcher constructor
+func NewMessageDispatcher(messenger Messenger, keeper replyer) *MessageDispatcher {
+	return &MessageDispatcher{messenger: messenger, keeper: keeper}
+}
+
+// DispatchMessages sends all messages.
+func (d MessageDispatcher) DispatchMessages(ctx sdk.Context, contractAddr sdk.AccAddress, ibcPort string, msgs []wasmvmtypes.CosmosMsg) error {
```
ethanfrey, line 34: Okay, this is just https://github.com/CosmWasm/wasmd/pull/502/files#diff-4e16010ab332946722b789b22962fb5a0814d529cd8de443b925d20ccd13705fL782-L792 without changes

x/wasm/keeper/msg_dispatcher.go
```diff
@@ -0,0 +1,174 @@
...
+			return err
+		}
+		// redispatch all events, (type sdk.EventTypeMessage will be filtered out in the handler)
+		ctx.EventManager().EmitEvents(events)
+	}
+	return nil
+}
+
+// dispatchMsgWithGasLimit sends a message with gas limit applied
+func (d MessageDispatcher) dispatchMsgWithGasLimit(ctx sdk.Context, contractAddr sdk.AccAddress, ibcPort string, msg wasmvmtypes.CosmosMsg, gasLimit uint64) (events []sdk.Event, data [][]byte, err error) {
```
ethanfrey, line 47: Copy of https://github.com/CosmWasm/wasmd/pull/502/files#diff-4e16010ab332946722b789b22962fb5a0814d529cd8de443b925d20ccd13705fL794-L818 without changes

x/wasm/keeper/msg_dispatcher.go
```diff
@@ -0,0 +1,174 @@
...
+	events, data, err = d.messenger.DispatchMsg(subCtx, contractAddr, ibcPort, msg)
+
+	// make sure we charge the parent what was spent
+	spent := subCtx.GasMeter().GasConsumed()
+	ctx.GasMeter().ConsumeGas(spent, "From limited Sub-Message")
+
+	return events, data, err
+}
+
+// DispatchSubmessages builds a sandbox to execute these messages and returns the execution result to the contract
```
ethanfrey, line 73: This comes from https://github.com/CosmWasm/wasmd/pull/502/files#diff-4e16010ab332946722b789b22962fb5a0814d529cd8de443b925d20ccd13705fL820-L890

with some changes

x/wasm/keeper/msg_dispatcher.go
```diff
@@ -0,0 +1,174 @@
...
+// that dispatched them, both on success as well as failure
+func (d MessageDispatcher) DispatchSubmessages(ctx sdk.Context, contractAddr sdk.AccAddress, ibcPort string, msgs []wasmvmtypes.SubMsg) ([]byte, error) {
+	var rsp []byte
+	for _, msg := range msgs {
+		switch msg.ReplyOn {
```
ethanfrey, line 78: Nice validation. Shall we add a `ReplyOn.Validate()` method to wasmvm? I guess that wouldn't return `sdkerrors` and we only use it one place.

x/wasm/keeper/msg_dispatcher.go
```diff
@@ -0,0 +1,174 @@
...
+			ctx.EventManager().EmitEvents(events)
+		}
+		// on failure, revert state from sandbox, and ignore events (just skip doing the above)
+
+		// we only callback if requested. Short-circuit here the two cases we don't want to
+		if msg.ReplyOn == wasmvmtypes.ReplySuccess && err != nil {
+			return nil, err
+		}
+		if msg.ReplyOn == wasmvmtypes.ReplyError && err == nil {
+			continue
```
ethanfrey, line 111: Good fix.

x/wasm/keeper/msg_dispatcher.go
```diff
@@ -0,0 +1,174 @@
...
+			Result: result,
+		}
+
+		// we can ignore any result returned as there is nothing to do with the data
+		// and the events are already in the ctx.EventManager()
+		rData, err := d.keeper.reply(ctx, contractAddr, reply)
+		switch {
+		case err != nil:
+			return nil, sdkerrors.Wrap(err, "reply")
+		case rData.Data != nil:
```
ethanfrey, line 147: And here we get the data from reply, that may override the original.

x/wasm/keeper/msg_dispatcher.go
```diff
@@ -0,0 +1,174 @@
...
+		}
+	}
+	return res
+}
+
+func sdkAttributesToWasmVmAttributes(attrs []abci.EventAttribute) []wasmvmtypes.EventAttribute {
+	res := make([]wasmvmtypes.EventAttribute, len(attrs))
+	for i, attr := range attrs {
+		res[i] = wasmvmtypes.EventAttribute{
+			Key:   string(attr.Key),
```
ethanfrey, line 169: I would love to know why they are `[]byte, []byte` in the sdk? (yet always used with `string`)
This mismatch may cause issues later

x/wasm/keeper/msg_dispatcher_test.go
```diff
@@ -0,0 +1,236 @@
...
+			msgs:    []wasmvmtypes.SubMsg{{ReplyOn: wasmvmtypes.ReplyError}},
+			replyer: noReplyCalled,
+			msgHandler: &wasmtesting.MockMessageHandler{
+				DispatchMsgFn: func(ctx sdk.Context, contractAddr sdk.AccAddress, contractIBCPortID string, msg wasmvmtypes.CosmosMsg) (events []sdk.Event, data [][]byte, err error) {
+					return nil, [][]byte{[]byte("myData")}, nil
+				},
+			},
+			expCommits: []bool{true},
+		},
+		"no reply on success without success": {
```
ethanfrey, line 37: Ah, you know this as `reply` is not set, so it would panic if it were called.

x/wasm/keeper/msg_dispatcher_test.go
```diff
@@ -0,0 +1,236 @@
...
+			},
+			msgHandler: &wasmtesting.MockMessageHandler{
+				DispatchMsgFn: func(ctx sdk.Context, contractAddr sdk.AccAddress, contractIBCPortID string, msg wasmvmtypes.CosmosMsg) (events []sdk.Event, data [][]byte, err error) {
+					return nil, nil, errors.New("my error")
+				},
+			},
+			expData:    []byte("myReplyData:2"),
+			expCommits: []bool{false, false},
+		},
+		"multiple msg - last reply with non nil": {
```
ethanfrey, line 169: Very nice and thorough list of cases

x/wasm/keeper/relay.go
```diff
@@ -145,11 +145,7 @@ func (k Keeper) OnRecvPacket(
 	// emit all events from this contract itself
 	events := types.ParseEvents(res.Attributes, contractAddr)
 	ctx.EventManager().EmitEvents(events)
-
-	if err := k.dispatchAll(ctx, contractAddr, contractInfo.IBCPortID, res.Submessages, res.Messages); err != nil {
-		return nil, err
-	}
-	return res.Acknowledgement, nil
+	return k.wasmVMResponseHandler.Handle(ctx, contractAddr, contractInfo.IBCPortID, res.Submessages, res.Messages, res.Acknowledgement)
```
ethanfrey, line 148: :100: amazing how you made this even simpler in this case (with the proper acknowledgement overwrite)

ethanfrey (approved):
Awesome work!

My one question is whether the `reply` interface that actually goes over FFI to a wasm contract properly distinguished between `nil` and `[]byte("")`. I will look more into that.

ethanfrey:
I checked the handling of `""` vs `null` in the Response.Data field. https://play.golang.org/p/EeiOLNtwRiR verified they are handled differently. Relevant wasmvm code:

We parse raw Rust-encoded response into a `ContractResult` type:

```go
    data, gasUsed, err := api.Reply(vm.cache, checksum, envBin, replyBin, &gasMeter, store, &goapi, &querier, gasLimit, vm.printDebug)
    if err != nil {
        return nil, gasUsed, err
    }

    var resp types.ContractResult
    err = json.Unmarshal(data, &resp)
```

This type defines `Data []byte` field, does that differentiate between `None` (json: `null`) and `Some(vec![])` (json: `""`)? The Go playgound above say yes.

```go
// ContractResult is the raw response from the instantiate/execute/migrate calls.
// This is mirrors Rust's ContractResult<Response>.
type ContractResult struct {
    Ok  *Response `json:"ok,omitempty"`
    Err string    `json:"error,omitempty"`
}

// Response defines the return value on a successful instantiate/execute/migrate.
// This is the counterpart of [Response](https://github.com/CosmWasm/cosmwasm/blob/v0.14.0-beta1/packages/std/src/results/response.rs#L73-L88)
type Response struct {
    // Submessages are like Messages, but they guarantee a reply to the calling contract
    // after their execution, and return both success and error rather than auto-failing on error
    Submessages []SubMsg `json:"submessages"`
    // Messages comes directly from the contract and is it's request for action
    Messages []CosmosMsg `json:"messages"`
    // base64-encoded bytes to return as ABCI.Data field
    Data []byte `json:"data"`
    // attributes for a log event to return over abci interface
    Attributes []EventAttribute `json:"attributes"`
}
```

x/wasm/keeper/keeper.go
```diff
@@ -1076,3 +948,39 @@ func (c BankCoinTransferrer) TransferCoins(ctx sdk.Context, fromAddr sdk.AccAddr
...
+func (h DefaultWasmVMContractResponseHandler) Handle(
+	ctx sdk.Context,
+	contractAddr sdk.AccAddress,
+	ibcPort string,
+	submessages []wasmvmtypes.SubMsg,
+	messages []wasmvmtypes.CosmosMsg,
+	origRspData []byte,
+) ([]byte, error) {
+	result := origRspData
+	switch rsp, err := h.md.DispatchSubmessages(ctx, contractAddr, ibcPort, submessages); {
```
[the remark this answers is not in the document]
ethanfrey: After researching `wasmvm`, I came to the conclusion this all works perfectly fine with the existing Go-Rust interface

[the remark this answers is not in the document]
alpe: ja, I realized later that it may not be easy to review this as it is displayed as new content. Apologies for this. Tests were not refactored though.

> Note: do we need to differentiate between nil and []byte{} here

If we consider empty `[]bye` a valid response from a contract then we need to be able to overwrite this by the reply. The `nil` result is considered an opt out to overwrite the response.

> After researching wasmvm, I came to the conclusion this all works perfectly fine with the existing Go-Rust interface

👏

x/wasm/keeper/msg_dispatcher.go
alpe: There would have been some opportunities for helper or validation methods but I got used to think of the types there as pure data object. Generally speaking it is faster to do it in wasmd than to work in 2 projects.

x/wasm/keeper/msg_dispatcher.go
alpe: No idea either. 🤷

alpe:
Thank you for this detailed review feedback. Especially for the time you took to look into the wasmvm response data de/serialization. Very happy that it makes sense 😄

x/wasm/keeper/keeper.go
```diff
@@ -125,11 +132,14 @@ func NewKeeper(
...
-	keeper.wasmVMQueryHandler = DefaultQueryPlugins(bankKeeper, stakingKeeper, distKeeper, channelKeeper, queryRouter, &keeper)
+
+	keeper.wasmVMQueryHandler = DefaultQueryPlugins(bankKeeper, stakingKeeper, distKeeper, channelKeeper, queryRouter, keeper)
 	for _, o := range opts {
-		o.apply(&keeper)
+		o.apply(keeper)
 	}
-	return keeper
+	// not updateable, yet
+	keeper.wasmVMResponseHandler = NewDefaultWasmVMContractResponseHandler(NewMessageDispatcher(keeper.messenger, keeper))
```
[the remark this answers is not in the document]
alpe: IMHO it makes perfect sense to have this as an extension point that can be set from outside so that people can implement their own response handling strategy. This can be very powerful but this would be incompatible with other `Options` we have already. I kept this out of this PR.

PR #503, opened 2021-04-27
Monitor WasmVM cache metrics

### LINKED ISSUE #500: Use the cache metrics from CosmWasm
See https://github.com/CosmWasm/wasmd/issues/486#issuecomment-820506511

> You can also try the GetMetrics(cache) call if you are interested.

alpe:
Resolves #500 

Example output: http://localhost:1317/metrics?format=prometheus
```
# HELP wasmvm_cache_elements_total Total number of elements in the cache
# TYPE wasmvm_cache_elements_total gauge
wasmvm_cache_elements_total{type="memory"} 1
wasmvm_cache_elements_total{type="pinned"} 0
# HELP wasmvm_cache_hits_total Total number of cache hits
# TYPE wasmvm_cache_hits_total counter
wasmvm_cache_hits_total{type="fs"} 1
wasmvm_cache_hits_total{type="memory"} 2
wasmvm_cache_hits_total{type="pinned"} 0
# HELP wasmvm_cache_misses_total Total number of cache misses
# TYPE wasmvm_cache_misses_total counter
wasmvm_cache_misses_total 0
# HELP wasmvm_cache_size_bytes Total number of elements in the cache
# TYPE wasmvm_cache_size_bytes gauge
wasmvm_cache_size_bytes{type="memory"} 3.428923e+06
wasmvm_cache_size_bytes{type="pinned"} 0
```


x/wasm/keeper/metrics.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	wasmvmtypes "github.com/CosmWasm/wasmvm/types"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const (
+	labelPinned = "pinned"
+	labelMemory = "memory"
+	labelFs     = "memory"
```
webmaster128, line 11: I don't know what those labels are used for, but should this be `"fs"`?

x/wasm/keeper/metrics.go
```diff
@@ -0,0 +1,65 @@
...
+func (p *WasmVMMetricsCollector) Collect(c chan<- prometheus.Metric) {
+	m := p.source.GetMetrics()
+	c <- prometheus.MustNewConstMetric(p.CacheHitsDescr, prometheus.CounterValue, float64(m.HitsPinnedMemoryCache), labelPinned)
+	c <- prometheus.MustNewConstMetric(p.CacheHitsDescr, prometheus.CounterValue, float64(m.HitsMemoryCache), labelMemory)
+	c <- prometheus.MustNewConstMetric(p.CacheHitsDescr, prometheus.CounterValue, float64(m.HitsFsCache), labelFs)
+	c <- prometheus.MustNewConstMetric(p.CacheMissesDescr, prometheus.CounterValue, float64(m.Misses))
+	c <- prometheus.MustNewConstMetric(p.CacheElementsDescr, prometheus.GaugeValue, float64(m.ElementsPinnedMemoryCache), labelPinned)
+	c <- prometheus.MustNewConstMetric(p.CacheElementsDescr, prometheus.GaugeValue, float64(m.ElementsMemoryCache), labelMemory)
+	c <- prometheus.MustNewConstMetric(p.CacheSizeDescr, prometheus.GaugeValue, float64(m.SizeMemoryCache), labelMemory)
+	c <- prometheus.MustNewConstMetric(p.CacheSizeDescr, prometheus.GaugeValue, float64(m.SizePinnedMemoryCache), labelPinned)
```
webmaster128, line 64: FYI: the number of elements and the size of elements in the file system cache cannot easily be obtained. We had to either scan the whole directory of potentially thousands of files or track the values when files are added or removed. Such a tracking would need to be on disk such that the values are not cleared when the node is restarted.

x/wasm/keeper/metrics.go
alpe: 🤦  definitively not "memory" thanks for finding this!

x/wasm/keeper/metrics.go
alpe: Thanks for the info. I was wondering about fs myself and will add this as a comment in case this comes up again.

ethanfrey:
Very cool. Looking at this line:

> wasmvm_cache_size_bytes{type="memory"} 3.428923e+06

@webmaster128 this is 3.5MB, seems to make sense given cache uses 10-20x the size of wasm. How is this calculated? Via heuristics based on wasm size, or can you actually measure the objects?

contrib/prometheus/README.md
```diff
@@ -28,6 +27,13 @@ prometheus-retention-time = 15
 # Check out the documentation for the list of available metrics.
 prometheus = true
 ```
+
+Test manually at:
+`http://localhost:1317/metrics?format=prometheus`
```
ethanfrey, line 32: Wait, we need the lcd server to display metrics?
Sounds like a left-over design. We should revisit this at some point, but fine for this PR.

x/wasm/keeper/options.go
```diff
@@ -74,3 +75,9 @@ func WithCoinTransferrer(x CoinTransferrer) Option {
 		k.bank = x
 	})
 }
+
+func WithVMCacheMetrics(r prometheus.Registerer) Option {
```
ethanfrey, line 79: :+1:

cmd/wasmd/root.go
```diff
@@ -199,13 +201,17 @@ func newApp(logger log.Logger, db dbm.DB, traceStore io.Writer, appOpts serverty
 	if err != nil {
 		panic(err)
 	}
-	var emptyWasmOpts []wasm.Option
+	var wasmOpts []wasm.Option
```
ethanfrey, line 204: Nice way to enable them

ethanfrey (approved):
Good stuff. Let's merge it in and add some real monitoring to the next testnet

contrib/prometheus/README.md
alpe: ja, there is another endpoint at port `26660` in `config.toml` but it took me a bit to figure out why `1317` was not returning any of our data (because of the missing format param)

webmaster128:
> @webmaster128 this is 3.5MB, seems to make sense given cache uses 10-20x the size of wasm. How is this calculated? Via heuristics based on wasm size, or can you actually measure the objects?

When adding the value to the cache, it is [the size of the artifact of the fire system](https://github.com/CosmWasm/cosmwasm/blob/v0.14.0-beta5/packages/vm/src/modules/file_system_cache.rs#L110-L114). Same when [loading from disk into memory](https://github.com/CosmWasm/cosmwasm/blob/v0.14.0-beta5/packages/vm/src/modules/file_system_cache.rs#L76-L82).

Experiments showed that this is a great estimate of the Module in memory because the artifact data is the majority of the size and the same on disk and in memory.

PR #508, opened 2021-04-29
Upgrade to wasmvm v0.14.0-rc1

alpe:
Upgrade to wasmvm v0.14.0-rc1

Deactivated `ibc_reflect_test`


x/wasm/types/types.go
```diff
@@ -260,17 +260,15 @@ func NewEnv(ctx sdk.Context, contractAddr sdk.AccAddress) wasmvmtypes.Env {
 	if ctx.BlockHeight() < 0 {
 		panic("Block height must never be negative")
 	}
-	sec := ctx.BlockTime().Unix()
-	if sec < 0 {
-		panic("Block (unix) time must never be negative ")
+	nano := ctx.BlockTime().UnixNano()
+	if nano < 1 {
```
webmaster128, line 264: A 0 time is well supported in this case. It is serialized as `"0"` because of `string` in
```
Time    uint64 `json:"time,string"`
```
. But this way works as well.

webmaster128 (approved):
Nice

ethanfrey:
Fixing contracts here: https://github.com/CosmWasm/cosmwasm/pull/910

I will compile and try to fix the test now

x/wasm/types/types.go
alpe: IMHO there is no realistic scenario with timestamp 0 and we should fail fast. While genesis block has height 0 it does contain a >0 timestamp.

alpe:
@ethanfrey  thanks a lot for the contract fix 🚀

PR #510, opened 2021-04-30
Release v0.16.0 changelog

alpe:
Update changelog


ethanfrey (approved):
Looks good. But it would be nice to add a comment on public API breaking changes. Notably the changing return value of `GetContractsByCode`. Maybe there are others.

I hit this in particular when updating CosmJS to the latest version. Others might care as well.

alpe:
> Looks good. But it would be nice to add a comment on public API breaking changes. Notably the changing return value of `GetContractsByCode`. Maybe there are others.

👍  I had added a label already to mark them

PR #521, opened 2021-05-26
Update changelog for v0.17.0

alpe:
Updates Changelog.md


ethanfrey (approved):
Good start

ethanfrey:
If you are going to tag v0.17.0 now is there anything else you want to put there?
Or is this just to unblock you with the json.RawMessage issue and v0.17.0 rather than 0.16.x as it is technically a breaking API change?

alpe:
I have moved the tasks for v0.17.0 to v0.18.0. This release is for the sdk upgrade and json cast only. Although there are no API changes on our side, it can still have impact when people unmarshal from proto to json. Therefore a v0.17.0 felt more appropriate than v0.16.1.

ethanfrey (approved):
looks good.

0.17 here we come!

PR #522, opened 2021-05-27
Localnet scripts

maurolacy:
Add some scripts to initialize and run a localnet, and store, instantiate and query a smart contract.

### DIFF
--- a/contrib/localnet/README.md
+++ b/contrib/localnet/README.md
@@ -1,0 +1,18 @@
+## Localnet configuration scripts
+
+For wasmd 0.14.x, CosmWasm 0.13.x.
+
+ - `localnet_*.sh`: Scripts to initialize and launch `wasmd` in a localnet configuration.
+
+In particular, `localnet_vars.sh` can be used to customize and setup the localnet.
+
+It can be invoked from the command line if needed (`source ./localnet_vars.sh`)
+in order to have the variables set in the current environment.
+In any case, it is being invoked by the other scripts to setup their environment.
+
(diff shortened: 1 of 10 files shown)


alpe:
Thanks for the contribution! 🌷  
Are the scripts intended to support testing #523 or should they go into master?
We have https://github.com/CosmWasm/wasmd/tree/master/contrib/local that I keep updated.

maurolacy:
> Are the scripts intended to support testing #523 or should they go into master?

I created them to support the experiments / tests related to #523. But they can be useful for others using CosmWasm 0.13.x. Not sure where to merge it, tough. Maybe the new "0.14" branch is a better place than master. Pity that we already released 0.14.2; but anyway.

> We have https://github.com/CosmWasm/wasmd/tree/master/contrib/local that I keep updated.

I've seen it. Not sure if it works with CosmWasm 0.13.x, tough. Maybe we can integrate two into one? Maybe we can do that, and create two versions / folders, for cw 0.13 and cw 0.14? I don't know.

maurolacy:
> Thanks for the info. If they were useful for you to test the patch then we should conserve them. I would rebase and merge them untested to the `0.14` branch.

OK with me.

> On master I would prefer a single set of local test scripts that are curated. Can update `contrib/local` or open a new issue if there is anything important missing from your scripts?

I'll try to use them for the original use case (I built my set because I didn't knew they existed in the first place) and report back.

maurolacy:
Rebased against 0.14.

orkunkl (approved):

Issue #410, opened 2021-02-16
Update Contract Address Derivation


ethanfrey:
Back in 2019, I devised a [custom scheme to provide a unique address](https://github.com/CosmWasm/wasmd/blob/e6735eb734c14f537809610480b15eeaace28219/x/wasm/internal/keeper/keeper.go#L625-L638) to each contract.

Basically: `sha256('c' | code_id | contract_counter)[:20]`

With https://github.com/cosmos/cosmos-sdk/pull/8415 they have now defined a [standard way to derive module accounts](https://github.com/cosmos/cosmos-sdk/blob/master/docs/architecture/adr-028-public-key-addresses.md#module-account-addresses). It looks something like:

```go
func Module(moduleName string, key []byte) []byte{
	return Hash("module", []byte(moduleName) + 0 + key)
}
```

Since the 0.16 release will be breaking with many CosmWasm changes, let's take advantage of that to future-proof our derivation to be more consistent with the sdk.

I would propose something like: `addr := Hash("cosmwasm", code_id | countract_counter)`, which is pretty similar to the old code, but the standard derivation.

We can also revisit how we create these unique key (code | contract) for each contract, although it does seem to work and our genesis import/export support this well.

ethanfrey:
Maybe the new standard uses 32 byte addresses that are not fully supported until cosmos sdk 0.42. We could do this in v0.17?

ethanfrey:
This was actually implemented in https://github.com/cosmos/cosmos-sdk/pull/9088

We are unlikely to import the upstream code in wasmd 1.0 (which will most likely be v0.42.x based), but we can ensure our derivation is the same as the upstream definition, so we can just swap out implementations later. 

Basically, we could copy this code into `x/wasm` for now: https://github.com/cosmos/cosmos-sdk/pull/9088/files#diff-6cfded83a08be06034f746e83ffa6bbd6c35430cd8ac05015e6bd942c7da1030L21-R70

And use something like `Compose("wasm", []Addressable{ codeId, contractCounter})`

robert-zaremba:
Check this PR: I backported that feature to Regen (which was released with 0.42 as well), and the usage:
https://github.com/regen-network/regen-ledger/pull/313/files#diff-4c9db4a1c1af45a68754c38528ea8894dccb4c75c8fc4a5203ba69cb0f57a80bR36

ethanfrey:
Thank you for the link (and the tip it was live)

[...]

PR #527, opened 2021-06-01
Make contract address builder public

alpe:
This can prevent some copy&paste as the contract address algorithm seems useful in other places as well. For example integration tests

Also includes:
* go mod tidy
* Use interface in keeper not struct type


ethanfrey (approved):
Nice one. I especially like placing all address derivation logic in one place, as we will soon modify it. https://github.com/CosmWasm/wasmd/issues/410

And being able to do that once and not have 3 different implementations make that a lot easier

PR #533, opened 2021-06-11
Add cost and api cost options

### LINKED ISSUE #525: Extendable gas costs
Gas costs are currently constants in the Code that can not be changed without forking. For example https://github.com/CosmWasm/wasmd/blob/master/x/wasm/keeper/keeper.go#L41

We should not enforce the same costs on every chain that is using this module. Some options are
- Params (create extra gas costs)
- Overwrite via compile time flag on a `var`
- Abstract provider interface that returns the value

alpe:
Resolves #525 

I have started with public vars but stopped as this would open the door for concurrent access from outside the module. As a library we should protect against that. A simple but safe alternative are constructor `Options`. It was not in my original lists of proposals though.

### DIFF
--- a/x/wasm/keeper/options.go
+++ b/x/wasm/keeper/options.go
@@ -81,3 +81,26 @@ func WithVMCacheMetrics(r prometheus.Registerer) Option {
 		NewWasmVMMetricsCollector(k.wasmVM).Register(r)
 	})
 }
+
+// WithCosts sets custom gas costs and multiplier.
+// See DefaultCompileCost, DefaultInstanceCost, DefaultGasMultiplier
+// Uses WithApiCosts with defaults and given multiplier.
+func WithCosts(compile, instance, multiplier uint64) Option {
+	return optsFn(func(k *Keeper) {
+		k.compileCost = compile
+		k.instanceCost = instance
+		k.gasMultiplier = multiplier
+		WithApiCosts(
+			DefaultGasCostHumanAddress*multiplier,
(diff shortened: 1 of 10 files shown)


x/wasm/keeper/options.go
```diff
@@ -81,3 +81,26 @@ func WithVMCacheMetrics(r prometheus.Registerer) Option {
 		NewWasmVMMetricsCollector(k.wasmVM).Register(r)
 	})
 }
+
+// WithCosts sets custom gas costs and multiplier.
+// See DefaultCompileCost, DefaultInstanceCost, DefaultGasMultiplier
+// Uses WithApiCosts with defaults and given multiplier.
+func WithCosts(compile, instance, multiplier uint64) Option {
```
ethanfrey, line 88: I like this as cleaner than global vars.

However, just want to point out, these options could be applied manually after construction, so don't prevent malicious usage of them (but make that more obvious)

`WithCosts(1, 2, 3)(keeper)` can be called anywhere.

ethanfrey (approved):
Looks like a good approach to allow this to be configured in `app.go` without adding runtime overhead (db reads for params approach)

x/wasm/keeper/options.go
alpe: The [`apply` method](https://github.com/CosmWasm/wasmd/blob/master/x/wasm/keeper/keeper.go#L44) is package private. We are safe with this.

PR #534, opened 2021-06-11
Cleanup keeper result types

### LINKED ISSUE #428: Replace sdk.Result in keeper.go
The keeper should not return any interface specific objects. Instead the `msg_server` or `handler` should be responsible to map into the proper interface types.

alpe:
Resolves #428

### DIFF
--- a/x/wasm/keeper/msg_dispatcher.go
+++ b/x/wasm/keeper/msg_dispatcher.go
@@ -144,7 +144,7 @@ func (d MessageDispatcher) DispatchSubmessages(ctx sdk.Context, contractAddr sdk
 		switch {
 		case err != nil:
 			return nil, sdkerrors.Wrap(err, "reply")
-		case rspData != nil:
+		case len(rspData) != 0:
 			rsp = rspData
 		}
 	}
--- a/x/wasm/keeper/msg_dispatcher.go
+++ b/x/wasm/keeper/msg_dispatcher.go
@@ -16,7 +16,7 @@ type Messenger interface {
 
 // replyer is a subset of keeper that can handle replies to submessages
 type replyer interface {
-	reply(ctx sdk.Context, contractAddress sdk.AccAddress, reply wasmvmtypes.Reply) (*sdk.Result, error)
+	reply(ctx sdk.Context, contractAddress sdk.AccAddress, reply wasmvmtypes.Reply) ([]byte, error)
 }
 
 // MessageDispatcher coordinates message sending and submessage reply/ state commits
@@ -140,12 +140,12 @@ func (d MessageDispatcher) DispatchSubmessages(ctx sdk.Context, contractAddr sdk
 
 		// we can ignore any result returned as there is nothing to do with the data
 		// and the events are already in the ctx.EventManager()
-		rData, err := d.keeper.reply(ctx, contractAddr, reply)
+		rspData, err := d.keeper.reply(ctx, contractAddr, reply)
 		switch {
 		case err != nil:
 			return nil, sdkerrors.Wrap(err, "reply")
-		case rData.Data != nil:
-			rsp = rData.Data
+		case rspData != nil:
+			rsp = rspData
 		}
 	}
 	return rsp, nil
--- a/x/wasm/keeper/msg_dispatcher_test.go
+++ b/x/wasm/keeper/msg_dispatcher_test.go
@@ -50,8 +50,8 @@ func TestDispatchSubmessages(t *testing.T) {
 				ReplyOn: wasmvmtypes.ReplySuccess,
 			}},
 			replyer: &mockReplyer{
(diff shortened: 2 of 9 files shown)


x/wasm/keeper/msg_dispatcher_test.go
```diff
@@ -50,8 +50,8 @@ func TestDispatchSubmessages(t *testing.T) {
 				ReplyOn: wasmvmtypes.ReplySuccess,
 			}},
 			replyer: &mockReplyer{
-				replyFn: func(ctx sdk.Context, contractAddress sdk.AccAddress, reply wasmvmtypes.Reply) (*sdk.Result, error) {
-					return &sdk.Result{Data: []byte("myReplyData")}, nil
+				replyFn: func(ctx sdk.Context, contractAddress sdk.AccAddress, reply wasmvmtypes.Reply) ([]byte, error) {
```
ethanfrey, line 53: Simpler for sure

x/wasm/keeper/msg_dispatcher.go
```diff
@@ -140,12 +140,12 @@ func (d MessageDispatcher) DispatchSubmessages(ctx sdk.Context, contractAddr sdk
...
 		// we can ignore any result returned as there is nothing to do with the data
 		// and the events are already in the ctx.EventManager()
-		rData, err := d.keeper.reply(ctx, contractAddr, reply)
+		rspData, err := d.keeper.reply(ctx, contractAddr, reply)
 		switch {
 		case err != nil:
 			return nil, sdkerrors.Wrap(err, "reply")
-		case rData.Data != nil:
-			rsp = rData.Data
+		case rspData != nil:
```
ethanfrey, line 147: Hmmm... this may be subtly different. We need to double-check between `[]byte{}` and `nil`.
I would prefer `case len(rspData) > 0:`

x/wasm/keeper/msg_dispatcher_test.go
```diff
@@ -184,6 +184,24 @@ func TestDispatchSubmessages(t *testing.T) {
...
+					}
+					return []byte("myReplyData:1"), nil
+				},
+			},
+			msgHandler: &wasmtesting.MockMessageHandler{
+				DispatchMsgFn: func(ctx sdk.Context, contractAddr sdk.AccAddress, contractIBCPortID string, msg wasmvmtypes.CosmosMsg) (events []sdk.Event, data [][]byte, err error) {
+					return nil, nil, errors.New("my error")
+				},
+			},
+			expData:    []byte{},
```
ethanfrey, line 202: I am not sure this is desired.
I see the power of explicitly clearing any data.
But when do we want that?

I have seen way too many mix-ups between `[]byte{}` and `nil` in Go, especially when it gets serialized and later deserialized (they both go to `nil` I believe). I would rather just say "no bytes means don't change".

Is there an issue to change this functionality? I would rather handle the two the same in this PR and we can open up a new Issue to discuss this one case (and not block this PR on it).

x/wasm/keeper/proposal_handler.go
```diff
@@ -123,14 +123,6 @@ func handleMigrateProposal(ctx sdk.Context, k types.ContractOpsKeeper, p types.M
 		sdk.NewAttribute(types.AttributeKeyContract, p.Contract),
 	)
 	ctx.EventManager().EmitEvent(ourEvent)
-
-	for _, e := range res.Events {
```
ethanfrey, line 127: Haha, we handle these events, but they are never emitted :facepalm:

ethanfrey (approved):
Looks good with one caveat about handling `[]byte{}` and `nil` differently

x/wasm/keeper/msg_dispatcher_test.go
alpe: I was puzzled by nil/ empty as well and then re-read the https://github.com/CosmWasm/cosmwasm/blob/main/SEMANTICS.md#handling-the-reply doc. This point is not mentioned explicit but I was interpreting Some/None as the differentiator. `Some(EmptyString)` would overwrite the result.
Current implementation is more complete but I don't have a strong opinion. I will address your concern

ethanfrey: I am not so confident that `None` / `Some(vec![])` / `Some(vec![1, 2, 3])` will be converted to
`nil` / `[]byte{}` / `[]byte{1, 2, 3}`. 

This may be the case, but having a JSON encoder/decoder between languages makes me skeptical.
Maybe @webmaster128 could weigh in if this is checked full stack?
Otherwise, I would go for the safe route and add a FIXME to verify if this gets handles properly between wasm and go

alpe: 👌  safe route is preferred

PR #535, opened 2021-06-11
Store contract response in event

### LINKED ISSUE #431: Contract response data in events
Follow up from https://github.com/CosmWasm/wasmd/issues/385#issuecomment-789658573

Required for gov types:
* Instantiate 
* Migrate

alpe:
Resolves #431

### DIFF
--- a/x/wasm/keeper/proposal_handler.go
+++ b/x/wasm/keeper/proposal_handler.go
@@ -95,7 +95,7 @@ func handleInstantiateProposal(ctx sdk.Context, k types.ContractOpsKeeper, p typ
 		sdk.NewAttribute(sdk.AttributeKeyModule, types.ModuleName),
 		sdk.NewAttribute(types.AttributeKeyCodeID, fmt.Sprintf("%d", p.CodeID)),
 		sdk.NewAttribute(types.AttributeKeyContract, contractAddr.String()),
-		sdk.NewAttribute(types.AttributeResultData, hex.EncodeToString(data)),
+		sdk.NewAttribute(types.AttributeResultDataHex, hex.EncodeToString(data)),
 	)
 	ctx.EventManager().EmitEvent(ourEvent)
 	return nil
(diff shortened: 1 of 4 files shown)


x/wasm/keeper/proposal_handler.go
```diff
@@ -112,15 +114,17 @@ func handleMigrateProposal(ctx sdk.Context, k types.ContractOpsKeeper, p types.M
...
-	_, err = k.Migrate(ctx, contractAddr, runAsAddr, p.CodeID, p.MigrateMsg)
+	data, err := k.Migrate(ctx, contractAddr, runAsAddr, p.CodeID, p.MigrateMsg)
 	if err != nil {
 		return err
 	}
 
 	ourEvent := sdk.NewEvent(
 		sdk.EventTypeMessage,
 		sdk.NewAttribute(sdk.AttributeKeyModule, types.ModuleName),
+		sdk.NewAttribute(types.AttributeKeyCodeID, fmt.Sprintf("%d", p.CodeID)),
```
ethanfrey, line 125: This is very important to return

x/wasm/keeper/proposal_handler.go
```diff
@@ -112,15 +114,17 @@ func handleMigrateProposal(ctx sdk.Context, k types.ContractOpsKeeper, p types.M
...
 		sdk.NewAttribute(types.AttributeKeyContract, p.Contract),
+		sdk.NewAttribute(types.AttributeResultData, hex.EncodeToString(data)),
```
ethanfrey, line 127: Seems useful, if we return it other places

x/wasm/keeper/proposal_handler.go
```diff
@@ -94,6 +95,7 @@ func handleInstantiateProposal(ctx sdk.Context, k types.ContractOpsKeeper, p typ
 		sdk.NewAttribute(sdk.AttributeKeyModule, types.ModuleName),
 		sdk.NewAttribute(types.AttributeKeyCodeID, fmt.Sprintf("%d", p.CodeID)),
 		sdk.NewAttribute(types.AttributeKeyContract, contractAddr.String()),
+		sdk.NewAttribute(types.AttributeResultData, hex.EncodeToString(data)),
```
ethanfrey, line 98: Want to call this `types.AttributeResultDataHex`? to make it clear it is hex encoded?
That is what they are doing with ibc binary data (rather than just `string(rawBytes)` as before)

ethanfrey: Are these also returned in the non-proposal methods? 
(too lazy to open up the IDE... but as long as they are all the same, looks good to me)

ethanfrey (approved):
nice

x/wasm/keeper/proposal_handler.go
alpe: I have added this to the other places, too

PR #537, opened 2021-06-11
Verify converted coin

alpe:
Better safe than sorry


x/wasm/keeper/handler_plugin_encoders_test.go
```diff
@@ -515,3 +515,59 @@ func TestEncoding(t *testing.T) {
 		})
 	}
 }
+
+func TestConvertWasmCoinToSdkCoin(t *testing.T) {
```
ethanfrey, line 519: This is a useful check to return errors early (at the cause), rather than some time later (worse if no one verifies again before storing it)

ethanfrey (approved):
Good one

PR #539, opened 2021-06-14
Charge gas for custom event attributes and messages

### LINKED ISSUE #203: Limit  init/migrate/execute payload message size
## Summary
There are currently no limits to the msg payload in `MsgInstantiateContract`, `MsgMigrateContract` and `MsgExecuteContract`. Should we set limit for external/ dispatched messages?

____

#### For Admin Use

- [ ] Not duplicate issue

### LINKED ISSUE #280: Let contracts not abuse the event system
There are no explicit restrictions in wasmd right now for contracts to emit events. We should make sure that this is not abused by malicious actors.

We could charge SDK gas per byte after the contract execution as suggested by @webmaster128.

alpe:
* Resolves #280 
* Resolves #203

Features
* Introduces GasRegister and extension point
* Moved event emitting and charging gas into  response handler method

### DIFF
--- a/x/wasm/keeper/gas_register.go
+++ b/x/wasm/keeper/gas_register.go
@@ -1,9 +1,11 @@
 package keeper
 
 import (
+	"github.com/CosmWasm/wasmd/x/wasm/types"
 	wasmvmtypes "github.com/CosmWasm/wasmvm/types"
 	storetypes "github.com/cosmos/cosmos-sdk/store/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 )
 
 const (
@@ -20,8 +22,11 @@ const (
 	// DefaultCompileCost is how much SDK gas we charge *per byte* for compiling WASM code.
 	DefaultCompileCost uint64 = 2
 	// DefaultEventAttributeDataCost is how much SDK gas we charge *per byte* for attribute data in events.
-	// This is len(key) + len(value)
+	// This is used with len(key) + len(value)
 	DefaultEventAttributeDataCost uint64 = 1
+	// DefaultContractMessageDataCost is how much SDK gas we charge *per byte* of the message that goes to the contract
+	// This is used with len(msg)
+	DefaultContractMessageDataCost uint64 = 1
 	// DefaultPerAttributeCost is how much SDK gas we charge per attribute count.
 	DefaultPerAttributeCost uint64 = 10
 	// DefaultEventAttributeDataFreeTier number of bytes of attribute data we do not charge.
@@ -55,6 +60,7 @@ type WasmGasRegisterConfig struct {
 	EventPerAttributeCost      sdk.Gas
 	EventAttributeDataCost     sdk.Gas
 	EventAttributeDataFreeTier int
+	ContractMessageDataCost    sdk.Gas
 }
 
 // DefaultGasRegisterConfig default values
@@ -66,6 +72,7 @@ func DefaultGasRegisterConfig() WasmGasRegisterConfig {
 		EventPerAttributeCost:      DefaultPerAttributeCost,
 		EventAttributeDataCost:     DefaultEventAttributeDataCost,
 		EventAttributeDataFreeTier: DefaultEventAttributeDataFreeTier,
+		ContractMessageDataCost:    DefaultContractMessageDataCost,
 	}
 }
 
@@ -81,6 +88,9 @@ func NewDefaultWasmGasRegister() WasmGasRegister {
 
 // NewWasmGasRegister constructor
 func NewWasmGasRegister(c WasmGasRegisterConfig) WasmGasRegister {
+	if c.GasMultiplier == 0 {
+		panic(sdkerrors.Wrap(sdkerrors.ErrLogic, "GasMultiplier can not be 0"))
+	}
 	return WasmGasRegister{
 		c: c,
 	}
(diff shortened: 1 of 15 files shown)


x/wasm/keeper/keeper.go
```diff
@@ -35,6 +35,9 @@ const DefaultInstanceCost uint64 = 40_000
 // DefaultCompileCost is how much SDK gas we charge *per byte* for compiling WASM code.
 const DefaultCompileCost uint64 = 2
 
+// DefaultEventAttributeCost is how much SDK gas we charge *per byte* for attribute data in events
+const DefaultEventAttributeCost uint64 = 10
```
ethanfrey, line 39: If they charge 2gas/byte to write to iavl storage, we should charge less to write to the event system.
iavl storage is more critical to keep small imo.

Maybe 1 instead of 10?

x/wasm/keeper/keeper.go
```diff
@@ -804,7 +789,24 @@ func (k Keeper) setContractInfoExtension(ctx sdk.Context, contractAddr sdk.AccAd
 
 // handleContractResponse processes the contract response
 func (k *Keeper) handleContractResponse(ctx sdk.Context, contractAddr sdk.AccAddress, ibcPort string, res *wasmvmtypes.Response) ([]byte, error) {
-	return k.wasmVMResponseHandler.Handle(ctx, contractAddr, ibcPort, res.Submessages, res.Messages, res.Data)
+	return k.doHandleWasmVMResponse(ctx, contractAddr, ibcPort, res.Submessages, res.Messages, res.Attributes, res.Data)
```
ethanfrey, line 792: Why this extra function.
`k.handleContractResponse` -> `k.doHandleWasmVMResponse` ?

Can't you just combine them and add this info in the function body of `handleContractResponse`?

x/wasm/keeper/keeper_test.go
```diff
@@ -284,7 +284,7 @@ func TestInstantiate(t *testing.T) {
 
 	gasAfter := ctx.GasMeter().GasConsumed()
 	if types.EnableGasVerification {
-		require.Equal(t, uint64(0x122a0), gasAfter-gasBefore)
+		require.Equal(t, uint64(0x12368), gasAfter-gasBefore)
```
ethanfrey, line 287: Okay, this example is like 200 gas more expensive?

x/wasm/types/types.go
```diff
@@ -297,25 +297,23 @@ func NewWasmCoins(cosmosCoins sdk.Coins) (wasmCoins []wasmvmtypes.Coin) {
...
 	attrs := []sdk.Attribute{sdk.NewAttribute(AttributeKeyContractAddr, contractAddr.String())}
-
+	var storedBytes int
 	// append attributes from wasm to the sdk.Event
 	for _, l := range wasmOutputAttrs {
 		// and reserve the contract_address key for our use (not contract)
 		if l.Key != AttributeKeyContractAddr {
 			attr := sdk.NewAttribute(l.Key, l.Value)
 			attrs = append(attrs, attr)
+			storedBytes += len(l.Key) + len(l.Value)
```
ethanfrey, line 311: seems like a reasonable counting method

x/wasm/keeper/keeper.go
```diff
@@ -804,7 +789,24 @@ func (k Keeper) setContractInfoExtension(ctx sdk.Context, contractAddr sdk.AccAd
...
+	contractAddr sdk.AccAddress,
+	ibcPort string,
+	subMsg []wasmvmtypes.SubMsg,
+	msgs []wasmvmtypes.CosmosMsg,
+	attrs []wasmvmtypes.EventAttribute,
+	data []byte,
+) ([]byte, error) {
+	// emit all events from this contract itself
+	events, customBytesLen := types.ParseEvents(attrs, contractAddr)
+	ctx.GasMeter().ConsumeGas(k.eventAttributeCost*customBytesLen, "Custom contract event attributes")
```
ethanfrey, line 807: I still kind of like giving a "free tier" like first 100 bytes are free, and charging afterwards.
To encourage people to produce some info, but not ridiculous amounts.

Of course by 100 I mean, "user configurable setting"

ethanfrey (commented):
Nice start.
Added some comments.

What do you think of a price per attribute + price per byte (like iavl storage access)?

x/wasm/keeper/keeper.go
alpe: Free tier and costs per attribute require 2 new cost attributes. It is reasonable to come up with some more advanced pricing model than just bytes charge. I only see that my keeper attributes do not scale well nor provide enough flexibility for custom costs models. 
I will redesign this into a new component that provides defaults but can be substituted with custom logic as extension point

x/wasm/keeper/keeper.go
```diff
@@ -802,22 +758,37 @@ func (k Keeper) setContractInfoExtension(ctx sdk.Context, contractAddr sdk.AccAd
...
-func (k Keeper) gasForContract(ctx sdk.Context) uint64 {
+func (k Keeper) runtimeGasForContract(ctx sdk.Context) uint64 {
 	meter := ctx.GasMeter()
 	if meter.IsOutOfGas() {
 		return 0
 	}
-	remaining := (meter.Limit() - meter.GasConsumedToLimit()) * k.gasMultiplier
-	return remaining
+	if meter.Limit() == 0 { // infinite gas meter with limit=0 and not out of gas
+		return math.MaxUint64
```
alpe, line 785: This was running into an overflow for an infinite gas meter (genesis) before.  Not sure if MaxUint64 makes sense here but it is the highest number available

x/wasm/keeper/gas_register.go
```diff
@@ -91,14 +101,21 @@ func (g WasmGasRegister) NewContractInstanceCosts(pinned bool, msgLen int) store
...
+		panic(sdkerrors.Wrap(types.ErrInvalid, "negative length"))
+	}
 	return g.c.CompileCost * uint64(byteLength)
 }
 
 func (g WasmGasRegister) InstantiateContractCosts(pinned bool, msgLen int) sdk.Gas {
+	if msgLen < 0 {
+		panic(sdkerrors.Wrap(types.ErrInvalid, "negative length"))
+	}
+	dataCosts := sdk.Gas(msgLen) * g.c.ContractMessageDataCost
```
ethanfrey, line 127: Nice charge

x/wasm/keeper/keeper.go
ethanfrey: I would ask @webmaster128 what the largest number we can send on the backend.

I am not sure if it is MaxUint64 or MaxInt64 (or something else). But yeah, let's expose what the Rust runtime can handle.

ethanfrey (approved):
Good stuff. I would just like Simon to review the proper max value

x/wasm/keeper/keeper.go
webmaster128: `MaxUint64` is good. If it does not work anywhere, it is a bug to be fixed.

I just wonder why we need a special treatment for `0`? Is this behaviour standard in the Cosmos SDK metering?

ethanfrey: `InfiniteGasMeter` is a bit "special".

You can keep subtracting gas and it never throws and error, just don't query it.

ethanfrey: https://github.com/cosmos/cosmos-sdk/blob/v0.42.6/store/types/gas.go#L133-L185

webmaster128: Alright, that's not my business. Encoding special cases into 0 just always feels like writing C code.

ethanfrey: Made issue https://github.com/cosmos/cosmos-sdk/issues/9577

PR #541, opened 2021-06-15
Start new gas costs component

alpe:
🚧  Early version

Refactor to a new component
Interface and tests are missing

### DIFF
--- a/x/wasm/keeper/gas_register.go
+++ b/x/wasm/keeper/gas_register.go
@@ -1,0 +1,117 @@
+package keeper
+
+import (
+	wasmvmtypes "github.com/CosmWasm/wasmvm/types"
+	storetypes "github.com/cosmos/cosmos-sdk/store/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// DefaultGasMultiplier is how many cosmwasm gas points = 1 sdk gas point
+// SDK reference costs can be found here: https://github.com/cosmos/cosmos-sdk/blob/02c6c9fafd58da88550ab4d7d494724a477c8a68/store/types/gas.go#L153-L164
+// A write at ~3000 gas and ~200us = 10 gas per us (microsecond) cpu/io
+// Rough timing have 88k gas at 90us, which is equal to 1k sdk gas... (one read)
+//
+// Please note that all gas prices returned to the wasmer engine should have this multiplied
+const DefaultGasMultiplier uint64 = 100
+
+// DefaultInstanceCost is how much SDK gas we charge each time we load a WASM instance.
+// Creating a new instance is costly, and this helps put a recursion limit to contracts calling contracts.
+const DefaultInstanceCost uint64 = 40_000
+
+// DefaultCompileCost is how much SDK gas we charge *per byte* for compiling WASM code.
+const DefaultCompileCost uint64 = 2
+
+// DefaultEventAttributeCost is how much SDK gas we charge *per byte* for attribute data in events
+const DefaultEventAttributeCost uint64 = 10
+
+type GasRegister struct {
+	instanceCost  sdk.Gas
+	compileCost   sdk.Gas
+	gasMultiplier sdk.Gas
+
+	eventAttributeCountCost  sdk.Gas
+	eventAttributeLengthCost sdk.Gas
+	freeTierAttributeLength  int
+}
+
+func DefaultGasRegister() GasRegister {
+	return GasRegister{
+		instanceCost:             DefaultInstanceCost,
+		compileCost:              DefaultCompileCost,
+		gasMultiplier:            DefaultGasMultiplier,
+		eventAttributeCountCost:  0,
+		eventAttributeLengthCost: DefaultEventAttributeCost,
+		freeTierAttributeLength:  0,
+	}
+}
+func NewGasRegister(instanceCost sdk.Gas, compileCost sdk.Gas, gasMultiplier sdk.Gas, eventAttributeCountCost sdk.Gas, eventAttributeLengthCost sdk.Gas, freeTierAttributeLength int) GasRegister {
+	return GasRegister{
+		instanceCost:             instanceCost,
+		compileCost:              compileCost,
+		gasMultiplier:            gasMultiplier,
+		eventAttributeCountCost:  eventAttributeCountCost,
+		eventAttributeLengthCost: eventAttributeLengthCost,
+		freeTierAttributeLength:  freeTierAttributeLength,
+	}
+}
+
+func (g GasRegister) NewContractInstanceCost(pinned bool, msgLen int, labelLength int) storetypes.Gas {
+	return g.InstantiateContractCost(pinned, msgLen)
+}
+
+func (g GasRegister) CompileCost(byteLength int, sourceCodeUrlLen int, builderLen int) storetypes.Gas {
+	return g.compileCost * uint64(byteLength)
+}
+
+func (g GasRegister) InstantiateContractCost(pinned bool, msgLen int) sdk.Gas {
+	if pinned {
+		return 0
+	}
+	return g.instanceCost
+}
+
+func (g GasRegister) ReplyCost(pinned bool, reply wasmvmtypes.Reply) sdk.Gas {
+	var eventGas sdk.Gas
+	msgLen := len(reply.Result.Err)
+	if reply.Result.Ok != nil {
+		msgLen += len(reply.Result.Ok.Data)
+		for _, e := range reply.Result.Ok.Events {
+			msgLen += len(e.Type)
+			eventGas += g.EventCosts(e.Attributes)
+		}
+	}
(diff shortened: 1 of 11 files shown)


x/wasm/keeper/gas_register.go
```diff
@@ -0,0 +1,117 @@
...
+
+// DefaultInstanceCost is how much SDK gas we charge each time we load a WASM instance.
+// Creating a new instance is costly, and this helps put a recursion limit to contracts calling contracts.
+const DefaultInstanceCost uint64 = 40_000
+
+// DefaultCompileCost is how much SDK gas we charge *per byte* for compiling WASM code.
+const DefaultCompileCost uint64 = 2
+
+// DefaultEventAttributeCost is how much SDK gas we charge *per byte* for attribute data in events
+const DefaultEventAttributeCost uint64 = 10
```
ethanfrey, line 25: Where does 10 come from?
Seems higher than IAVL, that's why I ask.

x/wasm/keeper/gas_register.go
```diff
@@ -0,0 +1,117 @@
...
+
+type GasRegister struct {
```
ethanfrey, line 27: This looks like a good idea.

x/wasm/keeper/gas_register.go
```diff
@@ -0,0 +1,117 @@
...
+	for _, l := range evts {
+		storedBytes += len(l.Key) + len(l.Value)
+	}
+	// apply free tier
+	if storedBytes <= g.freeTierAttributeLength {
+		storedBytes = 0
+	} else {
+		storedBytes -= g.freeTierAttributeLength
+	}
+	// total Length * costs + attribute count * costs
```
ethanfrey, line 100: very nice formula.

ethanfrey: Using `sdk.Int` to catch overflow?

ethanfrey (commented):
Looks like a good direction to go.

x/wasm/keeper/gas_register.go
alpe: Random value. I will change it to `1` as proposed by you. It is 10 here as I would have to update the tests as well ;-)

x/wasm/keeper/gas_register.go
alpe: ja, not sure if this required... but better safe than sorry

x/wasm/keeper/gas_register.go
```diff
@@ -0,0 +1,128 @@
...
+	// Please note that all gas prices returned to the wasmer engine should have this multiplied
+	DefaultGasMultiplier uint64 = 100
+	// DefaultInstanceCost is how much SDK gas we charge each time we load a WASM instance.
+	// Creating a new instance is costly, and this helps put a recursion limit to contracts calling contracts.
+	DefaultInstanceCost uint64 = 40_000
+	// DefaultCompileCost is how much SDK gas we charge *per byte* for compiling WASM code.
+	DefaultCompileCost uint64 = 2
+	// DefaultEventAttributeDataCost is how much SDK gas we charge *per byte* for attribute data in events.
+	// This is len(key) + len(value)
+	DefaultEventAttributeDataCost uint64 = 1
```
ethanfrey, line 24: More reasonable numbers here :+1:

x/wasm/keeper/gas_register.go
```diff
@@ -0,0 +1,128 @@
...
+	return GasRegister{
+		instanceCost:               DefaultInstanceCost,
+		compileCost:                DefaultCompileCost,
+		gasMultiplier:              DefaultGasMultiplier,
+		eventPerAttributeCost:      DefaultPerAttributeCost,
+		eventAttributeDataCost:     DefaultEventAttributeDataCost,
+		eventAttributeDataFreeTier: DefaultEventAttributeDataFreeTier,
+	}
+}
+func NewGasRegister(
```
ethanfrey, line 51: This can get quite confusing to construct with 5 items of the same type. Easy to mix up order instance and compile cost for example.

In such cases, I find making the members public and just constructing them like:

```golang
gas := GasRegister {
  InstanceCost: instance,
  CompileCost: compile,
  // ...
}
```

is more robust and more readable.

x/wasm/keeper/gas_register.go
```diff
@@ -0,0 +1,128 @@
...
+		eventAttributeDataCost:     eventAttributeLengthCost,
+		eventAttributeDataFreeTier: freeTierAttributeData,
+	}
+}
+
+func (g GasRegister) NewContractInstanceCost(pinned bool, msgLen int, labelLength int) storetypes.Gas {
+	return g.InstantiateContractCost(pinned, msgLen)
+}
+
+func (g GasRegister) CompileCost(byteLength int, sourceCodeUrlLen int, builderLen int) storetypes.Gas {
```
ethanfrey, line 73: I would remove `sourceCodeUrlLen int, builderLen int` as they have nothing to do with our work (and I would like to remove them). We charge already for the storage gas cost in CodeInfo. They don't factor into the compile work.

x/wasm/keeper/gas_register.go
```diff
@@ -0,0 +1,128 @@
...
+		instanceCost:               instanceCost,
+		compileCost:                compileCost,
+		gasMultiplier:              gasMultiplier,
+		eventPerAttributeCost:      eventAttributeCountCost,
+		eventAttributeDataCost:     eventAttributeLengthCost,
+		eventAttributeDataFreeTier: freeTierAttributeData,
+	}
+}
+
+func (g GasRegister) NewContractInstanceCost(pinned bool, msgLen int, labelLength int) storetypes.Gas {
```
ethanfrey, line 69: Same with `label` here. It gets stored by the sdk and never makes it down to wasm land.

msg does make it down to was land.

ethanfrey: On second look, why do you have these two calls. Why not combine them?

`NewContractInstanceCost`
`InstantiateContractCost`

x/wasm/keeper/gas_register.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+func (g GasRegister) InstantiateContractCost(pinned bool, msgLen int) sdk.Gas {
+	if pinned {
+		return 0
+	}
+	return g.instanceCost
+}
+
+func (g GasRegister) ReplyCost(pinned bool, reply wasmvmtypes.Reply) sdk.Gas {
```
ethanfrey, line 84: Nice

x/wasm/keeper/gas_register.go
```diff
@@ -0,0 +1,128 @@
...
+	var eventGas sdk.Gas
+	msgLen := len(reply.Result.Err)
+	if reply.Result.Ok != nil {
+		msgLen += len(reply.Result.Ok.Data)
+		for _, e := range reply.Result.Ok.Events {
+			msgLen += len(e.Type)
+			eventGas += g.EventCosts(e.Attributes)
```
ethanfrey, line 91: charging each event separately (not combined) will lead to some skew.
The "free tier" is no longer calculated on the total data returned but on each event.

So, I could have 10 events with 100 bytes of attributes and stay in the free tier, which is not desired. I would combine all attributes of all events and then calculate `EventsCost`.

x/wasm/keeper/keeper.go
```diff
@@ -467,10 +440,8 @@ func (k Keeper) reply(ctx sdk.Context, contractAddress sdk.AccAddress, reply was
 		return nil, err
 	}
 
-	// current thought is to charge gas like a fresh run, we can revisit whether to give it a discount later
-	if !k.IsPinnedCode(ctx, contractInfo.CodeID) {
-		ctx.GasMeter().ConsumeGas(k.instanceCost, "Loading CosmWasm module: reply")
-	}
+	replyCost := k.gasRegister.ReplyCost(k.IsPinnedCode(ctx, contractInfo.CodeID), reply)
```
ethanfrey, line 443: This is more clear and extensible.

ethanfrey (commented):
In general, this looks good. And it puts the gas pricing logic in one place so we can easily iterate on it without affecting the keepers any more.

I have a bunch of more detailed comments on the actual gas calculation logic.

ethanfrey:
That said, let's merge this with the other PR, and polish it there, so I can give one final review before you merge to master

x/wasm/keeper/gas_register.go
alpe: Good point. I was a bit afraid of updates from outside the package but I will make a config object with public attributes so that we can have both.

x/wasm/keeper/gas_register.go
alpe: They are very different logic operations for me that could be priced differently by any other chain.

alpe:
Thanks a lot for the feedback. This helped a lot. I will close this PR now and copy the code to the `attribute_cost_200 branch`.

PR #550, opened 2021-07-02
Multiply default wasm size by 2

orkunkl:
On testnets, we multiply this limit by 2. In the past we have seen contracts exceeding this limit. We can increase this by default if you see fit.


alpe:
Good feedback! 💐

ethanfrey:
1.2MB is big. But we have been hitting it in the past. Especially any contract that brings their own crypto (inside wasm)

PR #548, opened 2021-06-25
Upgrade to WasmVM v0.15.1

alpe:
Closes #532

- [x] Messages and submessages are merged; the `replyOn` type now got a new "never" case for the fire and forget messages
- [x] `replyOn` is not an integer but supports the Stringer interface (i.e. `.String()`)
- [x] The interface version was bumped from interface_version_5 to interface_version_6 such that CosmWasm 0.15.0 contracts need to be used
- [x] In some Wasm messages "send" was renamed to "funds" for consistency
- [x] The IBC acknowledgement type was changed slightly
- [x] There is a new Gov message (MsgVote) that needs to be connected
- [x] Deserializing contract results now cost gas. This needs to be configured in a new argument of `vm.Execute` and friends. We messed up storing the value in the `vm` instance, such that it needs to be passed to every call. Sorry for that! The deserialization cost is in gas per byte. It is a fraction, so you can use 20/1, 10/1, 1/1, 1/2, 1/30, ... in there.
- [x] Contract responses now contain events. The exiting `Attributes` should go into a main `wasm` event as before. All other events should be added with a `wasm-` prefix on the event type.
- [x] Set correct release version in `go.mod`
- [x] Upgrade Dockerfile with proper version checksum

## Blocker
~~https://github.com/CosmWasm/wasmvm/pull/235~~
~~https://github.com/CosmWasm/wasmvm/issues/236~~

## Inconsistent
https://github.com/CosmWasm/wasmd/issues/552
https://github.com/CosmWasm/wasmd/issues/553

## Improvements:
https://github.com/CosmWasm/wasmvm/issues/239
~~https://github.com/CosmWasm/wasmvm/issues/238~~

PR #559, opened 2021-07-21
Wasmvm upgrade 557

### LINKED ISSUE #557: Update to wasmvm / cosmwasm 0.16
There will be one more update before

Key Changes:
- Lots of changes to the `IbcChannel*Msg` types
- CounterPartyVersion is no longer in `IbcChannel` but in some of those messages
- Change standard `contract_address` event attribute (for `wasm` type) to `_contract_address`.
- Enforce the contract cannot return any `_*` attributes and thus conflict with any special "trusted" attributes from the runtime.
- Handle custom events from `IbcBasicResponse` and `IbcReceiveResponse` (not just attributes)

ethanfrey:
Closes #557 

Currently tested versions:

- Contracts: cosmwasm v0.16.-rc4
- Engine: wasmvm v0.16.0-rc2

Work done:

- [x] Update wasmd and test contracts 
- [x] Lots of changes to the `IbcChannel*Msg` types
- [x] CounterPartyVersion is no longer in `IbcChannel` but in some of those messages
- [x] Change standard `contract_address` event attribute (for `wasm` type) to `_contract_address`.
- [x] Enforce the contract cannot return any `_*` attributes and thus conflict with any special "trusted" attributes from the runtime.
- [x] Handle custom events from `IbcBasicResponse` and `IbcReceiveResponse` (not just attributes)


x/wasm/keeper/events.go
```diff
@@ -50,8 +51,9 @@ func contractSDKEventAttributes(customAttributes []wasmvmtypes.EventAttribute, c
 	attrs := []sdk.Attribute{sdk.NewAttribute(types.AttributeKeyContractAddr, contractAddr.String())}
 	// append attributes from wasm to the sdk.Event
 	for _, l := range customAttributes {
-		// and reserve the contract_address key for our use (not contract)
-		if l.Key != types.AttributeKeyContractAddr {
+		// FIXME: do we want to error here on invalid events
+		// and reserve all _* keys for our use (not contract)
```
webmaster128, line 55: We do!

webmaster128: Same for the min length check in `newCustomEvents` above. Why silently ignore events with invalid type length?

ethanfrey: This was Alex's choice.
I left a message on the v0.15.0 upgrade PR to that effect. (See https://github.com/CosmWasm/wasmd/pull/548#discussion_r674203149 )

I would make that change in a separate PR as it is unrelated to v0.16.0. But good not to miss it.

webmaster128: Could you create a ticket for those two places? Then we can merge here.

x/wasm/keeper/relay.go
```diff
@@ -60,8 +76,20 @@ func (k Keeper) OnConnectChannel(
 	env := types.NewEnv(ctx, contractAddr)
 	querier := k.newQueryHandler(ctx, contractAddr)
 
+	msg := wasmvmtypes.IBCChannelConnectMsg{}
+	if counterpartyVersion == "" {
```
uint, line 80: I don't really know this codebase, but would it be possible to rely on something more meaningful for this?
```go
if msgConfirm {
    // ..
}
```

uint:
LGTM outside of my one comment!

x/wasm/keeper/relay.go
ethanfrey: Good question, captured in https://github.com/CosmWasm/wasmd/issues/561

PR #562, opened 2021-07-26
Reject invalid attributes

### LINKED ISSUE #560: Return error when contract returns invalid events
Currently these cases are silently ignored.

* Event type < 2 characters long
* Attribute key or value empty ( == "")
* Attribute key starts with _

See https://github.com/CosmWasm/wasmd/pull/559#pullrequestreview-712822088

ethanfrey:
Closes #560

Return error on invalid events/attributes returned from contracts

### DIFF
--- a/x/wasm/keeper/events.go
+++ b/x/wasm/keeper/events.go
@@ -59,8 +59,7 @@ func contractSDKEventAttributes(customAttributes []wasmvmtypes.EventAttribute, c
 	attrs := []sdk.Attribute{sdk.NewAttribute(types.AttributeKeyContractAddr, contractAddr.String())}
 	// append attributes from wasm to the sdk.Event
 	for _, l := range customAttributes {
-		// FIXME: do we want to error here on invalid events
-		// and reserve all _* keys for our use (not contract)
+		// ensure key and value are non-empty (and trim what is there)
 		key := strings.TrimSpace(l.Key)
 		if len(key) == 0 {
 			return nil, sdkerrors.Wrap(types.ErrInvalidEvent, fmt.Sprintf("Empty attribute key. Value: %s", l.Value))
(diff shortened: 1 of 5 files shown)


x/wasm/keeper/events.go
```diff
@@ -34,28 +38,41 @@ func hasWasmModuleEvent(ctx sdk.Context, contractAddr sdk.AccAddress) bool {
...
+		}
+		value := strings.TrimSpace(l.Value)
+		if len(value) == 0 {
+			return nil, sdkerrors.Wrap(types.ErrInvalidEvent, fmt.Sprintf("Empty attribute value. Key: %s", key))
+		}
 		// and reserve all _* keys for our use (not contract)
-		if len(l.Key) > 0 && len(l.Value) > 0 && !strings.HasPrefix(l.Key, types.AttributeReservedPrefix) {
-			attrs = append(attrs, sdk.NewAttribute(l.Key, l.Value))
+		if strings.HasPrefix(key, types.AttributeReservedPrefix) {
+			return nil, sdkerrors.Wrap(types.ErrInvalidEvent, fmt.Sprintf("Attribute starts with %s: %s", types.AttributeReservedPrefix, key))
```
webmaster128, line 73:
```suggestion
			return nil, sdkerrors.Wrap(types.ErrInvalidEvent, fmt.Sprintf("Attribute key starts with reserved prefix %s: '%s'", types.AttributeReservedPrefix, key))
```

x/wasm/keeper/events_test.go
```diff
@@ -100,58 +101,106 @@ func TestNewCustomEvents(t *testing.T) {
...
 			src: wasmvmtypes.Events{{
 				Type: "boom",
 				Attributes: []wasmvmtypes.EventAttribute{
 					{Key: "some", Value: "data"},
 					{Key: "key", Value: ""},
+				},
+			}},
+			isError: true,
+		},
+		"error on empty value": {
```
webmaster128, line 136:
```suggestion
		"error on empty key": {
```

x/wasm/keeper/events.go
```diff
@@ -34,28 +38,41 @@ func hasWasmModuleEvent(ctx sdk.Context, contractAddr sdk.AccAddress) bool {
...
 	for _, l := range customAttributes {
-		// FIXME: do we want to error here on invalid events
+		// ensure key and value are non-empty (and trim what is there)
+		key := strings.TrimSpace(l.Key)
+		if len(key) == 0 {
+			return nil, sdkerrors.Wrap(types.ErrInvalidEvent, fmt.Sprintf("Empty attribute key. Value: %s", l.Value))
+		}
+		value := strings.TrimSpace(l.Value)
+		if len(value) == 0 {
+			return nil, sdkerrors.Wrap(types.ErrInvalidEvent, fmt.Sprintf("Empty attribute value. Key: %s", key))
```
webmaster128, line 69: Why is this check needed? Can't you have contracts that always set an attribute that is potentially empty (e.g. custom note)

ethanfrey: There is no way to search for this if it is empty. The syntax is `<type>.<key> = <value>`

I guess this might be allowed, I should check on the sdk side.

ethanfrey: I added a TODO and asked the sdk team if it is valid.
Shall we make the decision before merge?

webmaster128: Alright. I'm fine either way and do not know what the best way to handle this is.

webmaster128 (approved):

PR #563, opened 2021-07-27
Bump proto

### LINKED ISSUE #558: Bump proto definitions to version 1
The Cosmos SDK [is struggeling](https://github.com/cosmos/cosmos-sdk/issues/9446) with protobuf version changes to the degree that it becomes clear releasing Stargate with "v1beta1" [was a mistake](https://github.com/cosmos/cosmos-sdk/issues/9445). The plan is not to use the simple linear versioning with "v1" from 0.44 on (https://github.com/cosmos/cosmos-sdk/issues/9446#issuecomment-857828835).

We can avoid repeating this mistake and start the version counting now with v1.

Proposed changes in v1:

- [x] `MsgMigrateContract.migrateMsg` -> `.msg` (analogue to MsgExecuteContract)
- [x] `MsgInstantiateContract.initMsg` -> `.msg` (analogue to MsgExecuteContract)

ethanfrey:
Closes #558

- [x] Rename fields in proto
- [x] Rename fields in CLI / REST
- [x] All tests pass
- [x] Bump version to v1


proto/cosmwasm/wasm/v1/proposal.proto
```diff
@@ -65,8 +65,8 @@ message MigrateContractProposal {
   string contract = 4;
   // CodeID references the new WASM code
   uint64 code_id = 5 [ (gogoproto.customname) = "CodeID" ];
-  // MigrateMsg json encoded message to be passed to the contract on migration
-  bytes migrate_msg = 6;
+  // Msg json encoded message to be passed to the contract on migration
+  bytes msg = 6;
```
webmaster128, line 69: Nice changes. I was wondering: why is there no ExecuteContractProposal? Would this be useful?

x/wasm/client/proposal_handler_test.go
```diff
@@ -158,7 +158,7 @@ func TestGovRestHandlers(t *testing.T) {
 				"admin":       "cosmos100dejzacpanrldpjjwksjm62shqhyss44jf5xz",
 				"code_id":     "1",
 				"label":       "https://example.com/",
-				"init_msg":    "my/builder:tag",
+				"msg":         "my/builder:tag",
```
webmaster128, line 161: This is a strange and confusing value for a message

x/wasm/types/proposal.go
```diff
@@ -201,9 +201,9 @@ func (p InstantiateContractProposal) String() string {
   Admin:       %s
   Code id:     %d
   Label:       %s
-  InitMsg:     %q
+  Msg:     %q
```
webmaster128, line 204:
```suggestion
  Msg:         %q
```

x/wasm/types/proposal.go
```diff
@@ -269,8 +269,8 @@ func (p MigrateContractProposal) String() string {
   Contract:    %s
   Code id:     %d
   Run as:      %s
-  MigrateMsg   %q
-`, p.Title, p.Description, p.Contract, p.CodeID, p.RunAs, p.MigrateMsg)
+  Msg   %q
```
webmaster128, line 272:
```suggestion
  Msg          %q
```

x/wasm/types/proposal_test.go
```diff
@@ -474,7 +474,7 @@ func TestProposalStrings(t *testing.T) {
   Admin:       cosmos1qyqszqgpqyqszqgpqyqszqgpqyqszqgpjnp7du
   Code id:     1
   Label:       testing
-  InitMsg:     "{\"verifier\":\"cosmos1qyqszqgpqyqszqgpqyqszqgpqyqszqgpjnp7du\",\"beneficiary\":\"cosmos1qyqszqgpqyqszqgpqyqszqgpqyqszqgpjnp7du\"}"
+  Msg:     "{\"verifier\":\"cosmos1qyqszqgpqyqszqgpqyqszqgpqyqszqgpjnp7du\",\"beneficiary\":\"cosmos1qyqszqgpqyqszqgpqyqszqgpqyqszqgpjnp7du\"}"
```
webmaster128, line 477:
```suggestion
  Msg:         "{\"verifier\":\"cosmos1qyqszqgpqyqszqgpqyqszqgpqyqszqgpjnp7du\",\"beneficiary\":\"cosmos1qyqszqgpqyqszqgpqyqszqgpqyqszqgpjnp7du\"}"
```

x/wasm/types/proposal_test.go
```diff
@@ -487,7 +487,7 @@ func TestProposalStrings(t *testing.T) {
...
```
webmaster128, line 490:
```suggestion
  Msg:         "{\"verifier\":\"cosmos1qyqszqgpqyqszqgpqyqszqgpqyqszqgpjnp7du\",\"beneficiary\":\"cosmos1qyqszqgpqyqszqgpqyqszqgpqyqszqgpjnp7du\"}"
```

x/wasm/types/proposal_test.go
```diff
@@ -500,7 +500,7 @@ func TestProposalStrings(t *testing.T) {
   Admin:       
   Code id:     1
   Label:       testing
-  InitMsg:     "{\"verifier\":\"cosmos1qyqszqgpqyqszqgpqyqszqgpqyqszqgpjnp7du\",\"beneficiary\":\"cosmos1qyqszqgpqyqszqgpqyqszqgpqyqszqgpjnp7du\"}"
+  Msg:     "{\"verifier\":\"cosmos1qyqszqgpqyqszqgpqyqszqgpqyqszqgpjnp7du\",\"beneficiary\":\"cosmos1qyqszqgpqyqszqgpqyqszqgpqyqszqgpjnp7du\"}"
```
webmaster128, line 503:
```suggestion
  Msg:         "{\"verifier\":\"cosmos1qyqszqgpqyqszqgpqyqszqgpqyqszqgpjnp7du\",\"beneficiary\":\"cosmos1qyqszqgpqyqszqgpqyqszqgpqyqszqgpjnp7du\"}"
```

x/wasm/types/proposal_test.go
```diff
@@ -512,7 +512,7 @@ func TestProposalStrings(t *testing.T) {
   Contract:    cosmos18vd8fpwxzck93qlwghaj6arh4p7c5n89uzcee5
   Code id:     1
   Run as:      cosmos1qyqszqgpqyqszqgpqyqszqgpqyqszqgpjnp7du
-  MigrateMsg   "{\"verifier\":\"cosmos1qyqszqgpqyqszqgpqyqszqgpqyqszqgpjnp7du\"}"
+  Msg   "{\"verifier\":\"cosmos1qyqszqgpqyqszqgpqyqszqgpqyqszqgpjnp7du\"}"
```
webmaster128, line 515:
```suggestion
  Msg          "{\"verifier\":\"cosmos1qyqszqgpqyqszqgpqyqszqgpqyqszqgpjnp7du\"}"
```

x/wasm/types/wasmer_engine.go
```diff
@@ -82,7 +82,7 @@ type WasmerEngine interface {
 	// replace it. This allows it to run a migration step if needed, or return an error if unable to migrate
 	// the given data.
 	//
-	// MigrateMsg has some data on how to perform the migration.
+	// Msg has some data on how to perform the migration.
```
webmaster128, line 85: This propably refers to the argument `migrateMsg` which is not renamed. Let's use the variable name here in text, no matter which one we use.

webmaster128 (commented):
🐎

proto/cosmwasm/wasm/v1/proposal.proto
ethanfrey: Hmmm.. what would it do?

Migrate will override any other auth.

InstantiateProposal is useful if we want to instantiate a code ID which is governance restricted (you can set it so only governance can create new contracts, to avoid eg. everyone making cw20 tokens on a non-DeFi chain)

What would the Execute Proposal do?

All I can think of is to execute a message as an arbitrary account. Which is as powerful as "allow governance to forge a message from arbitrary user". You can make an issue to describe this idea if you think it would be useful and let's wait for Alex to discuss it.

x/wasm/client/proposal_handler_test.go
ethanfrey: You are right. I should set this to something different than a builder.

x/wasm/types/wasmer_engine.go
ethanfrey: Over aggressive renaming.

Should I rename the variables in this interface as well? I guess it has nothing to do with proto and is non API-breaking, so we can do it later.

webmaster128: I don't mind. I don't even understand why this file exist and those signatures are not imported from wasmvm.

webmaster128 (approved):

x/wasm/types/wasmer_engine.go
ethanfrey: This is an interface that matches wasmvm (the implementation), but allows him to swap out the vm with a mock that let's us test mock contracts in Go. This makes for much simpler unit testing of the higher level logic (eg. go contract that returns such and such events in reply, then see how the are assembled through the keeper)

PR #564, opened 2021-07-27
Trim code info

### LINKED ISSUE #540: Remove (often unused) contract urls from CodeInfo
Early 2020, we chose to store the `builder` and `url` for the code in the `CodeInfo` to enable easier validation of the code. However, in practice, this is often unfilled or filled with incorrect values. And no tooling has been produced to properly fill it out nor verify contracts using it.

In the meantime, most `cosmwasm-plus` based contracts now store correct and up-to-date metadata in the contract itself to enable (future) migrations. We could display this in frontend UIs if present and remove the older `builder`/`url` fields.

The [cw2 spec](https://github.com/CosmWasm/cosmwasm-plus/tree/main/packages/cw2) defines a format and raw storage location for the contract metadata. And you can view this set in many [contracts like `cw20-base`](https://github.com/CosmWasm/cosmwasm-plus/blob/main/contracts/cw20-base/src/contract.rs#L20-L22)

ethanfrey:
Closes #540

(More changes to protobuf as we bump to v1)


proto/cosmwasm/wasm/v1/proposal.proto
```diff
@@ -20,11 +20,8 @@ message StoreCodeProposal {
   string run_as = 3;
   // WASMByteCode can be raw or gzip compressed
   bytes wasm_byte_code = 4 [ (gogoproto.customname) = "WASMByteCode" ];
-  // Source is a valid absolute HTTPS URI to the contract's source code,
-  // optional
-  string source = 5;
-  // Builder is a valid docker image name with tag, optional
-  string builder = 6;
+  // 5 and 6 intentionally left empty (used in v1beta1)
```
webmaster128, line 23: Maybe this leads to better doc generation? See https://developers.google.com/protocol-buffers/docs/proto#reserved

```suggestion
  // Used in v1beta1
  reserved 5, 6;
```

webmaster128 (commented):
That would make things simpler foor sure

proto/cosmwasm/wasm/v1/proposal.proto
ethanfrey: I vaguely remembered there was some way to mark it as reserved in protobuf.
Nice find.

ethanfrey: I just did that fix-up

PR #567, opened 2021-07-27
Adjust default msg gas

### LINKED ISSUE #553: Revisit fee for wasm msg size
With the new deserialization fee in wasmvm we charge twice

ethanfrey:
Closes #553

### DIFF
--- a/x/wasm/keeper/gas_register.go
+++ b/x/wasm/keeper/gas_register.go
@@ -26,7 +26,7 @@ const (
 	DefaultEventAttributeDataCost uint64 = 1
 	// DefaultContractMessageDataCost is how much SDK gas is charged *per byte* of the message that goes to the contract
 	// This is used with len(msg)
-	DefaultContractMessageDataCost uint64 = 1
+	DefaultContractMessageDataCost uint64 = 0
 	// DefaultPerAttributeCost is how much SDK gas we charge per attribute count.
 	DefaultPerAttributeCost uint64 = 10
 	// DefaultPerCustomEventCost is how much SDK gas we charge per event count.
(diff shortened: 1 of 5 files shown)


x/wasm/keeper/gas_register.go
```diff
@@ -144,7 +145,7 @@ func (g WasmGasRegister) ReplyCosts(pinned bool, reply wasmvmtypes.Reply) sdk.Ga
 		msgLen += len(reply.Result.Ok.Data)
 		var attrs []wasmvmtypes.EventAttribute
 		for _, e := range reply.Result.Ok.Events {
-			msgLen += len(e.Type)
+			eventGas += sdk.Gas(len(e.Type)) * g.c.EventAttributeDataCost
```
webmaster128, line 148: I think this better expresses the intend because the factor is: _gas per byte_

```suggestion
			eventGas += sdk.Gas(len(e.Type) * g.c.EventAttributeDataCost)
```

webmaster128: Oh boy, those types. Multiplying gas with gas makes no sense at all but is done everywhere. I give up here.

x/wasm/keeper/gas_register.go
```diff
@@ -26,7 +26,8 @@ const (
 	DefaultEventAttributeDataCost uint64 = 1
 	// DefaultContractMessageDataCost is how much SDK gas is charged *per byte* of the message that goes to the contract
 	// This is used with len(msg)
-	DefaultContractMessageDataCost uint64 = 1
+	// Note: also used for error fields on reply, and data on reply
```
webmaster128, line 29: Maybe we should add that the contract charges gas when deserializing those messages.

webmaster128 (commented):
👍

x/wasm/keeper/gas_register.go
ethanfrey: It's not about intent. It is about forcing the GO compiler.

sdk.Gas is an alias for `uint64`. `len()` returns `int`. You cannot multiply `int` with `uint64` without a cast first. Thus I had to cast it before multiplying.

PR #568, opened 2021-07-28
Ibc query only returns open channels

### LINKED ISSUE #556: IbcQuery only returns channels in Established state
Looking at https://github.com/CosmWasm/wasmd/blob/master/x/wasm/keeper/query_plugins.go#L197-L253, it will return channels in any state,

Outside of the handshake, the only state we care about is `STATE_OPEN` which can send and receive packets: https://github.com/cosmos/ibc-go/blob/2c880a22e9f9cc75f62b527ca94aa75ce1106001/proto/ibc/core/channel/v1/channel.proto#L53-L68

Returning closed channels in the query could confuse a contract. I would thus propose to filter in the query and only return channels in `STATE_OPEN` from the `IbcQuery` queries.

Note: the alternative is to return all and add a state enum to IbcChannel to capture this info. After some discussion, we don't see any reason to provide these non-open channels, it only adds complexity, and there is a backwards compatible approach to add this info if we do want to support these in a later 1.x version.

ethanfrey:
Closes #556

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 **Implemented Enhancements:**
 - Reject invalid events/attributes returned from contracts [\#560](https://github.com/CosmWasm/wasmd/pull/560)
+- IBC Query methods from Wasm contracts only return OPEN channels [\#568](https://github.com/CosmWasm/wasmd/pull/568)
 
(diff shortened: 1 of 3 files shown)


webmaster128 (commented):
LGTM but I'm not very familiar with this tech

ethanfrey:
It's really just adding a filter to a state. If you don't see obvious typos, then it is likely good.

PR #570, opened 2021-07-28
Better ibc contract interface

### LINKED ISSUE #561: Split Connect/Close callbacks in IbcContract interface into 2 variants.
Question: do we want to do so?

We use an argument to construct IbcConnectInit or IbcConnectTry from an IbcConnectOpen callback. We could split this into 2 methods, or we could use a IbcConnectOpenMsg as an argument (which captures the 2 variants).

It could definitely be cleaner than mine, which I did to make minimal changes. I'd like feedback from @alpe to do this.

ethanfrey:
Closes #561

We can make this less verbose once https://github.com/CosmWasm/wasmvm/pull/254 is merged and tagger

### DIFF
--- a/x/wasm/keeper/relay.go
+++ b/x/wasm/keeper/relay.go
@@ -9,6 +9,8 @@ import (
 	"time"
 )
 
+var _ types.IBCContractKeeper = (*Keeper)(nil)
+
 // OnOpenChannel calls the contract to participate in the IBC channel handshake step.
 // In the IBC protocol this is either the `Channel Open Init` event on the initiating chain or
 // `Channel Open Try` on the counterparty chain.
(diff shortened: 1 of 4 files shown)


ethanfrey:
After merging this, I would cut a v0.18.0-rc1 as the remaining issues are mainly docs, with the possible change of an event type or two that we emit. Pretty close to release-ready API and something to cut CosmJS against

x/wasm/keeper/relay.go
```diff
@@ -31,18 +31,6 @@ func (k Keeper) OnOpenChannel(
 	env := types.NewEnv(ctx, contractAddr)
 	querier := k.newQueryHandler(ctx, contractAddr)
 
-	msg := wasmvmtypes.IBCChannelOpenMsg{}
-	if counterpartyVersion == "" {
```
webmaster128, line 35: Nice we don't need to rely on such logic anymore

webmaster128 (approved):

PR #573, opened 2021-08-02
Remove obsolete cli params

alpe:
Minor cleanup


ethanfrey (approved):
Good catch. I removed the flag references but forgot the usage

PR #578, opened 2021-08-04
Add organisation to grpc gateway path

alpe:
This is a convention used in the [cosmos-sdk](https://github.com/cosmos/cosmos-sdk/blob/v0.42.8/proto/cosmos/bank/v1beta1/query.proto#L16)

The GRPC gateway is a generated REST interface that is translates queries/responses into/from protobuf.


webmaster128 (approved):
LGTM.

This does not affect CosmJS at all because the querry types are already in `cosmwasm/wasm/v1/query`. It only affects the LCD API, which we do not use.

CHANGELOG.md
```diff
@@ -15,9 +15,15 @@
   [\#565](https://github.com/CosmWasm/wasmd/pull/565)
 
 **Implemented Enhancements:**
-
+- Wasmvm upgrade 557 [\#559](https://github.com/CosmWasm/wasmd/pull/559) ([ethanfrey](https://github.com/ethanfrey))
```
ethanfrey, line 18: Thanks for adding the changeling

contrib/local/03-grpc-queries.sh
```diff
@@ -6,26 +6,26 @@ DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
 echo "-----------------------"
 PROTO_THRD="$DIR/../../third_party/proto"
 PROTO_WASMD="$DIR/../../proto"
-PROTO_WASMD_QUERY="$PROTO_WASMD/cosmwasm/wasm/v1beta1/query.proto"
+PROTO_WASMD_QUERY="$PROTO_WASMD/cosmwasm/wasm/v1/query.proto"
```
ethanfrey, line 9: Good stuff here

proto/cosmwasm/wasm/v1/query.proto
```diff
@@ -46,11 +47,11 @@ service Query {
   }
   // Code gets the binary code and metadata for a singe wasm code
   rpc Code(QueryCodeRequest) returns (QueryCodeResponse) {
-    option (google.api.http).get = "/wasm/v1/code/{code_id}";
+    option (google.api.http).get = "/cosmwasm/wasm/v1/code/{code_id}";
```
ethanfrey, line 50: Make sense. This is REST path.

ethanfrey (approved):
lgtm

Issue #574, opened 2021-08-03
Expose required code "features" in store code event


alpe:
Get the features from the analyzer

webmaster128:
The funcationality is available in `.AnalyzeCode(…).RequiredFeatures` released with wasmvm 0.16-0-rc3.

ethanfrey:
Closed by #586

Issue #552, opened 2021-07-05
Events/ event attributes on reply


alpe:
Currently we append them as new custom events or wasm event (with attributes). 
This means there can be duplicate events for `wasm` or custom event types. Not sure if this is (still) an issue for Tendermint. 

I would also be important to document the desired behaviour in cosmwasm. On Reply the `data` would be overwritten but events/attributes are not.

ethanfrey:
This definitely needs documentation. The whole events system does.

Tendermint does still have this issue... It will improve in 0.35, see [comment](https://github.com/CosmWasm/wasmd/issues/448#issuecomment-888083059) and https://github.com/tendermint/tendermint/pull/6634

It would also be good to write some tests to demonstrate expected behavior as those seem to stay up-to-date much better than docs. (And are always more clear)

ethanfrey:
Defined in #581 
Closed by #586 

We will have multiple `wasm` events, but separated by the various entry point events

Issue #440, opened 2021-03-05
Revisit event types 


alpe:
We emit events of type `sdk.EventTypeMessage` mostly. They could become more useful with specific types.

ethanfrey:
Nice idea, the `TypedEvent` stuff, right?

I saw this as I was parsing the abci.Events returned when dispatching an sdk.Msg and passing them to the EventManager and wondering why the method was deprecated. It would be good to use this for the ones we generate manually.

alpe:
Sorry, I should have been more precise. This is not about the deprecated method but the string we pass as first argument https://github.com/CosmWasm/wasmd/blob/v0.15.1/x/wasm/internal/keeper/msg_server.go#L33 .
If we would use more concrete types like "contract_stored" instead of "message" then it would be easier for a client to subscribe to the right one. 
I found some good example in the [distribution module](https://github.com/cosmos/cosmos-sdk/blob/v0.41.4/x/distribution/types/events.go#L5-L10)

Ref [sdk-events](https://docs.cosmos.network/master/core/events.html)

PR #581, opened 2021-08-05
Document event handling

### LINKED ISSUE #448: Filter out Message events before calling submessage response
Follow up from https://github.com/CosmWasm/wasmd/pull/441#discussion_r590474261

ethanfrey:
This goes into great detail on both how Tendermint and the Cosmos SDK handle events, as well as how we work with them in `x/wasm` and when interfacing with smart contracts.

This documents a spec on the desired format for implementing #440

Closes #448 and remove all filtering of `message` events.

Document a solution to #552 which is a small adjustment from the current behaviour.

Documents a format for #574

We may want to raise new issues for any other changes defined here if they don't fit in the above 3 issues.

### DIFF
--- a/EVENTS.md
+++ b/EVENTS.md
@@ -1,0 +1,69 @@
+# Event System
+
+## Usage in the SDK
+
+Events are an essential part of the Cosmos SDK. They are similar to "logs" in Ethereum and allow a blockchain
+app to attach key-value pairs to a transaction that can later be used to search for it or extract some information
+in human readable form. Events are not written to the application state, nor do they form part of the AppHash,
+but mainly intended for client use (and become an essential API for any reactive app or app that searches for txs). 
+
+In contrast, transactions also have a binary "data" field that is part of the AppHash (provable with light client proofs,
+part of consensus). This data is not searchable, but given a tx hash, you can be gauranteed what the data returned is.
+This is often empty, but sometimes custom protobuf formats to return essential information from an execution.
+
+Every message in the SDK may add events to the EventManager and these are then added to the final ABCI result that is returned
+to Tendermint. Events are exposed in 3 different ways over the Tendermint API (which is the only way a client can query).
+First of all is the `events` field on the transaction result (when you query a transaction by hash, you can see all event emitted
+by it). Secondly is the `log` field on the same transaction result. And third is the query interface to search or subscribe for
+transactions. 
+
+The `log` field actually has the best data. It contains an array of array of events. The first array is one entry per incoming message.
+Transactions in the Cosmos SDK may consist of multiple messages that are executed atomically. Maybe we send tokens, then issue a swap
+on a DEX. Each action would return it's own list of Events and in the logs, these are separated. For each message, it maintains a list
+of Events, exactly in the order returned by the application. This is JSON encoded and can be parsed by a client.
+
+In Tendermint 0.35, the `events` field will be one flattened list of events over all messages. Just appending the lists returned
+from each message. However, currently (until Tendermint 0.34 used in Cosmos SDK 0.40-0.43), they are flattened on type. Meaning all events
+with type `wasm` get merged into one. This makes the API not very useful to understanding more complex events currently. (TODO: link PR fixing this)
+
+In the search/subscribe interface, you can query for transactions by `AND`ing a number of conditions. Each is expressed like
+`<type>.<key>=<value>`. For example, `message.signer=cosmos1234567890`. It will return all transactions that emitted an event matching this filter.
+
+### Examples
+
+TODO: show event structure.
+
+TODO: contrast flattened/unflattened events
+
+### Standard Events in the SDK
+
+TODO: what is added by the AnteHandlers
+
+TODO: what is emitted by bank (transfer event), as this is a very important base event
+
+## Usage in wasmd
+
+In `x/wasm` we also use Events system. On one hand, `x/wasm` emits standard event for each message it processes to convey,
+for example, "uploaded code, id 6" or "executed code, address wasm1234567890". Furthermore, it allows contracts to
+emit custom events based on their execution state, so they can for example say "dex swap, BTC-ATOM, in 0.23, out 512"
+which require internal knowledge of the contract and is very useful for custom dApp UIs.
+
+In addition, when a smart contract executes a SubMsg and processes the reply, it receives not only the `data` response
+from the message exection, but also the list of events 
+
+### Standard Events in x/wasm
+
+TODO: document what we emit in `x/wasm` regardless of the contract return results
+
+### Emitted Custom Events from a Contract
+
+TODO: document how we process attributes and events fields in the Response.
+
+* Base event `wasm`
+* Event name mangling (prepend `wasm-`)
+* Append trusted attribute _contract_address
+* Validation requirements (_ reserved, non-empty, min-length 2 for type)
+
+## Event Details for wasmd
+
+Beyond the basic Event system and emitted events, we 
--- a/EVENTS.md
+++ b/EVENTS.md
@@ -37,7 +37,7 @@ TODO: contrast flattened/unflattened events
 
 ### Standard Events in the SDK
 
-TODO: what is added by the AnteHandlers
+TODO: what is added by the AnteHandlers (message.signer? auth?)
 
 TODO: what is emitted by bank (transfer event), as this is a very important base event
 
@@ -66,4 +66,24 @@ TODO: document how we process attributes and events fields in the Response.
 
 ## Event Details for wasmd
 
-Beyond the basic Event system and emitted events, we 
+Beyond the basic Event system and emitted events, we must handle more advanced cases in `x/wasm`
+and thus add some more logic to the event processing. Remember that CosmWasm contracts dispatch other
+messages themselves, so far from the flattened event structure, or even a list of list (separated
+by message index in the tx), we actually have a tree of messages, each with their own events. And we must
+flatten that in a meaningful way to feed it into the event system.
+
+Furthermore, with the sub-message reply handlers, we end up with eg. "Contract A execute", "Contract B execute",
+"Contract A reply". If we return all events by all of these, we may end up with many repeated event types and
+a confusing results, especially for Tendermint 0.34 where they are merged together.
+
+While designing this, we wish to make something that is usable with Tendermint 0.34, but focus on using the
+behavior of Tendermint 0.35+ (which is the same behavior as we have internally in the SDK... submessages
+all have their own list of Events). Thus, we may emit more events than in previous wasmd versions (as we assume
+they will be returned in an ordered list rather than merged).
+
+### Combining Events from Sub-Messages
+
+TODO
+### Exposing Events to Reply
+
+TODO
--- a/EVENTS.md
+++ b/EVENTS.md
@@ -167,14 +167,12 @@ sdk.NewEvent(
     "instantiate",
     sdk.NewAttribute("code_id", fmt.Sprintf("%d", msg.CodeID)),
     sdk.NewAttribute("_contract_addr", contractAddr.String()),
-    sdk.NewAttribute("result", hex.EncodeToString(data)),
 )
 
 // Execute Contract
 sdk.NewEvent(
     "execute",
     sdk.NewAttribute("_contract_addr", contractAddr.String()),
-    sdk.NewAttribute("result", hex.EncodeToString(data)),
 )
 
 // Migrate Contract
@@ -183,7 +181,6 @@ sdk.NewEvent(
     // Note: this is the new code id that is being migrated to
     sdk.NewAttribute("code_id", fmt.Sprintf("%d", msg.CodeID)),
     sdk.NewAttribute("_contract_addr", contractAddr.String()),
-    sdk.NewAttribute("result", hex.EncodeToString(data)),
 )
 
 // Set new admin
@@ -339,7 +336,6 @@ sdk.NewEvent(
 sdk.NewEvent(
     "execute",
     sdk.NewAttribute("_contract_addr", contractAddr.String()),
-    sdk.NewAttribute("result", hex.EncodeToString(data)),
 ),
 sdk.NewEvent(
     "wasm",
@@ -352,7 +348,6 @@ sdk.NewEvent(
     "instantiate",
     sdk.NewAttribute("code_id", fmt.Sprintf("%d", msg.CodeID)),
     sdk.NewAttribute("_contract_addr", newContract.String()),
-    sdk.NewAttribute("result", hex.EncodeToString(initData)),
 )
 // didn't emit any attributes, but one event
 sdk.NewEvent(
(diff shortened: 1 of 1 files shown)


ethanfrey:
@alpe  I took inspiration from your PR #583 and your issues, and went on to define the entire event system.
This is not (yet) cut in stone, so please make suggestions on how to improve it. Adding more fields to the custom `execute` etc events? Renaming `code_id` to `_code_id`? Injecting `_code_id` in all `wasm` events? etc...

This is one possible definition, let's adjust it to make it the best it can be.

ethanfrey:
PD Does it make sense to use Go types when listing events? (It was a lot shorter than writing proper JSON all those places, which is why I did it, but maybe there is a better format we could use)

EVENTS.md
```diff
@@ -0,0 +1,416 @@
...
+consistent way possible.
+
+### Combining Events from Sub-Messages
+
+Each time a contract is executed, it not only returns the `message` event from its call, the `execute` event for the
+contact and the `wasm` event with any custom fields from the contract itself. It will also return the same set of information
+for all messages that it returned, which were later dispatched.
+
+Until 0.18.0, we stripped out the `message` events that were returned when dispatching `Response.messages`. However,
+they contain useful information to trace the topography of the call, especially when we call into native modules,
```
alpe, line 310: If all sdk modules would follow the `distribution` event design, I would totally agree on this approach. Unfortunately the **bank** module emits a `message` event in the keeper. Probably others as well. 
We call the keeper when we transfer coins to the contract account. This would result into 2 different message events where only the wasm one is correct.
A more consistent approach was the message filtering so that we only track the message initiating the whole process via event. We do loose information going this way, but I would prefer this rather than emitting false positives.
We could also address the bank message issue by filtering out the bank message in our `BankCoinTransferrer`. For native modules calling the bank keeper this would not work though.

EVENTS.md
```diff
@@ -0,0 +1,416 @@
...
+
+### Emitted Custom Events from a Contract
+
+When a CosmWasm contract returns a `Response` from one of the calls, it may return a list of attributes as well as a list
+of events (in addition to data and a list of messages to dispatch). These are then processed in `x/wasm` to create events that
+are emitted to the blockchain.
+
+Every contract execution, be it execute, instantiate, migrate, reply, will receive a `wasm` type event. This event will
+always be tagged with `_contract_address` by the Go module, so this is trust-worthy. The contract itself cannot overwrite
+this field. (QUESTION: do we want to emit `_code_id` as well for the code id that was just executed?) Beyond this, if the
```
alpe, line 236: The data would be available but I would prefer to keep the attributes to a minimum.
If people are interested in historical code versions we have the migration event and also the code history.

EVENTS.md
```diff
@@ -0,0 +1,416 @@
...
+Until 0.18.0, we stripped out the `message` events that were returned when dispatching `Response.messages`. However,
+they contain useful information to trace the topography of the call, especially when we call into native modules,
+and it makese sense to maintain the full information.
+
+Note that these are all appended in order called. So if I execute a contract, which returns two messages, one to instantiate a new
+contract and the other to set the withdrawl address, while also using `ReplyOnSuccess` for the instantiation (to get the
+address), it will emit a series of events that looks something like this:
+
+```go
+/// original execution (top-level message)
```
alpe, line 318: This list of events describe the process flow but as events are emitted **after** execution the actual order would be different. I think L321 would be the last event emitted when the original message was processed. L330 would be the first one when we handle the first contract response.

EVENTS.md
```diff
@@ -0,0 +1,416 @@
...
+    sdk.NewAttribute("sender", contractAddr.String()),  
+),
+sdk.NewEvent(
+    "instantiate",
+    sdk.NewAttribute("code_id", fmt.Sprintf("%d", msg.CodeID)),
+    sdk.NewAttribute("_contract_addr", newContract.String()),
+    sdk.NewAttribute("result", hex.EncodeToString(initData)),
+)
+sdk.NewEvent(
+    "wasm",
```
alpe, line 348: Looking at this event with some distance now, I don't see how it adds value. Before the contract address was not available but it is now by the wasm operation event. Would be nice to drop this functionality and let people emit custom events when needed. That would be easier to explain in dev doc and no "empty" event would be emitted by default.

EVENTS.md
```diff
@@ -0,0 +1,416 @@
...
+```
+
+No further information will be added to the generic "message" type, but rather be contained in a more context-specific event type.
+Here are some examples:
+
+```go
+// Store Code
+sdk.NewEvent(
+    "store_code",
+    // Update in 0.18: _code_id is also a reserved prefix
```
alpe, line 158: Do we need to make it a reserved prefix". For custom events we are injecting the contract address so there was the requirement but we are not doing the same for code_id. Maybe flattening?

EVENTS.md
```diff
@@ -0,0 +1,416 @@
...
+
+// instantiating contract (returned message)
+sdk.NewEvent(
+    "message",
+    sdk.NewAttribute("module", "wasm"),
+    sdk.NewAttribute("sender", contractAddr.String()),  
+),
+sdk.NewEvent(
+    "instantiate",
+    sdk.NewAttribute("code_id", fmt.Sprintf("%d", msg.CodeID)),
```
alpe, line 343: This and others, please add `_` prefix to code_id

alpe (commented):
Excellent writeup and best doc on event internals I have seen in cosmos. ❤️

EVENTS.md
ethanfrey: Afaik, we call the router for most dispatching (except ibc) and thus hit the message server as well.

I also hear your point about the non standard issue. But it is very useful to have a separator between the different sub messages.

How about this:
1. Filter out the inconsistent message event emitted by the module.
2. Always prepend a standard event type on the beginning of each submessage. We can define what it is, it is created by x/wasm and will be very consistent.

EVENTS.md
ethanfrey: Fair enough, it is not really needed unless someone wants to look for all cw20 transfers.

But a better query engine could handle that as well without overloading events. I will remove

EVENTS.md
ethanfrey: We can change what happens. I want to document the desired behaviour in here (which will be actual behaviour with the 0.18.0 tag), not the current behaviour.

We have control in x/wasm to do something like:

* Keeper.execute called
* call wasmvm.execute
* emit `execute` message including the data from the response
* emit `wasm` and `wasm-*` events from the contract
* for each message in Response:
  * emit some "start sub message" event (if desired)
  * call the execution logic here (and emit all events like this, recursively - filtering message event)
  * emit some "stop sub message event"?
* emit any other events we wish

ethanfrey: As to the `message: module=wasm`, this is emitted in the message server and we can perform this any way we want to.
We can add it top the event manager before calling the keeper, or afterwards. I think placing it first (as in distribution) makes more sense than at the end (in bank).

What do you think the *desired* behaviour here is?

EVENTS.md
ethanfrey: So, if I return non-empty `attributes`, then it will get this `wasm` event (tagged).
If I return empty attributes, it will not get this event, right?

The one issue I want to bring up (which came up from Secret network some time ago) is they want to be able to search for all actions on a contract easily. We can now search for `execute._contract_address="foo"`. However, maybe we also want to include `instantiate._contract_address="foo"` and `migrate._contract_address="foo"`. I guess all data is there.

So, yeah, if we say don't emit `wasm` event on empty attributes, I am fine with that

EVENTS.md
ethanfrey: Let's leave this as code_id, as we will not be injecting it into `wasm*` events

ethanfrey:
I integrated feedback on the two clear suggestions (no _code_id injection and not emitted wasm event at all if no attributes in the Response).

Let's talk more about the use of `message` events and how to separate sub-events

EVENTS.md
alpe: I would go one step further and drop the whole event. Why stick with a `wasm` event when we have custom events now for a contract? I don't see the use case for a client anymore. But there is complexity in wasmd to handle the case and complexity in wasm contracts for people to learn and understand the framework. Both sides would benefit from a clean design that does not support 2 ways of storing data in events.

EVENTS.md
```diff
@@ -0,0 +1,419 @@
...
+they contain useful information to trace the topography of the call, especially when we call into native modules,
+and it makese sense to maintain the full information.
+
+Note that these are all appended in order called. So if I execute a contract, which returns two messages, one to instantiate a new
+contract and the other to set the withdrawl address, while also using `ReplyOnSuccess` for the instantiation (to get the
+address), it will emit a series of events that looks something like this:
+
+```go
+/// original execution (top-level message)
+sdk.NewEvent(
```
ethanfrey, line 322: Folding in the two open issues, I would agree with filtering out all `message` events except the one top-level one (what was included in transaction), but injecting some custom start/stop events for submsg boundaries. These have no attributes so cannot be searched, only used when processing the events to recreated the execution tree.

I would propose something like this: 

```go
/// original execution (top-level message is the only one that gets the message tag)
sdk.NewEvent(
    "message",
    sdk.NewAttribute("module", "wasm"),
    sdk.NewAttribute("sender", msg.Sender),  
),

// this is where we enter the keeper, and all we would see if this was called internally from another contract
sdk.NewEvent(
    "execute",
    sdk.NewAttribute("_contract_addr", contractAddr.String()),
    sdk.NewAttribute("result", hex.EncodeToString(data)),
),
sdk.NewEvent(
    "wasm",
    sdk.NewAttribute("_contract_addr", contractAddr.String()),
    sdk.NewAttribute("custom", "from contract"),
),

// instantiating contract (returned message)
sdk.NewEvent(
    "submsg_start",
),
sdk.NewEvent(
    "instantiate",
    sdk.NewAttribute("code_id", fmt.Sprintf("%d", msg.CodeID)),
    sdk.NewAttribute("_contract_addr", newContract.String()),
    sdk.NewAttribute("result", hex.EncodeToString(initData)),
)
// didn't emit any attributes, but one event
sdk.NewEvent(
    "wasm-custom",
    sdk.NewAttribute("_contract_addr", newContract.String()),
    sdk.NewAttribute("foobar", "baz"),
),

// handling the reply
sdk.NewEvent(
    "reply",
    sdk.NewAttribute("_contract_addr", contractAddr.String()),
),
sdk.NewEvent(
    "wasm",
    sdk.NewAttribute("_contract_addr", contractAddr.String()),
    sdk.NewAttribute("custom", "from contract"),
),
// note this is emitted after handling the reply
sdk.NewEvent(
    "submsg_stop",
),

// calling the distribution module
sdk.NewEvent(
    "submsg_start",
),
sdk.NewEvent(
    "set_withdraw_address",
    sdk.NewAttribute("withdraw_address", withdrawAddr.String()),
),
sdk.NewEvent(
    "submsg_stop",
),
```

What do you think?

ethanfrey (commented):
What do you think of this proposal?

If you like it, I can flesh it out with a more complex example, where we have a multi-level tree and both instantiate and reply dispatch messages to demonstrate the ordering here.

EVENTS.md
alpe: IMHO to emit the events in order of processing is better than what we do now. It should not matter but for debugging and flattening, I can see the benefit.

EVENTS.md
alpe: Let's not add events for tracing purpose, only. reply/ start/ stop do not add value for clients IMHO

EVENTS.md
ethanfrey: 90% of contracts just want to emit attributes.
Only a few emit multiple events.
We didn't want to make it that much more complex for all of them, which is why we didn't remove the attributes when we introduced events. 

But you have a point. And with the new builder API from @uint we could even let people just `add_attribute` still and just expose them as a `wasm-default` event (so we keep the same cosmwasm-std api, but change the wasm interface between contract and runtime).

That is a larger discussion, and 0.16 is already cut. I feel it might be too late to make this change, but we could consider it for 1.0.0-beta.

That said, I would document both paths for now as we will support both in 0.18. If we change that for 1.0.0-beta, we can update the doc as well.

EVENTS.md
ethanfrey: I take your point on start and stop.

I am not sure if reply doesn't make sense. I guess it is only a separator to clarify if the calling contract did something or the subcontract did it, which may not be clear with some of the staking messages. I will think more on it.

ethanfrey:
@alpe I revised this and I think addressed you concerns in https://github.com/CosmWasm/wasmd/pull/581/commits/be3fcf1b73588992a0c131018eb532a3a4f95b9b

Note that I would strongly like to mark every CosmWasm contract entry point called, as this can be very useful to understand a contract.

I also remove the (potentially very large) result attribute from the events https://github.com/CosmWasm/wasmd/pull/581/commits/79f3fadf45780b55636d303c79eaf726ab9e2dca as I see no need to return them now that we have better handling of the data response with submessages and can return a chosen value to the top level

EVENTS.md
```diff
@@ -0,0 +1,404 @@
...
+    "unpin_code",
+    sdk.NewAttribute("code_id", strconv.FormatUint(msg.CodeID, 10)),
+)
+
+// Emitted when processing a submessage reply
+sdk.NewEvent(
+    "reply",
+    sdk.NewAttribute("_contract_addr", contractAddr.String()),
+    sdk.NewAttribute("success", strconv.FormatBool(err == nil)),
```
alpe, line 215: Can you elaborate what "success" means in this context? When the reply handling fails with an error we revert the TX. 
To trace the replay call, we could store the submessage reply condition as an attribute: Always, Success, Error.

alpe (commented):
Nice updates! 🌻  Two points left where we had disagreed:
- `Reply` : I was focused on business app working with events but I could imagine that there is space for technical ones as well to expose internals like block explorers or IBC network graphs do. 🤷  So let's find some good attributes and do this
- `wasmvmtypes.EventAttribute` : I do understand the value for Rust devs to create events easily but in the overall design there should not be 2 ways to create an event. Custom events are more powerful and should be preferred instead. Let's remove EventAttributes from the response 🙏  For the contract developers we could have a nice function to initialize a custom event with some defaults instead

EVENTS.md
ethanfrey: This should be renamed to give more clarity.

When reply is called, it receives the result of a sub message processing, which is either "Ok" or "Err".

If I have RelyOn::Always, is is useful to see if I am processing a success call, or handling an error callback.
The main use-case is for devs having some visibility.

Do you have an idea for a better name?

ethanfrey: How about `("submsg", "ok")` and `("submsg", "err")` as the two options here? To show if we are processing a success case or handling an error.

Better suggestions welcome

alpe: how about: "mode": "handle_success" || "handle_failure"

alpe (approved):

PR #583, opened 2021-08-05
Some examples/comments

alpe:
Do not merge 🚨 

Some examples from SDK and comments.

I would like to see examples of flattened events. I am worried now this may open the door for fraud. 🤷

### DIFF
--- a/EVENTS.md
+++ b/EVENTS.md
@@ -1,0 +1,69 @@
+# Event System
+
+## Usage in the SDK
+
+Events are an essential part of the Cosmos SDK. They are similar to "logs" in Ethereum and allow a blockchain
+app to attach key-value pairs to a transaction that can later be used to search for it or extract some information
+in human readable form. Events are not written to the application state, nor do they form part of the AppHash,
+but mainly intended for client use (and become an essential API for any reactive app or app that searches for txs). 
+
+In contrast, transactions also have a binary "data" field that is part of the AppHash (provable with light client proofs,
+part of consensus). This data is not searchable, but given a tx hash, you can be gauranteed what the data returned is.
+This is often empty, but sometimes custom protobuf formats to return essential information from an execution.
+
+Every message in the SDK may add events to the EventManager and these are then added to the final ABCI result that is returned
+to Tendermint. Events are exposed in 3 different ways over the Tendermint API (which is the only way a client can query).
+First of all is the `events` field on the transaction result (when you query a transaction by hash, you can see all event emitted
+by it). Secondly is the `log` field on the same transaction result. And third is the query interface to search or subscribe for
+transactions. 
+
+The `log` field actually has the best data. It contains an array of array of events. The first array is one entry per incoming message.
+Transactions in the Cosmos SDK may consist of multiple messages that are executed atomically. Maybe we send tokens, then issue a swap
+on a DEX. Each action would return it's own list of Events and in the logs, these are separated. For each message, it maintains a list
+of Events, exactly in the order returned by the application. This is JSON encoded and can be parsed by a client.
+
+In Tendermint 0.35, the `events` field will be one flattened list of events over all messages. Just appending the lists returned
+from each message. However, currently (until Tendermint 0.34 used in Cosmos SDK 0.40-0.43), they are flattened on type. Meaning all events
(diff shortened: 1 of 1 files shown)


ethanfrey:
> I would like to see examples of flattened events. I am worried now this may open the door for fraud. 🤷

They are only available via Tendermint (Cosmos SDK does not flatten).

If you have:

```json
[
  { "type": "wasm", "attributes": [
    {"key": "_contract_address", "value": "12345"}, 
    {"key": "transfer", "value": "failed"}]},
  { "type": "wasm", "attributes": [
    {"key": "_contract_address", "value": "76543"}, 
    {"key": "transfer", "value": "success"}]}
]
```

This will end up in the Tendermint queries as:

```json
[
  { "type": "wasm", "attributes": [
    {"key": "_contract_address", "value": "12345"}, 
    {"key": "transfer", "value": "failed"}]},
    {"key": "_contract_address", "value": "76543"}, 
    {"key": "transfer", "value": "success"}]}
]
```

If we tag each on with the _contract_address, the info is there and the onus is on the client to carefully parse. (Not just contract address is there and transfer=success is there, but the order.

That said, CosmJS has long considered this field unreliable and only returns events from parsing the log field, which has all proper data.

EVENTS.md
```diff
@@ -35,21 +35,85 @@ TODO: show event structure.
...
+```
+While the custom event type is emitted in the [`keeper.go`](https://github.com/cosmos/cosmos-sdk/blob/v0.42.9/x/distribution/keeper/keeper.go#L74-L77):
+```go
+sdk.NewEvent(
+    types.EventTypeSetWithdrawAddress,
+    sdk.NewAttribute(types.AttributeKeyWithdrawAddress, withdrawAddr.String()),
+),
+```
 
-TODO: what is added by the AnteHandlers (message.signer? auth?)
```
ethanfrey, line 40: This is emitted on every transaction - https://github.com/cosmos/cosmos-sdk/blob/v0.42.9/x/auth/ante/sigverify.go#L103-L120 not so interesting for us but good to reference.

Also, it would make reading easier IMO if `sdk.AttributeKeyModule` was replaced by "module" or whatever. Anyone using events will need to find the strings for the client.

EVENTS.md
```diff
@@ -35,21 +35,85 @@ TODO: show event structure.
...
+sdk.NewEvent(
+    types.EventTypeCreateValidator,
+    sdk.NewAttribute(types.AttributeKeyValidator, msg.ValidatorAddress),
+    sdk.NewAttribute(sdk.AttributeKeyAmount, msg.Value.Amount.String()),
+)
+```
+
+They also emit one or more other custom type event in the module that describe the operation executed with more details or metadata.
+```go
+sdk.NewEvent(
```
ethanfrey, line 51: This is the same as above? Did you intend to link something else?

EVENTS.md
ethanfrey: And THIS is what I was thinking of when I mentioned AnteHandlers: https://github.com/cosmos/cosmos-sdk/blob/6888de1d86026c25197c1227dae3d7da4d41a441/baseapp/baseapp.go#L746-L748

```go
msgEvents := sdk.Events{
			sdk.NewEvent(sdk.EventTypeMessage, sdk.NewAttribute(sdk.AttributeKeyAction, eventMsgName)),
		}
		msgEvents = msgEvents.AppendEvents(msgResult.GetEvents())
```

Unfortunately the calculation of that string changed between 0.42.9 and 0.43.0 (for new message service modules):

0.42.9: `msgFqName = svcMsg.MethodName` https://github.com/cosmos/cosmos-sdk/blob/v0.42.9/baseapp/baseapp.go#L709

0.43.0: `eventMsgName = sdk.MsgTypeURL(msg)` https://github.com/cosmos/cosmos-sdk/blob/v0.43.0-rc3/baseapp/baseapp.go#L720

It uses a longer URL. This did break relayers

ethanfrey (commented):
Some response to your comments. Links to other parts of the sdk

ethanfrey:
Thanks for the input. Integrated feedback into #581 manually

EVENTS.md
alpe: ja, first example should have been the message type. sorry

PR #585, opened 2021-08-09
Use shellcheck and shfmt on contrib scripts

husio:
This commit does not introduce any functionality change. It is only
applying suggestions from the shellcheck linter and reformats the code
using shfmt.

Shellcheck is a base shell linter that helps avoiding common shell
script mistakes.
Shfmt formats shell code similar to other Xfmt (i.e. gofmt) tools. No
need to format code manually or argue about the style anymore!

Both tools can be integrated with many editors or IDE.


husio:
Starting with a useless commit :see_no_evil:

alpe:
Nice, thanks 💐

alpe (approved):

PR #586, opened 2021-08-10
Refactor events

### LINKED ISSUE #440: Revisit event types

alpe:
Resolves #440 

Implements [EVENTS.md](./EVENTS.md)
Contains:
* Filter out `bank` module event of type `message` when tokens are transferred to a contract
* New event type `gov_contract_result` for contract `execute` and `migrate` in the gov proposal handler

Left to do:
- [x] test new events in keeper
- [x] fix `TestIBCReflectContract` by a new contract version that reads from the right event
- [x] order of events


ethanfrey:
> fix TestIBCReflectContract by a new contract version that reads from the right event

Uh oh... is this broken in cosmwasm-plus??? and I just tagged. I will fix that

EVENTS.md
```diff
@@ -402,6 +402,17 @@ If the original contract execution example above was actually the result of a me
 and it registered a ReplyOn clause, the `reply` function on that contract would receive the entire 11 events in the example
 above, and would need to use the `message` markers to locate the segment of interest.
 
+## Governance Events
```
ethanfrey, line 405: nice addition

x/wasm/keeper/keeper.go
```diff
@@ -178,6 +180,16 @@ func (k Keeper) create(ctx sdk.Context, creator sdk.AccAddress, wasmCode []byte,
 	}
 	codeInfo := types.NewCodeInfo(checksum, creator, *instantiateAccess)
 	k.storeCodeInfo(ctx, codeID, codeInfo)
+
+	evt := sdk.NewEvent(
```
ethanfrey, line 184: lgtm

x/wasm/keeper/keeper.go
```diff
@@ -299,6 +311,11 @@ func (k Keeper) instantiate(ctx sdk.Context, codeID uint64, creator, admin sdk.A
 		return nil, nil, sdkerrors.Wrap(err, "dispatch")
 	}
 
+	ctx.EventManager().EmitEvent(sdk.NewEvent(
+		types.EventTypeInstantiate,
+		sdk.NewAttribute(types.AttributeKeyContractAddr, contractAddress.String()),
+		sdk.NewAttribute(types.AttributeKeyCodeID, fmt.Sprintf("%d", codeID)),
```
ethanfrey, line 317: What about standarizing on `strconv.FormatUint(codeID, 10)`? (Or on this)

Store code does a different conversion, consistency is nice.

x/wasm/keeper/keeper.go
```diff
@@ -337,6 +354,11 @@ func (k Keeper) execute(ctx sdk.Context, contractAddress sdk.AccAddress, caller
 	if err != nil {
 		return nil, sdkerrors.Wrap(err, "dispatch")
 	}
+
+	ctx.EventManager().EmitEvent(sdk.NewEvent(
```
ethanfrey, line 358: nice, but same point about ordering. always emit these events *before* the `k.wasmvm.xxx()` calls. All info is available there and this makes sense to prefix the actual wasm messages with the keeper trigger.

that is also how the order is defined in the spec.

x/wasm/keeper/keeper.go
```diff
@@ -401,6 +423,12 @@ func (k Keeper) migrate(ctx sdk.Context, contractAddress sdk.AccAddress, caller
...
+		types.EventTypeMigrate,
+		sdk.NewAttribute(types.AttributeKeyCodeID, fmt.Sprintf("%d", newCodeID)),
```
ethanfrey, line 429: nice, but same point on codeId formatting

x/wasm/keeper/keeper.go
```diff
@@ -433,6 +461,11 @@ func (k Keeper) Sudo(ctx sdk.Context, contractAddress sdk.AccAddress, msg []byte
 	if err != nil {
 		return nil, sdkerrors.Wrap(err, "dispatch")
 	}
+
+	ctx.EventManager().EmitEvent(sdk.NewEvent(
```
ethanfrey, line 465: thanks to add this, but same ordering point as everywhere else

x/wasm/keeper/keeper.go
```diff
@@ -466,6 +499,11 @@ func (k Keeper) reply(ctx sdk.Context, contractAddress sdk.AccAddress, reply was
...
```
ethanfrey, line 503: nice

x/wasm/keeper/keeper.go
```diff
@@ -775,7 +823,7 @@ func (k *Keeper) handleContractResponse(
 	attributeGasCost := k.gasRegister.EventCosts(attrs, evts)
 	ctx.GasMeter().ConsumeGas(attributeGasCost, "Custom contract event attributes")
 	// emit all events from this contract itself
-	if len(attrs) != 0 || !hasWasmModuleEvent(ctx, contractAddr) {
+	if len(attrs) != 0 {
```
ethanfrey, line 826: :+1:

x/wasm/keeper/keeper.go
```diff
@@ -996,6 +1046,12 @@ func (c BankCoinTransferrer) TransferCoins(ctx sdk.Context, fromAddr sdk.AccAddr
 	if sdkerr != nil {
 		return sdkerr
 	}
+	for _, e := range em.Events() {
```
ethanfrey, line 1049: Nice way to handle this filtering in one place

x/wasm/keeper/keeper_test.go
```diff
@@ -1042,9 +1042,16 @@ func TestMigrateWithDispatchedMessage(t *testing.T) {
 				{"module": "bank"},
 			},
 		},
+		{
```
ethanfrey, line 1045: The events look mostly right, but the ordering all wrong.

I would expect 

1. no `message` events (they should be dropped inside the keeper, the module: wasm one added in msg server
2. first event is migrate
3. then wasm
4. then transfer

basically, placing migrate on top and filtering out the messages and this looks better to me. let me look at the keeper code again

x/wasm/keeper/proposal_handler.go
```diff
@@ -58,18 +56,8 @@ func handleStoreCodeProposal(ctx sdk.Context, k types.ContractOpsKeeper, p types
...
-	}
-
-	ourEvent := sdk.NewEvent(
-		sdk.EventTypeMessage,
-		sdk.NewAttribute(sdk.AttributeKeyModule, types.ModuleName),
-		sdk.NewAttribute(types.AttributeKeyCodeID, fmt.Sprintf("%d", codeID)),
-	)
-	ctx.EventManager().EmitEvent(ourEvent)
-	return nil
+	_, err = k.Create(ctx, runAsAddr, p.WASMByteCode, p.InstantiatePermission)
```
ethanfrey, line 59: good to remove message.
no gov-result event here? (not that I see the need for it)

x/wasm/keeper/proposal_handler.go
```diff
@@ -85,19 +73,15 @@ func handleInstantiateProposal(ctx sdk.Context, k types.ContractOpsKeeper, p typ
...
 
-	ourEvent := sdk.NewEvent(
-		sdk.EventTypeMessage,
-		sdk.NewAttribute(sdk.AttributeKeyModule, types.ModuleName),
-		sdk.NewAttribute(types.AttributeKeyCodeID, fmt.Sprintf("%d", p.CodeID)),
-		sdk.NewAttribute(types.AttributeKeyContractAddr, contractAddr.String()),
-		sdk.NewAttribute(types.AttributeResultDataHex, hex.EncodeToString(data)),
-	)
-	ctx.EventManager().EmitEvent(ourEvent)
+	ctx.EventManager().EmitEvent(sdk.NewEvent(
```
ethanfrey, line 81: looks good

x/wasm/keeper/proposal_integration_test.go
```diff
@@ -111,8 +111,12 @@ func TestInstantiateProposal(t *testing.T) {
 	}}
 	assert.Equal(t, expHistory, wasmKeeper.GetContractHistory(ctx, contractAddr))
 	// and event
-	require.Len(t, em.Events(), 2, "%#v", em.Events())
-	require.Len(t, em.Events()[1].Attributes, 4)
+	require.Len(t, em.Events(), 3, "%#v", em.Events())
+	require.Equal(t, types.WasmModuleEventType, em.Events()[0].Type)
```
ethanfrey, line 115: This is the `type: wasm` event?

Again, I figure this should be after instantiate

x/wasm/keeper/keeper.go
```diff
@@ -299,6 +311,11 @@ func (k Keeper) instantiate(ctx sdk.Context, codeID uint64, creator, admin sdk.A
 		return nil, nil, sdkerrors.Wrap(err, "dispatch")
 	}
 
+	ctx.EventManager().EmitEvent(sdk.NewEvent(
```
ethanfrey, line 314: Please emit this *before* calling `k.wasmVM.Instantiate`.

We have all needed info there, and this is the ordering from the spec.

ethanfrey (commented):
Nice start.

Most of my comments are about ordering.
The other issue is it seems the message events aren't all stripped (or only stripped at the msg_server level). We need to strip them inside the keeper, especially before passing the events into `reply`, so we don't expose it.

I just saw one test case below with 2 message type events and was quite surprised these were not already removed.

ethanfrey:
I fixed the ibc reflect contract in https://github.com/CosmWasm/cosmwasm/pull/1046 and copied it here so the test now works.

With my review comments above, this is ready to merge in my eyes.

x/wasm/keeper/keeper.go
alpe: The order of events was intentionally not addressed as I wanted minimal impact on tests. I should have added this to the todos in the description. sorry.

x/wasm/keeper/proposal_handler.go
alpe: The event is only emitted when we interact with a contract as the result data would not be available otherwise.

x/wasm/keeper/keeper.go
```diff
@@ -418,17 +418,17 @@ func (k Keeper) migrate(ctx sdk.Context, contractAddress sdk.AccAddress, caller
 	k.addToContractCodeSecondaryIndex(ctx, contractAddress, historyEntry)
 	k.storeContractInfo(ctx, contractAddress, contractInfo)
 
-	// dispatch submessages then messages
+	ctx.EventManager().EmitEvent(sdk.NewEvent(
```
ethanfrey, line 421: Okay, good enough location to move it.

x/wasm/keeper/keeper.go
```diff
@@ -1072,12 +1072,19 @@ func NewDefaultWasmVMContractResponseHandler(md msgDispatcher) *DefaultWasmVMCon
...
+	em := sdk.NewEventManager()
 	result := origRspData
-	switch rsp, err := h.md.DispatchSubmessages(ctx, contractAddr, ibcPort, messages); {
+	switch rsp, err := h.md.DispatchSubmessages(ctx.WithEventManager(em), contractAddr, ibcPort, messages); {
 	case err != nil:
 		return nil, sdkerrors.Wrap(err, "submessages")
 	case rsp != nil:
 		result = rsp
 	}
+	// emit non message type events only
```
ethanfrey, line 1083: This is good to clear, but in the spec, we need to clear the messages *before* we possibly place them inside reply.

I believe you need to do that inside DispatchSubmessages

x/wasm/keeper/keeper_test.go
```diff
@@ -301,6 +306,15 @@ func TestInstantiate(t *testing.T) {
 		Msg:       json.RawMessage(initMsgBz),
 	}}
 	assert.Equal(t, exp, keepers.WasmKeeper.GetContractHistory(ctx, gotContractAddr))
+
+	// and events emitted
+	expEvt := sdk.Events{
```
ethanfrey, line 311: looks good and matching spec

x/wasm/keeper/keeper_test.go
```diff
@@ -530,6 +545,12 @@ func TestExecute(t *testing.T) {
 	require.NotNil(t, contractAcct)
 	assert.Equal(t, sdk.Coins(nil), bankKeeper.GetAllBalances(ctx, contractAcct.GetAddress()))
 
+	// and events emitted
+	require.Len(t, em.Events(), 5)
+	expEvt := sdk.NewEvent("execute",
```
ethanfrey, line 550: event 0 is the message and the other ones?
I find it nice to test all the types at least, makes it clear what is (expected to) happen

x/wasm/keeper/keeper_test.go
```diff
@@ -1014,6 +1035,13 @@ func TestMigrateWithDispatchedMessage(t *testing.T) {
 	assert.Equal(t, "burnt 1 keys", string(data))
 	type dict map[string]interface{}
 	expEvents := []dict{
+		{
```
ethanfrey, line 1038: :+1:

x/wasm/keeper/keeper_test.go
```diff
@@ -1491,15 +1590,60 @@ func TestNewDefaultWasmVMContractResponseHandler(t *testing.T) {
...
 
-			gotData, gotErr := d.Handle(sdk.Context{}, RandomAccountAddress(t), "ibc-port", msgs, spec.srcData)
+	specs := map[string]struct {
+		rsp     wasmvmtypes.Response
+		expData []byte
+		expErr  bool
+		expEvt  sdk.Events
+	}{
+		"all good": {
+			rsp:     wasmvmtypes.Response{Data: []byte("foo")},
```
ethanfrey, line 1622: I would add a case where reply returns an attribute or event and ensure those get returned as well (I am sure they are, but if you want to write table tests to cover the use cases...)

x/wasm/keeper/msg_server.go
```diff
@@ -53,17 +54,17 @@ func (m msgServer) InstantiateContract(goCtx context.Context, msg *types.MsgInst
...
-		return nil, err
-	}
-
 	ctx.EventManager().EmitEvent(sdk.NewEvent(
 		sdk.EventTypeMessage,
 		sdk.NewAttribute(sdk.AttributeKeyModule, types.ModuleName),
 		sdk.NewAttribute(sdk.AttributeKeySender, msg.Sender),
 	))
 
+	contractAddr, data, err := m.keeper.Instantiate(ctx, msg.CodeID, senderAddr, adminAddr, msg.Msg, msg.Label, msg.Funds)
```
ethanfrey, line 63: Good. Message in front of everything :+1:

x/wasm/module_test.go
```diff
@@ -238,12 +238,12 @@ func TestHandleExecute(t *testing.T) {
 	require.Equal(t, "cosmos14hj2tavq8fpesdwxxcu44rty3hh90vhuc53mp6", contractBech32Addr)
 	// this should be standard x/wasm message event,  init event, plus a bank send event (2), with no custom contract events
 	require.Equal(t, 4, len(res.Events), prettyEvents(res.Events))
-	require.Equal(t, "transfer", res.Events[0].Type)
-	require.Equal(t, "wasm", res.Events[1].Type)
-	assertAttribute(t, "_contract_address", contractBech32Addr, res.Events[1].Attributes[0])
+	require.Equal(t, "message", res.Events[0].Type)
```
ethanfrey, line 241: This is very nice and clear...

message (top-level), transfer (pay fees sent with message), instantiate + wasm (actual contract execution)

x/wasm/keeper/keeper.go
ethanfrey: TODO: I will raise this as a new issue (and to add tests if it works or not)

ethanfrey (approved):
I think this is good to merge. And we can do a follow-up PR to look into any little issues left.

I think it is 98% correct, just one issue about message event and reply

x/wasm/keeper/keeper.go
ethanfrey: See #587

PR #588, opened 2021-08-12
Start filter out message events

### LINKED ISSUE #587: Ensure proper events sent in reply
In EVENTS.md we define not only the events emitted by the chain but the events that are passed into reply blocks. 

Note: you can use the `reflect` contract to test this, it will just store what it gets in the `reply` block into it's local storage to be queried later.

This is follow up from https://github.com/CosmWasm/wasmd/pull/586#discussion_r686823220

In #586, we filter out all `message` type events after [dispatching all submessages](https://github.com/CosmWasm/wasmd/blob/7b2cfa48ed7edcbda6df58e5687bee2976d6cf59/x/wasm/keeper/keeper.go#L1073-L1090) However, I would argue that we must filter them *before* returning the events to the `reply` block, which is what is in the spec.

Looking into the internals (and assuming the default implementation... all the interfaces make flow analysis a bit tricky), we wee how we [already make a temporary event manager](https://github.com/CosmWasm/wasmd/blob/b0b1fa88bc6875b04b47063504e7810f955a8355/x/wasm/keeper/msg_dispatcher.go#L83-L99) before dispatching the message. 

When then only [re-emit them if the event succeeds](https://github.com/CosmWasm/wasmd/blob/b0b1fa88bc6875b04b47063504e7810f955a8355/x/wasm/keeper/msg_dispatcher.go#L101-L106), which avoids errors from poluting the event stack. Interestingly, there seem to be two sources of events and I am unclear what goes where:

```
if err == nil {
	commit()
	ctx.EventManager().EmitEvents(em.Events())
	ctx.EventManager().EmitEvents(events)
} // on failure, revert state from sandbox, and ignore events (just skip doing the above)
```

Which events are in `events` and which in `em.Events()`. Can we drop `em.Events()`?

Later on we see that only the [`events` are passed into the reply block](https://github.com/CosmWasm/wasmd/blob/b0b1fa88bc6875b04b47063504e7810f955a8355/x/wasm/keeper/msg_dispatcher.go#L125-L130) on success.

We should clear this up. I would propose the following:

* Verify what is coming from the two event sources (`events` and `em.Events()`)
* Combine them (or drop one of them) and filter out `message` type events to produce `filteredEvents`.
* `ctx.EventManager().EmitEvents(filteredEvents)` instead of these two calls
* pass in ` filteredEvents` to the reply call

This will ensure no messages make it to reply. There are never dups (between the two event sources) and it removes the need to [filter later in Handle](https://github.com/CosmWasm/wasmd/blob/7b2cfa48ed7edcbda6df58e5687bee2976d6cf59/x/wasm/keeper/keeper.go#L1073-L1090)

alpe:
🚧  WIP
Resolves #587

### DIFF
--- a/x/wasm/keeper/events.go
+++ b/x/wasm/keeper/events.go
@@ -63,3 +63,13 @@ func contractSDKEventAttributes(customAttributes []wasmvmtypes.EventAttribute, c
 	}
 	return attrs, nil
 }
+
+func filterOutMessageTypeEvents(events sdk.Events) sdk.Events {
+	var r sdk.Events
+	for _, e := range events {
+		if e.Type != sdk.EventTypeMessage {
+			r = append(r, e)
+		}
+	}
+	return r
+}
--- a/x/wasm/keeper/handler_plugin.go
+++ b/x/wasm/keeper/handler_plugin.go
@@ -150,6 +150,8 @@ func (h IBCRawPacketHandler) DispatchMsg(ctx sdk.Context, _ sdk.AccAddress, cont
 		return nil, nil, sdkerrors.Wrapf(types.ErrEmpty, "ibc channel")
 	}
 
+	em := sdk.NewEventManager()
+	ctx = ctx.WithEventManager(em)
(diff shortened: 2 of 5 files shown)


alpe:
obsolete

PR #591, opened 2021-08-13
Follow up on duplicate event emition

alpe:
Follow up on #440

Events were returned as value of the message processing call but also emitted via context event manager when talking directly to sdk keepers. With this PR context events are captured and returned for our handler plugins.

An open question for is what to do with custom handlers that are not updated?
Before we appended context events to the returned ones. With this PR they are **silently discarded** (see https://github.com/CosmWasm/wasmd/compare/polish_event_emitting?expand=1#diff-171d7d96b1521ea0c90ca4c1d112065104dab4d6c1219f84f60a5c2dc3d28d6dR85). 
Another option would be to be strict and fail the TX.

### DIFF
--- a/x/wasm/keeper/msg_dispatcher.go
+++ b/x/wasm/keeper/msg_dispatcher.go
@@ -82,28 +82,28 @@ func (d MessageDispatcher) DispatchSubmessages(ctx sdk.Context, contractAddr sdk
 		}
 		// first, we build a sub-context which we can use inside the submessages
 		subCtx, commit := ctx.CacheContext()
-		em := sdk.NewEventManager()
-		subCtx = subCtx.WithEventManager(em)
+		subCtx = subCtx.WithEventManager(sdk.NewEventManager()) // discard any context events
 
 		// check how much gas left locally, optionally wrap the gas meter
 		gasRemaining := ctx.GasMeter().Limit() - ctx.GasMeter().GasConsumed()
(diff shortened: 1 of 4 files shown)


ethanfrey:
Interesting approach. Let's look at this after cutting 0.18

x/wasm/keeper/msg_dispatcher.go
```diff
@@ -82,28 +82,28 @@ func (d MessageDispatcher) DispatchSubmessages(ctx sdk.Context, contractAddr sdk
...
+			data   [][]byte
+			err    error
+		)
 		if limitGas {
 			events, data, err = d.dispatchMsgWithGasLimit(subCtx, contractAddr, ibcPort, msg.Msg, *msg.GasLimit)
 		} else {
 			events, data, err = d.messenger.DispatchMsg(subCtx, contractAddr, ibcPort, msg.Msg)
 		}
 
+		events = filterEvents(events)
```
ethanfrey, line 102: I see how the usage gets simpler with only one path to handle.

But this puts a lot more work on the people writing handlers.
Since there is much less chance someone forks this dispatcher, than someone adds a new handler (for custom messages on any chain), I think this actually makes the code a bit more fragile, and some events may get silently dropped with hard-to-debug issues.

ethanfrey (commented):
Upon reflection, I think this adds complexity with little value for code maintainability.
Let's just keep it how it is.

PR #598, opened 2021-08-30
Add decorator options

alpe:
I played today with the query and message handler extension points and they are pretty hard to extend.
They likely require a keeper instance. With the decorator keeper options this can be simplified.
I have modified one of our tests to show the old vs new setup.

### DIFF
--- a/x/wasm/keeper/options.go
+++ b/x/wasm/keeper/options.go
@@ -21,23 +21,39 @@ func WithWasmEngine(x types.WasmerEngine) Option {
 }
 
 // WithMessageHandler is an optional constructor parameter to set a custom handler for wasmVM messages.
-// This option should not be combined with Option `WithMessageEncoders`.
+// This option should not be combined with Option `WithMessageEncoders` or `WithMessageHandlerDecorator`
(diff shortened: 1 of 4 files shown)


x/wasm/keeper/options.go
```diff
@@ -21,23 +21,39 @@ func WithWasmEngine(x types.WasmerEngine) Option {
 }
 
 // WithMessageHandler is an optional constructor parameter to set a custom handler for wasmVM messages.
-// This option should not be combined with Option `WithMessageEncoders`.
+// This option should not be combined with Option `WithMessageEncoders` or `WithMessageHandlerDecorator`
```
husio, line 24: Go documentation does not support any markup language. I think wrapping code in back quotes (`) is not very common, since it does not do anything. 

Since each public token comment starts with that token name (here `WithMessageHandler`), it is inconsistent that the first occurrence is not surrounded by back quotes and further code mentions are.

:nail_care:

husio (approved):
Looks good! :sandwich: 

Please keep in mind that I don't get all the design implications yet.

PR #605, opened 2021-09-13
Bump wasmvm to v0.16.1

alpe:
Patch version upgrade


ethanfrey (approved):
Nice. An improved debugging experience is always welcome.

Issue #601, opened 2021-09-08
Add transaction index implemented as counter


webmaster128:
We'd like to expose the index of a transaction in a block in https://github.com/CosmWasm/cosmwasm/pull/1077. This allows contracts to get a pair `(block height, transaction index)` that uniquely identifies the transaction on a given chain. This transaction index should be easy to work with in other tools, e.g. by looking in a block explorer or using a call like

```
curl -sS "https://rpc.cosmos-hub.app.beta.starport.cloud/block?height=7590762" | jq .result.block.data.txs
```

In https://github.com/CosmWasm/wasmd/blob/v0.18.0/proto/cosmwasm/wasm/v1/types.proto#L123-L131 the gas meter is used for ordering, but the values from this are hard to work with. So what we look for is an 0-based index 0, 1, 2, ….

ethanfrey:
Good request. And made me curious was tx[5] is?  It is about 23KB after base64 decoding...

A pdf embedded in a gov proposal?

PR #608, opened 2021-09-13
TX counter spike

alpe:
🚧  SPIKE: do not merge

See #601

In this spike I have added an ante handler that counts the TX in a block and passes this down to the wasmvm Env.

* The counter is persisted in the KV store as check/deliver/simulate processes handle their DB state accordingly
* Simulations and external queries run with TX count `0`. 

Not included:
* How the TX counter is passed to the contracts
* Tests

Does this work for you @reuvenpo @assafmo ?

### DIFF
--- a/app/ante.go
+++ b/app/ante.go
@@ -1,0 +1,38 @@
+package app
+
+import (
+	wasmkeeper "github.com/CosmWasm/wasmd/x/wasm/keeper"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/x/auth/ante"
+	"github.com/cosmos/cosmos-sdk/x/auth/signing"
+	"github.com/cosmos/cosmos-sdk/x/auth/types"
+)
+
+// NewAnteHandler returns an AnteHandler that checks and increments sequence
+// numbers, checks signatures & account numbers, and deducts fees from the first
+// signer.
+func NewAnteHandler(
+	ak ante.AccountKeeper, bankKeeper types.BankKeeper,
+	sigGasConsumer ante.SignatureVerificationGasConsumer,
+	signModeHandler signing.SignModeHandler,
+	txCounterStoreKey sdk.StoreKey,
+) sdk.AnteHandler {
+	// copied sdk https://github.com/cosmos/cosmos-sdk/blob/v0.42.9/x/auth/ante/ante.go
+	return sdk.ChainAnteDecorators(
+		ante.NewSetUpContextDecorator(), // outermost AnteDecorator. SetUpContext must be called first
+		ante.NewRejectExtensionOptionsDecorator(),
+		ante.NewMempoolFeeDecorator(),
+		ante.NewValidateBasicDecorator(),
+		ante.TxTimeoutHeightDecorator{},
+		ante.NewValidateMemoDecorator(ak),
+		ante.NewConsumeGasForTxSizeDecorator(ak),
+		ante.NewRejectFeeGranterDecorator(),
+		ante.NewSetPubKeyDecorator(ak), // SetPubKeyDecorator must be called before all signature verification decorators
+		ante.NewValidateSigCountDecorator(ak),
+		ante.NewDeductFeeDecorator(ak, bankKeeper),
+		ante.NewSigGasConsumeDecorator(ak, sigGasConsumer),
+		ante.NewSigVerificationDecorator(ak, signModeHandler),
+		ante.NewIncrementSequenceDecorator(ak),
+		wasmkeeper.NewCountTXDecorator(txCounterStoreKey),
+	)
+}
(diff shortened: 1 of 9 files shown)


assafmo:
I don't really understand why it's necessary for contracts to be aware of their in-block/in-tx position. 😅

x/wasm/types/types.go
```diff
@@ -256,6 +256,10 @@ func NewEnv(ctx sdk.Context, contractAddr sdk.AccAddress) wasmvmtypes.Env {
 	if nano < 1 {
 		panic("Block (unix) time must never be empty or negative ")
 	}
+	txCounter := TXCounter(ctx)
+	// todo: where to set the counter
+	_ = txCounter
```
alpe, line 261: https://github.com/CosmWasm/cosmwasm/pull/1077/files#diff-9818bb419db10c5127f7290971c04fdaa13ae8aa7495f9d900ee688e33d15d4aR25

assafmo:
> I don't really understand why it's necessary for contracts to be aware of their in-block/in-tx position.

To clarify my comment above, we talked in a private chat with @webmaster128, @ethanfrey & @reuvenpo about a way for contract calls within the same tx to have different `Env`s in order for them to be able to e.g. generate unique IDs. A few of the ideas were to have a "call counter" or an in-block gas position passed down to the contract as fields in `Env`.

I mistakenly thought that this PR is about a "call counter" and not a "tx counter", and that's what my above comment refers to. Sorry for the misunderstanding. :hearts:

x/wasm/types/types.go
ethanfrey: This will be available in 1.0.0-beta
(We can pre-release test with 1.0.0-soon, but let's put this merge on hold until we are ready to move wasmd support to that.. or start a branch for 1.0.0 compatibility)

ethanfrey:
Now unblocked with #617 
I agree with the design, let's get a final version wired up

app/ante.go
```diff
@@ -0,0 +1,38 @@
...
+		ante.NewValidateMemoDecorator(ak),
+		ante.NewConsumeGasForTxSizeDecorator(ak),
+		ante.NewRejectFeeGranterDecorator(),
+		ante.NewSetPubKeyDecorator(ak), // SetPubKeyDecorator must be called before all signature verification decorators
+		ante.NewValidateSigCountDecorator(ak),
+		ante.NewDeductFeeDecorator(ak, bankKeeper),
+		ante.NewSigGasConsumeDecorator(ak, sigGasConsumer),
+		ante.NewSigVerificationDecorator(ak, signModeHandler),
+		ante.NewIncrementSequenceDecorator(ak),
+		wasmkeeper.NewCountTXDecorator(txCounterStoreKey),
```
ethanfrey, line 36: good place to add

ethanfrey (commented):
A few comments to clean it up, but the approach is good.

x/wasm/keeper/ante.go
```diff
@@ -0,0 +1,40 @@
...
+			txCounter = val
+		}
+	}
+	store.Set(types.TXCounterPrefix, encodeHeightCounter(currentHeight, txCounter+1))
+
+	return next(types.WithTXCounter(ctx, txCounter), tx, simulate)
+}
+
+func encodeHeightCounter(height int64, counter uint64) []byte {
+	return append(sdk.Uint64ToBigEndian(uint64(height)), sdk.Uint64ToBigEndian(counter)...)
```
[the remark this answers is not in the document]
alpe: I would use a `uint32` type for the counter now but otherwise leave it as it is. We can always refactor as needed. Depending on the case we another ante handler may be a better option to keep things separated. 🤷

x/wasm/keeper/ante.go
```diff
@@ -0,0 +1,40 @@
...
+	return &CountTXDecorator{storeKey: storeKey}
+}
+
+func (a CountTXDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (sdk.Context, error) {
+	if simulate {
+		return next(types.WithTXCounter(ctx, 0), tx, simulate)
+	}
+	currentHeight := ctx.BlockHeight()
+	store := ctx.KVStore(a.storeKey)
+	var txCounter uint64 = 0
```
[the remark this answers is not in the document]
alpe: 🤔  I tried to follow the advice but there is also the case where the store is empty that requires a `0` init value. I am not sure if it would get easier by adding `else` blocks. Instead I have added more code comments.

webmaster128 (commented):
Nice, as far as I can tell.

Can we make the counter uint32 consistently? This is what I used on the contract side:

```rust
#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
pub struct TransactionInfo {
    /// The position of this transaction in the block. The first
    /// transaction has index 0.
    ///
    /// This allows you to get a unique transaction indentifier in this chain
    /// using the pair (`env.block.height`, `env.transaction.index`).
    ///
    pub index: u32,
}
```

alpe:
Closed in favour of #621

PR #609, opened 2021-09-13
Optimize BalanceQuery

### LINKED ISSUE #606: Optimize BalanceQuery in query plugin
We can use [GetBalance](https://github.com/cosmos/cosmos-sdk/blob/v0.42.9/x/bank/keeper/view.go#L26) now instead of iterating through the list

alpe:
Resolves #606


ethanfrey (approved):
yes, good catch

PR #611, opened 2021-09-14
Ensure query isolation

### LINKED ISSUE #610: Ensure Queries are executed read only
We could fail on any write operation or discard any changes as the SDK does.

alpe:
Resolves #610

### DIFF
--- a/x/wasm/keeper/keeper_test.go
+++ b/x/wasm/keeper/keeper_test.go
@@ -1652,6 +1652,37 @@ func TestReply(t *testing.T) {
 	}
 }
 
+func TestQueryIsolation(t *testing.T) {
+	ctx, keepers := CreateTestInput(t, false, SupportedFeatures)
+	k := keepers.WasmKeeper
+	var mock wasmtesting.MockWasmer
+	wasmtesting.MakeInstantiable(&mock)
+	example := SeedNewContractInstance(t, ctx, keepers, &mock)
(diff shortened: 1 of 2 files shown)


x/wasm/keeper/query_plugins.go
```diff
@@ -46,15 +46,15 @@ type GRPCQueryRouter interface {
 var _ wasmvmtypes.Querier = QueryHandler{}
 
 func (q QueryHandler) Query(request wasmvmtypes.QueryRequest, gasLimit uint64) ([]byte, error) {
-	// set a limit for a subctx
+	// set a limit for a subCtx
 	sdkGas := q.gasRegister.FromWasmVMGas(gasLimit)
-	subctx := q.Ctx.WithGasMeter(sdk.NewGasMeter(sdkGas))
+	subCtx, _ := q.Ctx.WithGasMeter(sdk.NewGasMeter(sdkGas)).CacheContext()
```
ethanfrey, line 51: This means all writes will be silently ignores, right?
That should be documented, but seems safe enough.

x/wasm/keeper/keeper_test.go
```diff
@@ -1652,6 +1652,37 @@ func TestReply(t *testing.T) {
 	}
 }
 
+func TestQueryIsolation(t *testing.T) {
```
ethanfrey, line 1655: Nice test

ethanfrey (approved):
Very nice, clean, minimal change providing an important safety guarantee.
I like it.

PR #612, opened 2021-09-14
Bump Go version to  1.16

### LINKED ISSUE #602: Bump Go version
Although there is 1.17.1 out already let's be more conservative and use Go 1.16.x with the latest security patches.

alpe:
Resolves #602

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/CosmWasm/wasmd
 
-go 1.15
+go 1.16
 
 require (
 	github.com/CosmWasm/wasmvm v0.16.1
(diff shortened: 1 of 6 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/CosmWasm/wasmd
 
-go 1.15
+go 1.16
```
ethanfrey, line 3: You cannot set go 1.16.8 here?

ethanfrey (approved):
Looks good.

go.mod
alpe: It is not possible to define the patch version here. I got: ` invalid go version '1.16.8': must match format 1.23`.
But I have added the `1.16.8` version to the readme

PR #617, opened 2021-09-22
Update to Cosmwasm v1.0.0-soon

ethanfrey:
This does not support the new features (TransactionInfo and ContractInfoQuery), but updates all dependencies so those can easily be added in follow-up PRs. (I think TransactionInfo in #608 is ready once it rebases on this)


alpe (approved):
👍  excellent

ethanfrey:
Merged. #608 is now unblocked

Issue #618, opened 2021-09-22
CosmWasm 1.0 tracking issue


ethanfrey:
Points needed:

- [x] https://github.com/CosmWasm/wasmd/pull/617
- [x] https://github.com/CosmWasm/wasmd/issues/601 / https://github.com/CosmWasm/wasmd/pull/608
- [x] https://github.com/CosmWasm/wasmd/issues/584
- [ ] Finally, upgrade to CosmWasm v1.0.0-beta
- [x] Increase gas multiplier by 150_000 to 15_000_000

@webmaster128 please add any other updates needed (pinning, gas, cache info??)

PR #621, opened 2021-09-27
Implement tx counter for transaction info

### LINKED ISSUE #601: Add transaction index implemented as counter

alpe:
Resolves #601 

Based on the #608 spike

Changes: 
- Use `uint32` for the tx counter
- Sets wasmvm TransactionInfo
- Non default `0` counter but empty TransactionInfo on simulations
- Tests
- Better doc

### DIFF
--- a/INTEGRATION.md
+++ b/INTEGRATION.md
@@ -63,6 +63,10 @@ from the Cosmos SDK, and enabled them in `app.go`. If so, you can just look
 at [`wasmd/app/app.go`](https://github.com/CosmWasm/wasmd/blob/master/app/app.go#)
 for how to do so (just search there for lines with `wasm`).
 
+`wasmd` also comes with a custom `ante handler` that adds the TX position in the block into the context
+and passes it to the contracts. In order to support this feature you would need to add our custom
+ante handler into the `ante handler chain` as in: [`app/ante.go`](https://github.com/CosmWasm/wasmd/blob/master/app/ante.go)
+
 ### Copied into your app
 
 Sometimes, however, you will need to copy `x/wasm` into your app. This should
--- a/app/ante.go
+++ b/app/ante.go
@@ -1,0 +1,38 @@
+package app
+
+import (
+	wasmkeeper "github.com/CosmWasm/wasmd/x/wasm/keeper"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/x/auth/ante"
+	"github.com/cosmos/cosmos-sdk/x/auth/signing"
+	"github.com/cosmos/cosmos-sdk/x/auth/types"
+)
+
+// NewAnteHandler returns an AnteHandler that checks and increments sequence
+// numbers, checks signatures & account numbers, and deducts fees from the first
+// signer.
+func NewAnteHandler(
+	ak ante.AccountKeeper, bankKeeper types.BankKeeper,
+	sigGasConsumer ante.SignatureVerificationGasConsumer,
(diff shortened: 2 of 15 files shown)


app/ante.go
```diff
@@ -0,0 +1,38 @@
...
+		ante.NewValidateMemoDecorator(ak),
+		ante.NewConsumeGasForTxSizeDecorator(ak),
+		ante.NewRejectFeeGranterDecorator(),
+		ante.NewSetPubKeyDecorator(ak), // SetPubKeyDecorator must be called before all signature verification decorators
+		ante.NewValidateSigCountDecorator(ak),
+		ante.NewDeductFeeDecorator(ak, bankKeeper),
+		ante.NewSigGasConsumeDecorator(ak, sigGasConsumer),
+		ante.NewSigVerificationDecorator(ak, signModeHandler),
+		ante.NewIncrementSequenceDecorator(ak),
+		wasmkeeper.NewCountTXDecorator(txCounterStoreKey),
```
ethanfrey, line 36: We should add some note that chains using x/wasm need to add this custom AnteHandler if they want Tx Index info.

Not sure where that goes? Integration.md?

ethanfrey (approved):
Nice clean version.

:+1:

ethanfrey:
Please make sure to tick the box in https://github.com/CosmWasm/wasmd/issues/618 after merging

INTEGRATION.md
```diff
@@ -63,6 +63,10 @@ from the Cosmos SDK, and enabled them in `app.go`. If so, you can just look
 at [`wasmd/app/app.go`](https://github.com/CosmWasm/wasmd/blob/master/app/app.go#)
 for how to do so (just search there for lines with `wasm`).
 
+`wasmd` also comes with a custom `ante handler` that adds the TX position in the block into the context
```
ethanfrey, line 66: Nice

contrib/local/02-contracts.sh
```diff
@@ -8,7 +8,7 @@ echo "## Add new CosmWasm contract"
 RESP=$(wasmd tx wasm store "$DIR/../../x/wasm/keeper/testdata/hackatom.wasm" \
   --from validator --gas 1500000 -y --chain-id=testing --node=http://localhost:26657 -b block)
 
-CODE_ID=$(echo "$RESP" | jq -r '.logs[0].events[0].attributes[-1].value')
+CODE_ID=$(echo "$RESP" | jq -r '.logs[0].events[1].attributes[-1].value')
```
ethanfrey, line 11: The ante handler added an event?
Or this is just an older bug you just found?

ethanfrey (approved):
lgtm

contrib/local/02-contracts.sh
alpe: This was a 🐛 that I found with manual tests. Not related to the ante handler.

PR #629, opened 2021-10-07
Reference standard IBC events

### LINKED ISSUE #593: Finalize IBC Events
These are not yet defined in EVENTS.md.
They were not critical to finalise for 0.18, as other cosmwasm contracts cannot listen to those (so they are not contract API breaking), but it is good to document and stabilise those before a 1.0 release

ethanfrey:
Closes #593


alpe (approved):
Nice summary!

PR #632, opened 2021-10-07
Improve comment on DefaultGasMultiplier

webmaster128:


alpe (approved):
Thanks for the detailed doc! 💐

PR #634, opened 2021-10-07
Benchmarks for gas pricing

ethanfrey:
Prepares #566 
Prepares #631 

- [x] Run Instantiation (pinned/unpinned) benchmarks in CI
- [x] Run secp256k1 verification benchmarks in CI (as reference for the 1 TerraGas/ms wasm definition)
- [x] Run compilation benchmarks in CI
- [x] Calculate new gas prices

I will update prices in a new PR, once we agree on the benchmarks.

PR #636, opened 2021-10-08
Update gas costs

### LINKED ISSUE #566: Adjust default gas settings
See the constants here. Many may warrant some adjustments. Especially the default gas multiplier.

https://github.com/CosmWasm/wasmd/blob/f6002fe94244ff2ac08e18f67ff9d19dc9b49fd7/x/wasm/keeper/gas_register.go#L11-L36

I would also revisit some of the algorithms, like how the pricing works on Reply, which is a bit different charge for events than other items.

### LINKED ISSUE #631: Check and refine DefaultGasMultiplier
There is CosmWasm gas and Cosmos SDK gas. With 1.0 CosmWasm gas strives to consume 1 Teragas/millisecond on an off-the-shelf Intel machine (e.g. CI machine). Anything significatly cheaper of more expensive (in the order of magnitudes) justifies repricing.

From some very early measurements we set DefaultGasMultiplier to 100 and increased it by a factor of 150_000 because between 0.16 and 1.0 CosmWasm was changed to charge 150_000 instead of 1 gas per Wasm operation. But we don't know accurate this initial 100 is after two years of development.

To refine the multiplier, we should figure out the amount of gas Cosmos SDK consumes per millisecond. This can be done via secp256k1 signature verification. We know it costs 1000 Cosmo SDK gas. We need to benchmark it and calculate the rest from there.

alpe:
resolves #566 
resolves #631 

Applies gas costs proposed in #634

### DIFF
--- a/x/wasm/keeper/gas_register.go
+++ b/x/wasm/keeper/gas_register.go
@@ -27,12 +27,12 @@ const (
 	// such as hardcoding them in contracts.
 	//
 	// Please note that all gas prices returned to wasmvm should have this multiplied.
-	DefaultGasMultiplier uint64 = 15_000_000
+	DefaultGasMultiplier uint64 = 140_000_000
 	// DefaultInstanceCost is how much SDK gas we charge each time we load a WASM instance.
 	// Creating a new instance is costly, and this helps put a recursion limit to contracts calling contracts.
-	DefaultInstanceCost uint64 = 40_000
+	DefaultInstanceCost uint64 = 60_000
 	// DefaultCompileCost is how much SDK gas is charged *per byte* for compiling WASM code.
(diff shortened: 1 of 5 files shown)


x/wasm/keeper/gas_register.go
```diff
@@ -27,12 +27,12 @@ const (
 	// such as hardcoding them in contracts.
 	//
 	// Please note that all gas prices returned to wasmvm should have this multiplied.
-	DefaultGasMultiplier uint64 = 15_000_000
+	DefaultGasMultiplier uint64 = 140_000_000
```
ethanfrey, line 30: Good changes.

I would add some text where this came from.
Minimally linking to the Benchmarking PR and the discussion.
(For these 3 items)

Issue #643, opened 2021-10-13
Fix error messages for LastCodeID in initGenesis


fkneeland-figure:
The current error message in the initGenesis function if the lastCodeID is incorrect fails to print out the current but incorrect value.  We should update the error statement to log all the fields to make debugging around this error easier in the future.

PR #644, opened 2021-10-13
Fixed wrong error message when lastCodeID value is incorrect

fkneeland-figure:
The current error message gives an incorrect log statement in the initGenesis function if the lastCodeID is incorrect and fails to print out the current but incorrect value. I updated the log statement as well as added all the fields to make debugging around this error easier in the future.

#643

### DIFF
--- a/x/wasm/keeper/genesis.go
+++ b/x/wasm/keeper/genesis.go
@@ -1,6 +1,7 @@
 package keeper
 
 import (
+	"fmt"
 	"github.com/CosmWasm/wasmd/x/wasm/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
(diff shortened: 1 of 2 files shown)


x/wasm/keeper/genesis.go
```diff
@@ -55,11 +55,13 @@ func InitGenesis(ctx sdk.Context, keeper *Keeper, data types.GenesisState, staki
 	}
 
 	// sanity check seq values
-	if keeper.peekAutoIncrementID(ctx, types.KeyLastCodeID) <= maxCodeID {
-		return nil, sdkerrors.Wrapf(types.ErrInvalid, "seq %s must be greater %d ", string(types.KeyLastCodeID), maxCodeID)
+	seqVal := keeper.peekAutoIncrementID(ctx, types.KeyLastCodeID)
+	if seqVal <= maxCodeID {
+		return nil, sdkerrors.Wrapf(types.ErrInvalid, "seq %s with value: %d must be less than or equal to maxCodeID: %d ", string(types.KeyLastCodeID), seqVal, maxCodeID)
```
alpe, line 60: The error message is not correct anymore. The sequence must be greater as defined in Line 59

x/wasm/keeper/genesis.go
```diff
@@ -55,11 +55,13 @@ func InitGenesis(ctx sdk.Context, keeper *Keeper, data types.GenesisState, staki
...
 	}
-	if keeper.peekAutoIncrementID(ctx, types.KeyLastInstanceID) <= uint64(maxContractID) {
-		return nil, sdkerrors.Wrapf(types.ErrInvalid, "seq %s must be greater %d ", string(types.KeyLastInstanceID), maxContractID)
+	seqVal = keeper.peekAutoIncrementID(ctx, types.KeyLastInstanceID)
+	if seqVal <= uint64(maxContractID) {
+		return nil, sdkerrors.Wrapf(types.ErrInvalid, "seq %s with value: %d must be less than or equal to maxContractID: %d ", string(types.KeyLastInstanceID), seqVal, maxContractID)
```
alpe, line 64: same as above

alpe (changes requested):
Thanks a lot for contributing to the project! 💐 
I have commented on the linked issue to give some background on the sequence values.

x/wasm/keeper/genesis.go
fredkneeland: Ah, I see I have this backwards as the check determines if we throw an error... thanks!

PR #645, opened 2021-10-13
Add golangci lint check on pull requests

fkneeland-figure:
Resolves #620 
I added a github action to run the golangci lint test on all non-test golang files when a pr is created.

PR #648, opened 2021-10-14
Go Formatting update

fkneeland-figure:
Resolves: #649 

The command `make format` makes a bunch of modifications.  This pr just contains the changes from running the automatic go formatters.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 **Implemented Enhancements:**
 
+- Remove unused flags from command prompt for storing contract [\#647](https://github.com/CosmWasm/wasmd/issues/647)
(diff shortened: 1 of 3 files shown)


alpe:
This will conflict with #645 as most files were touched there already. I would suggest to set this PR on hold and start fresh again after 645 is merged.

fredkneeland:
@alpe I think it would be better if we merged this pr, as all the changes here are just automatic formatting change.  Unfortunately, I added these commits to #645 as I didn't realize I would have to make a bunch of additional changes.  So there won't be any merge conflicts if this is merged first, but it will greatly reduce the complexity of #645 as it won't have all these autoformatting changes added on top.  Thanks!

PR #658, opened 2021-10-21
Fix message representation for signing

### LINKED ISSUE #642: Amino JSON representation of inner message in Msg{Instantiate,Migrate,Execute}Contract
With the change in https://github.com/CosmWasm/wasmd/pull/520 we most likely changed the JSON representation of the `msg` field in `MsgInstantiateContract`, `MsgMigrateContract` and `MsgExecuteContract` by accident. It is now a base64 encoded string of the JSON. The problem with this is that Ledger users cannot read the content of the smart contract interaction anymore. The same applies for other generic Amino JSON signers that do not have a dedicated UI for those message types.

While most of the #520 makes sense, this diff in the autogenerated ` x/wasm/types/tx.pb.go` is problematic:
Bildschirmfoto 2021-10-13 um 14 40 56

As you can see in https://play.golang.org/p/ois4zkjeRbr, the type defines the JSON serialization.

Seems like we should have a custom seralizer for those types that
1. Keeps the `[]byte` type for the inner messages because we have no clue what is in there
2. Converts to JSON (can fail) and then embeds the message as an object as part of the signdoc generation

alpe:
Resolves #642 

This PR introduces  the `RawContractMessage` type for incoming and outgoing data to a contract. 
I have also revisited some other places where we were using `json.RawMessage` currently or before the change:

The protobuf diffs describe the changes best

### DIFF
--- a/x/wasm/types/tx.go
+++ b/x/wasm/types/tx.go
@@ -10,6 +10,8 @@ import (
 )
 
 // RawContractMessage defines a json message that is sent or returned by a wasm contract.
+// This type can hold any type of bytes. Until validateBasic is called there should not be
+// any assumptions made that the data is valid syntax or semantic.
 type RawContractMessage []byte
 
 func (r RawContractMessage) MarshalJSON() ([]byte, error) {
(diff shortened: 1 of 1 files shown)


x/wasm/types/tx.go
```diff
@@ -2,12 +2,43 @@ package types
...
+	}
+	return r, nil
+}
+
+func (a *RawContractMessage) UnmarshalJSON(b []byte) error {
+	// copied from json.RawMessage#UnmarshalJSON
+	if a == nil {
+		return errors.New("unmarshalJSON on nil pointer")
+	}
+	*a = append((*a)[0:0], b...)
```
ethanfrey, line 28: Using this type rather than json.RawMessage everywhere gives flexibility.

What about doing `ValidateBasic()` in `UnmarshalJSON()`? Then we guarantee it is always valid

PR #668, opened 2021-11-05
Add limiting wasm smart query decorator

### LINKED ISSUE #655: Limit query contract depth
Inspired by https://github.com/terra-money/core/pull/584

We want to prevent from recursive smart queries that consume a lot of resources but may not cost much fees.
Options we have:

a) allow level n depth of queries where each contract that runs a smart query increases the depth by 1.
```
example:  
  A(0) -calls-> B, B(1) -calls -> C(2) -calls->D 
  A(0) -calls-> B, A(0) -calls -> C(1) -calls->D
```
b) allow n number of total smart queries in total
```
example:  
  A(0) -calls-> B, B(1) -calls -> C(2) -calls->D 
  A(0) -calls-> B, A(1) -calls -> C(2) -calls->D

```

alpe:
🚧  DRAFT - untested
Resolves #655

### DIFF
--- a/x/wasm/keeper/query_plugins.go
+++ b/x/wasm/keeper/query_plugins.go
@@ -92,6 +92,7 @@ func DefaultQueryPlugins(
 	channelKeeper types.ChannelKeeper,
 	queryRouter GRPCQueryRouter,
 	wasm wasmQueryKeeper,
+	maxQueryDepth uint32,
 ) QueryPlugins {
 	return QueryPlugins{
 		Bank:     BankQuerier(bank),
@@ -99,7 +100,7 @@ func DefaultQueryPlugins(
 		IBC:      IBCQuerier(wasm, channelKeeper),
 		Staking:  StakingQuerier(staking, distKeeper),
 		Stargate: StargateQuerier(queryRouter),
-		Wasm:     WasmQuerier(wasm),
+		Wasm:     LimitedDepthWasmQuerier(maxQueryDepth, WasmQuerier(wasm)),
 	}
 }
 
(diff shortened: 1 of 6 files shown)


x/wasm/keeper/query_plugins.go
```diff
@@ -499,7 +500,23 @@ func WasmQuerier(k wasmQueryKeeper) func(ctx sdk.Context, request *wasmvmtypes.W
...
 		}
-		return nil, wasmvmtypes.UnsupportedRequest{Kind: "unknown WasmQuery variant"}
+		return nil, wasmvmtypes.UnsupportedRequest{Kind: "unknown wasm query variant"}
+	}
+}
+
+// LimitedDepthWasmQuerier decorator to WasmQuerier to limit the depth of recursive calls
+func LimitedDepthWasmQuerier(maxDepth uint32, next func(ctx sdk.Context, request *wasmvmtypes.WasmQuery) ([]byte, error)) func(ctx sdk.Context, request *wasmvmtypes.WasmQuery) ([]byte, error) {
+	return func(ctx sdk.Context, request *wasmvmtypes.WasmQuery) ([]byte, error) {
+		counter, ok := types.QueryDepthCounter(ctx)
```
ethanfrey, line 510: ah, storing this in context is nice to handle this over multiple execution paths.

but what are the requirements / expected behaviour? can you document that first?

e.g. If there is a limit of 10 depths, and I dispatch 3 wasm messages and each make 4 queries, I do believe this will trigger the issue, even if not nested.

But... I am not really sure where these contexts will get unwrapped, so maybe this only will trigger when Contract 1 -> contract 2 -> ... Contract 11  and not multiple messages using the querier one after another.

x/wasm/types/types.go
```diff
@@ -309,5 +312,6 @@ func DefaultWasmConfig() WasmConfig {
 		SmartQueryGasLimit: defaultQueryGasLimit,
 		MemoryCacheSize:    defaultMemoryCacheSize,
 		ContractDebugMode:  defaultContractDebugMode,
+		SmartQueryDepth:    defaultSmartQueryDepth,
```
ethanfrey, line 315: This can be overridden by genesis? Or per machine?

```
SmartQueryGasLimit: defaultQueryGasLimit,
MemoryCacheSize:    defaultMemoryCacheSize,
ContractDebugMode:  defaultContractDebugMode,
```

should not be consensus breaking, in that different nodes may have different values and they will handle external queries differently, have different performance profiles, etc but will return the same result for all transactions.

`SmartQueryDepth` will actually abort transactions and could cause consensus issues if different on different validator nodes.

Either we only allow setting it in the Go binary, or we set it in genesis (as x/params?). WasmConfig was read from a local toml file, right?

ethanfrey (commented):
Interesting issue.

I need to think how the contexts work (and would love some texts on this both nested and in sibling messages)

I think we need another approach for the configuration as this is consensus critical

alpe:
Not the right solution

PR #674, opened 2021-11-17
Limit simulation gas

### LINKED ISSUE #670: Prevent infinite gas consumption in simulation queries
- configurable per node
- default to block gas limit
- must not affect consensus

After some deeper analysis and discussion on #655 in our internal discord with @ethanfrey and @webmaster128 ,
we decided on a new path to solve the root cause. Or what may was the the root cause from the original issue on T.
No further infos provided 🤷‍♂️  
The problem described in #655 addresses only a subset of fraud scenarios as simulations run in the sdk with [infinite gas](https://github.com/cosmos/cosmos-sdk/blob/v0.42.10/x/auth/ante/setup.go#L71-L72)

alpe:
Resolves #670 

Adds a new startup param/config and ante handler to limit gas for simulations:
```sh
--wasm.simulation_gas_limit string                Set the max gas that can be spent when executing a simulation TX
```

### DIFF
--- a/x/wasm/keeper/ante.go
+++ b/x/wasm/keeper/ante.go
@@ -53,3 +53,44 @@ func encodeHeightCounter(height int64, counter uint32) []byte {
 func decodeHeightCounter(bz []byte) (int64, uint32) {
 	return int64(sdk.BigEndianToUint64(bz[0:8])), binary.BigEndian.Uint32(bz[8:])
 }
+
+// LimitSimulationGasDecorator ante decorator to limit gas in simulation calls
+type LimitSimulationGasDecorator struct {
+	gasLimit *sdk.Gas
+}
+
+// NewLimitSimulationGasDecorator constructor accepts nil value to fallback to block gas limit.
+func NewLimitSimulationGasDecorator(gasLimit *sdk.Gas) *LimitSimulationGasDecorator {
+	if gasLimit != nil && *gasLimit == 0 {
+		panic("gas limit must not be zero")
+	}
+	return &LimitSimulationGasDecorator{gasLimit: gasLimit}
+}
+
+// AnteHandle that limits the maximum gas available in simulations only.
+// A custom max value can be configured and will be applied when set. The value should not
+// exceed the max block gas limit.
+// Different values on nodes are not consensus breaking as they affect only
+// simulations but may have effect on client user experience.
+//
+// When no custom value is set then the max block gas is used as default limit.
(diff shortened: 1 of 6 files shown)


x/wasm/keeper/ante.go
```diff
@@ -53,3 +53,44 @@ func encodeHeightCounter(height int64, counter uint32) []byte {
...
+func (d LimitSimulationGasDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (sdk.Context, error) {
+	if !simulate {
+		// Wasm code is not executed in checkTX so that we don't need to limit it further.
+		// Tendermint rejects the TX afterwards when the tx.gas > max block gas.
+		// On deliverTX we rely on the tendermint/sdk mechanics that ensure
+		// tx has gas set and gas < max block gas
+		return next(ctx, tx, simulate)
+	}
+
+	// apply custom node gas limit
```
ethanfrey, line 86: Looks good. Maybe this could be "simplified"? But fine as is.

```go
if d.gasLimit != nil {
  ctx = ctx.WithGasMeter(sdk.NewGasMeter(*d.gasLimit))
} else if maxGas := ctx.ConsensusParams().GetBlock().MaxGas; maxGas > 0 { 
  ctx = ctx.WithGasMeter(sdk.NewGasMeter(sdk.Gas(maxGas)))
}
return next(ctx, tx, simulate)
```

x/wasm/keeper/ante_test.go
```diff
@@ -107,3 +111,77 @@ func TestCountTxDecorator(t *testing.T) {
 		})
 	}
 }
+func TestLimitSimulationGasDecorator(t *testing.T) {
```
ethanfrey, line 114: nice tests

ethanfrey (approved):
lgtm. nice fix

x/wasm/types/types.go
```diff
@@ -296,6 +296,10 @@ func NewWasmCoins(cosmosCoins sdk.Coins) (wasmCoins []wasmvmtypes.Coin) {
 
 // WasmConfig is the extra config required for wasm
 type WasmConfig struct {
+	// SimulationGasLimit is the max gas to be used in a tx simulation call.
+	// When not set the consensus max block gas is used instead
+	SimulationGasLimit *uint64
+	// SimulationGasLimit is the max gas to be used in a smart query contract call
```
maurolacy, line 302: Just taking a look at this. You mean, `SmartQueryGasLimit`.

PR #656, opened 2021-10-19
New sdk new wasm

faddat:
Hi!

I had to make a few judgement calls when producing this, but pretty sure that I made the correct ones.  

I kept cosmwasm/wasmd's gas settings at all times, and brought in provenance's changes that account for the lack of sdk.addrlen

I also included some of their comments, some of em were interesting 

wdyt?

PR #678, opened 2021-11-21
Upgrade to cosmos-sdk v0.44

### LINKED ISSUE #501: Update to Cosmos SDK v0.44
Currently, pre-release 
https://github.com/cosmos/cosmos-sdk/releases/tag/v0.43.0-alpha1

may be worth basing 1.0 RFC off here, as it contains quite a few breaking changes and the future state sync hook patch will be on top of it

sunnya97:
closes #501


sunnya97:
I just realized that there is already a PR for this in #656, and I think that's actually farther along in the process than I am.  Should probably work off of that.

ethanfrey:
Hi Sunny, there are a lot of talks about this upgrade in the issue and on discord. Not sure if everything is written down.

In any case, there are at least 3 known forks that add 0.44 support. It is unclear all the design decisions/tradeoffs each one makes. But I think there is no need for another one. The best way to help move this forward is to reach out to @alpe and maybe offer a review of some code.

BTW, the biggest open question I think is with the 0.44 upgrade whether contracts should maintain 20 bytes "legacy" addresses like secp256k1, or use the new 32 byte "module account" address format.

ethanfrey:
We have communicated putting this into our 1.0 roadmap and will get started on it in a week or two.

sunnya97:
Got it!  I'm gonna close this PR for now then.  Btw, I would say I have a preference for the new 32 byte addresses, but I wonder if we do need to provide legacy support for old addresses?

faddat:
https://github.com/CosmWasm/wasmd/commit/928275857acc4be93f50e62c7523ad450a3b1a66#commitcomment-61521105


Functionally speaking I am pretty good with the beta2-ibc2 branch at notional-labs/wasmd 

but.....

It would be hell to merge in and out (think like osmosis fork of cosmos sdk)

So I am going to basically re-do your work here, using it as a template of sorts.  I've also put together a whooooole bunch of tests and I hope to conclude with some kind of contract sanity test.  

Cheers!

PR #679, opened 2021-11-23
Go 1.17 provides a much clearer go.mod file

faddat:


alpe:
Thanks for the PR and little push to move to v1.17. 
Meanwhile gaiad and sdk are on v1.17 as well so let's do it as well. There are some other places like `.circleci/config` and the `Dockerfile`  that would need the upgrade as well. Would you upgrade them as well?

faddat:
Happy to make those changes :)

faddat:
Pretty sure that the remaining error is an issue on your build machine.

alpe:
Very nice. Thanks for the contribution 💐 
Benchmark tests are flickering indeed. It is on my list. Sorry for the inconvenience.

PR #682, opened 2021-11-24
Create build-test.yml

faddat:
This will make it easier for people forking and woking on this repo, so they won't need circleci.  I'll move over the other tests and things soon, too!

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -1,0 +1,55 @@
+on: [push, pull_request]
+name: every commit
+jobs:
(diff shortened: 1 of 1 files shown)


alpe:
I appreciate your effort but I don't think we should support 2 build systems. It would be easier if we can discuss this in an issue before talking about a concrete solution.
It is an assumption that people prefer github actions over circleCI that needs to be backed by some data/ references.

We also have the `/contrib` folder where this could go as an example to pick up with additional doc but no further support.

PR #692, opened 2021-12-07
Use replace statements to enforce consistent versioning.

faddat:
Hi, I think this is a potoentially serious issue.  

Before this commit, there are versions of tm as early as 0.34.0 and sdk 0.42.5 and iavl 16 being imported into wasmd.  After the replace statements are added, the versioning is consistent throughout the application, which ought to at the very least improve safety.  

This also fixes CI.  I think.


ethanfrey:
`go mod tidy` wouldn't have fixed this?
any idea where those imports come from? usually it is just leftover in the go.sum file from earlier

faddat:
Yep, see the PR on wasmd.  That accounts for at least some of them.

faddat:
@ethanfrey if we cut a release of wasmvm tagged like beta3, I can try this out and see how many of the replaces I can remove from go.mod here.

alpe:
👍  Thanks for the PR. Nice findings. 
I will have an 👀  on the replace statements 
🤔 The CircleCI executor was redundant with the docker image, I think. Please open an issue if you have other problems with CI

peterbourgon:
> Before this commit, there are versions of tm as early as 0.34.0 and sdk 0.42.5 and iavl 16 being imported into wasmd. 

This likely isn't true, and I think stems from a misunderstanding of the meaning of entries in the go.sum file. Also, [the replace directive isn't meant to be used](https://github.com/golang/go/wiki/Modules#when-should-i-use-the-replace-directive) as a way to pin versions, or to be long-lived, and it's not respected by downstream consumers.

This PR should be reverted.

ethanfrey:
Those replace statements just caused me confusion in another update PR.
Maybe we should revisit this

PR #693, opened 2021-12-07
changes to 42 series wasmd (CI & testing)

faddat:
In setting up a rigorous ci system, I discovered a usage of math.rand instead of crypto.rand. 

I don't know how much it matters, but gosec (the informal fork) flagged it, so here's a PR :)


x/wasm/types/test_fixtures.go
```diff
@@ -50,7 +50,7 @@ func randBytes(n int) []byte {
 	r := make([]byte, n)
 	rand.Read(r)
 	return r
-}
+}beta2
```
ethanfrey: I bet that was a typo

ethanfrey (commented):
Hmmm... for test cases, goes this matter?

I guess it gets picked up as `test_fixtures.go` is not technically test code

x/wasm/types/test_fixtures.go
faddat: yep

faddat:
Before I start working on forking again, I will omnibus all of my changes for "baseline ci" here.

x/wasm/types/test_fixtures.go
faddat: re: does it matter?

honestly when I do stuff sometimes I just choose to be a slave to the linters

PR #701, opened 2021-12-16
`ContractInfo` query: return `NoSuchContract` system error instead of sdk error

### LINKED ISSUE #687: Use proper SystemError::NoSuchContract on ContractInfo if missing
We return "normal" SDK errors rather than cosmwasm errors on many cases. This just gives strings to the contracts, which are not the most useful if we need to check if a contract is missing or not.

Let's make sure to return the proper wasmvm SystemErr type for these cases (at least missing contract).

Context (and workaround): https://github.com/confio/tgrade-contracts/pull/368/files#diff-25abef21c13511fbb128155747fbe8e908d0827b23e56ab0718c0c1b52acb44eR1206-R1215

maurolacy:
Closes #687.

### DIFF
--- a/x/wasm/keeper/query_plugins.go
+++ b/x/wasm/keeper/query_plugins.go
@@ -487,7 +487,7 @@ func WasmQuerier(k wasmQueryKeeper) func(ctx sdk.Context, request *wasmvmtypes.W
 			}
 			info := k.GetContractInfo(ctx, addr)
 			if info == nil {
-				return nil, sdkerrors.Wrap(sdkerrors.ErrUnknownAddress, request.ContractInfo.ContractAddr)
+				return nil, wasmvmtypes.NoSuchContract{Addr: request.ContractInfo.ContractAddr}
 			}
 
 			res := wasmvmtypes.ContractInfoResponse{


maurolacy:
Do we also map `sdkerrors.ErrInvalidAddress` in https://github.com/CosmWasm/wasmd/blob/9a17505cf8b637942cad78e294aa1377d7f7b2f6/x/wasm/keeper/query_plugins.go#L486

to a system error? For that I would need to create a proper system error in wasmvm.

maurolacy:
And, what about errors returned from the other queriers? (Bank, IBC, Staking, and Stargate)

x/wasm/keeper/query_plugins.go
```diff
@@ -487,7 +487,7 @@ func WasmQuerier(k wasmQueryKeeper) func(ctx sdk.Context, request *wasmvmtypes.W
 			}
 			info := k.GetContractInfo(ctx, addr)
 			if info == nil {
-				return nil, sdkerrors.Wrap(sdkerrors.ErrUnknownAddress, request.ContractInfo.ContractAddr)
+				return nil, wasmvmtypes.NoSuchContract{Addr: request.ContractInfo.ContractAddr}
```
ethanfrey, line 490: Ethan's idea - make wasmvm aware of Cosmos SDK error types

Implementing the following methods on the wasmvmtypes errors will allow them to be reported properly.

https://github.com/cosmos/cosmos-sdk/blob/master/types/errors/abci.go#L103
https://github.com/cosmos/cosmos-sdk/blob/master/types/errors/abci.go#L128

ethanfrey: Other approach by Alex:

* Map some known sdk errors (like these ErrUnknownAddress) to wasmvmtypes errors here https://github.com/CosmWasm/wasmd/blob/9a17505cf8b637942cad78e294aa1377d7f7b2f6/x/wasm/keeper/query_plugins.go#L59
* Define concrete sdk friendly errors in wasmd, don't use  `sdkerrors.ErrUnknownAddress` use `wasmtypes.ErrNoSuchContract`, which is an sdkerrors.Error type and registered in the mapping

ethanfrey: For Alex's ideas, see Register here https://github.com/cosmos/cosmos-sdk/blob/master/types/errors/errors.go

We can add a custom one that handles mapping to wasmvmtypes.

ethanfrey: We need the actual info of the address to form the proper wasmvm type.
Thus... we cannot use sdkerrors.Error, which doesn't have a place for that info.

We could make a custom type in x/wasm that exposes `ABCICode()` and `Codespace()` and `Error()`. This type is called "NoSuchContract{Addr}" and we map this one case to a custom wasmvm error on line 59 (above comment)

ethanfrey (commented):
Thoughts on refactoring error handling

ethanfrey:
What's the progress here? Is it blocked on something?

I think this is a good solution, and while we can do a larger overhaul, this fixes a clear issue with contracts now.

maurolacy:
I plan to improve on this after holidays. I agree this is good enough for now. We can merge if you / @alpe agree.

alpe:
I was waiting for the other PR but 👍  to have this first.

alpe (approved):

PR #714, opened 2022-01-04
Upgrade wasmvm to v1.0.0-beta5

alpe:
* Bump wasmvm version to [v1.0.0-beta5]}(https://github.com/CosmWasm/wasmvm/releases/tag/v1.0.0-beta5)
* Handle new `IBCPacketReceive` response


x/wasm/relay_pingpong_test.go
```diff
@@ -241,9 +242,9 @@ func (p player) IBCPacketReceive(codeID wasmvm.Checksum, env wasmvmtypes.Env, ms
 
 	if maxVal := store.Get(maxValueKey); maxVal != nil && otherCount > sdk.BigEndianToUint64(maxVal) {
 		errMsg := fmt.Sprintf("max value exceeded: %d got %d", sdk.BigEndianToUint64(maxVal), otherCount)
-		return &wasmvmtypes.IBCReceiveResponse{
+		return &wasmvmtypes.IBCReceiveResult{Ok: &wasmvmtypes.IBCReceiveResponse{
```
ethanfrey, line 245: Would it make sense to have a helper for this constructor?
Could be added in the next wasmvm update if you think it would help

x/wasm/keeper/relay.go
```diff
@@ -131,8 +131,11 @@ func (k Keeper) OnRecvPacket(
 	if execErr != nil {
 		return nil, sdkerrors.Wrap(types.ErrExecuteFailed, execErr.Error())
 	}
+	if res.Err != "" { // handle error case as before https://github.com/CosmWasm/wasmvm/commit/c300106fe5c9426a495f8e10821e00a9330c56c6
```
ethanfrey, line 134: Yes, I agree with these changes.
Current contracts shouldn't hit this branch, but if they do, they behave the same as before the update.

ethanfrey (approved):
lgtm

x/wasm/relay_pingpong_test.go
alpe: ja, would be nice to have constructor methods in wasmvm for ok and err. 👍

Issue #697, opened 2021-12-13
Update ibc handling to accommodate ibc v3


ethanfrey:
With the current ibc implementation, we take great care a contract never errors on ibc_recv_packet. It must also take care to roll back any state internally on error (using submessage to self if needed). Any error it returns would revert the entire transaction, rather than signal a failure in the acknowledgement.

In ibcv2, we have 4 cases to cover:
1. success
2. application failure and NO rollback
3. application error and rollback app state
4. tx error and rollback ibc state as well

(1 and 2 we handled by the current version, only difference is the interpretation of bytes in the ack data)

In https://github.com/CosmWasm/wasmvm/blob/main/lib.go#L540-L547 we unmarshall the response from the contract and consider a `Result::Err` from the contract the same as a VM error and convert both to a Go error. (VM error is "out of gas", "panic", "no ibc entry point" or other hard abort)

One step suggested by @alpe would be returning the complete `IBCReceiveResult` rather than unmarshalling it in wasmvm. That will provide more info to the caller. However, how do we handle these two cases:

3. contract returned error
4. vm returned error

Until wasmd 0.21 (on cosmos-sdk 0.42) both are handled the same and abort the transaction, which means the ibc packet is never recorded as received, so no ack can be returned to the sender. Can we do something better now?

ethanfrey:
Personally, I would like to take advantage of the new interface. Here are the conditions we want to ensure keep working:

* If a contract returns success with `[]byte("xyzzy")` as ack data, those exact bytes will be received by the counterpart chain to give them full control
* This should support both utf-8 (JSON) data as well as binary (protobuf) data formats to match any native protocol
* On out of gas error we should abort the transaction (return go error), so it can be retried (not recorded like the app called it a failure)

Here are some open cases to cover.

* The contract returns an error result (eg "InsufficientBalance"). It would be nice to let it return an error and have that be passed to the other side as an error. And rollback all state changes.
* The contract panics. Maybe we should consider that as above (application error) or maybe we should handle that like out of gas (abort tx)

Why do I say this? Look at the hoops we must do for a very simple handler:

* [On recv, make a special function and handle all errors separately](https://github.com/CosmWasm/cw-plus/blob/main/contracts/cw20-ics20/src/ibc.rs#L167-L197)
* [Ensure we catch the error case for any submessages](https://github.com/CosmWasm/cw-plus/blob/main/contracts/cw20-ics20/src/ibc.rs#L332-L354)
[...]

Issue #696, opened 2021-12-10
Change SDKMessageHandler to accept an interface


channa-figure:
## Summary

The Provenance team has been working on a feature for adding addition fees for messages.  We ran into an issue with the `MsgServiceRouter` being a struct instead of an interface.  This caused an issue with trying to customize the behavior of message handling to fit our needs, i.e. we needed to add additional charges to msgs that were dispatched by a wasm contract.  We ended up making a fork of the cosmos-sdk and adding interface to the `MsgServiceRouter`.  We also changed the wasmd module's `SDKMessageHandler` to accept that interface: [change on our wasmd fork](https://github.com/CosmWasm/wasmd/commit/86a7c964346321a399229025abbb8b3a727afef4)   

## Proposal

Change the `SDKMessageHandler` to accept an interface for the `MsgServiceRouter` to allow users to inject custom behavior on msg routing.

iramiller:
This is really only needed on the 0.44.x series as the upcoming middleware refactor provides ways to control the behavior of the system we are looking for.

ethanfrey:
I like the idea of making it more extensible. And the general Go maxim "interfaces in, structs out"

Unless Alex sees a code issue here, I approve the idea. The current calling code should work unchanged, we just allow more possibilities

alpe:
An interface makes perfect sense. Thanks for highlighting your use case 👍

alpe:
This was implemented

PR #717, opened 2022-01-05
Upgrade to cosmos-sdk v0.45.0

### LINKED ISSUE #501: Update to Cosmos SDK v0.44

alpe:
Resolves #501 

This upgrade includes
* Upgrade to cosmos-sdk version [v0.45.0](https://github.com/cosmos/cosmos-sdk/releases/tag/v0.45.0)
* Upgrade to new IBC version: [v2.0.2](https://github.com/cosmos/ibc-go/releases/tag/v2.0.2)
* Use **32 byte** address length for contract addresses
* Replace msg.Route() based handlers with ADR 031 request type routing for contract messages
* Add [feegrant](https://github.com/cosmos/cosmos-sdk/tree/v0.44.5/x/feegrant) module to demo app
* Add [authz](https://github.com/cosmos/cosmos-sdk/tree/v0.44.5/x/authz) module to demo app

Follow ups:
- Should we add support for "feegrant" to cosmwasm?
- Should we add support for "authz" to cosmwasm?
- Should we integrate IBC version negotiation, https://github.com/CosmWasm/wasmvm/issues/269?
- Should we pass the IBC relayer address to the contract?
- Should we do the IBC receive package (n)ack error cases differently?


ethanfrey:
Thank you @alpe for this PR.

Re the follow ups. This is a very important question before merging, or at least before tagging a release.

* Should we do the IBC receive package (n)ack error cases differently?

-> This is an important question we should work out in the design docs and make a decision there before implementing. This has been heavily discussed in https://github.com/CosmWasm/wasmd/issues/697 and changes were made to the wasmvm interface to expose all needed info to the runtime, but we need to document the behavior.

These ones would be nice to have by 1.0 but require extending the CosmWasm interfaces in a backwards compatible way:

* Should we add support for "feegrant" to cosmwasm?
* Should we add support for "authz" to cosmwasm?
* Should we integrate IBC version negotiation, Support IBC version negotiation in contracts wasmvm#269?
* Should we pass the IBC relayer address to the contract?

-> I added a tracking issue for this in CosmWasm https://github.com/CosmWasm/cosmwasm/issues/1200 Please add more thoughts/comments on these to that issue

ethanfrey:
@faddat @iramiller @ValarDragon @RiccardoM @litvintech

I would love any comments here as you all have been working on 0.44 forks.

channa-figure:
I really like the PR the changes are clean and look good.  I have been looking forward to seeing the differences and similarities between our forks changes.

However, we were hoping that issue #696 would have made it in this PR.  I see it is currently target for milestone v1.0.0.  Is it possible to bump it into this?  If not, we will just maintain a fork and wait until v1.0.0.

faddat:
Hi!

a moment to collect thoughts :)

ethanfrey:
All of those require extensions to the CosmWasm interface.
I would get this in with the same API, and then work on those once we support them in CosmWasm (another release).
There is a linked tracking issue in CosmWasm to discuss this as well.

ethanfrey:
> > Should we pass the IBC relayer address to the contract?
> I'm a reasonably large relayer, and I'm not sure what benefit that could bring -- in fact, it may be best to keep contracts unaware of who is doing the relaying since we (relayers) provide a generic service.

This will be needed if we want to provide bonuses / fees / incentives to relayers in the future. It should not be used for auth

ethanfrey:
Unless anyone has concerns, I feel this is good to merge

ethanfrey (approved):
good stuff

go.mod
```diff
@@ -101,24 +108,30 @@ require (
...
+	gopkg.in/ini.v1 v1.66.2 // indirect
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
+	nhooyr.io/websocket v1.8.6 // indirect
 )
 
 replace (
+	github.com/99designs/keyring => github.com/cosmos/keyring v1.1.7-0.20210622111912-ef00f8ac3d76
 	github.com/confio/ics23/go => github.com/confio/ics23/go v0.6.6
-	github.com/cosmos/cosmos-sdk => github.com/cosmos/cosmos-sdk v0.42.11
+	github.com/cosmos/cosmos-sdk => github.com/cosmos/cosmos-sdk v0.45.0-rc1
```
tomtau, line 126:
```suggestion
	github.com/cosmos/cosmos-sdk => github.com/cosmos/cosmos-sdk v0.45.0
```

ethanfrey: @alpe why is this even here?
`github.com/cosmos/cosmos-sdk v0.45.0` is set above and we don't change repos.

I would remove it

ethanfrey (commented):
good point

go.mod
tomtau: I think it's because of ibc-go which may be bringing in a different version of cosmos-sdk

ethanfrey: Turns out they were added here: https://github.com/CosmWasm/wasmd/pull/692/files

With a recent comment they are not a good idea: https://github.com/CosmWasm/wasmd/pull/692#issuecomment-1017078882

I fixed them with your hint, but maybe they need to be removed as well....

alpe: ja, a transitive dependency. I did not look where it came from but just wanted to ensure we have the right sdk version set.

alpe:
lgtm 🎉

PR #720, opened 2022-01-17
SDK 045

faddat:
Hi, to the best of my knowledge, 045 has some gas changes that will be useful for cw, so I've gone ahead and bumped this to v0.45.0


ethanfrey:
Thanks for this PR.
It seems to fail with the new SetOrderMigrations assertion.
I am not sure what the suggested values for that are (I am sure there are some from simapp). It would be nice to see this pass CI.

alpe:
@faddat  thanks for the PR! There was a bit more work with app.go and some gas tests settings. I am closing this now in favour of #721

PR #721, opened 2022-01-17
Upgrade to SDK v0.45.0-rc1

alpe:
Upgrade to [Cosmos-SDK 0.45.0-rc1](https://github.com/cosmos/cosmos-sdk/releases/tag/v0.45.0-rc1)

For ref https://github.com/cosmos/cosmos-sdk/issues/10656

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -29,6 +29,7 @@ import (
 	authtx "github.com/cosmos/cosmos-sdk/x/auth/tx"
 	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
 	"github.com/cosmos/cosmos-sdk/x/auth/vesting"
+	vestingtypes "github.com/cosmos/cosmos-sdk/x/auth/vesting/types"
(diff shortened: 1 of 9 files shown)


app/app.go
```diff
@@ -536,22 +537,27 @@ func NewWasmApp(
 		slashingtypes.ModuleName,
 		evidencetypes.ModuleName,
 		stakingtypes.ModuleName,
-		ibchost.ModuleName,
-	)
-	app.mm.SetOrderEndBlockers(
```
ethanfrey, line 541: A lot new modules were added here.
Were they missing before or does it now just require everyone to have an empty end blocker implementation?

ethanfrey (approved):
lgtm
thanks for getting this working.
curious about the extra end blockers, but looks good if the ci is happy

app/app.go
alpe: All modules need to be listed now

PR #726, opened 2022-01-20
Autopin gov contracts

### LINKED ISSUE #702: Auto-pin code when uploaded by governance
If it is important enough to have a governance vote to StoreWasm, it is most likely important enough to Pin.

Let's do this by default. Either add a field which must be explicitly set to disable auto-pin (empty is pin), or just always auto-pin and force them to manually vote to unpin if there are too many codes.

Thinking of some concrete use cases here, where this will be a very helpful default.

ethanfrey:
Closes #702

Simply calls pin after creation if we create it via a gov proposal


ethanfrey:
@alpe This should be fine. I just don't cover the case of PinCode returning an error, thus the patch cover issue

alpe (approved):
lgtm

ethanfrey:
Changelog entry added. This should also be ready to merge (merge conflicts on CHANGELOG with #727 but we can just keep both lines)

PR #727, opened 2022-01-20
You must explicitly declare --no-admin on cli instantiate if that is what you want

### LINKED ISSUE #719: Clearer user expectations around admin field
As discussed on discord, a user must either set:

`--admin=<admin-addr>` when instantiating

or explicitly say they are not going to set an admin:

`--no-admin`

Otherwise an error will be raised.

This is to avoid the potential 🦶 🔫  caused by users not RTFM and realising too late that CW migration functionality is denied to them because they didn't set an admin.

ethanfrey:
Closes #719

### DIFF
--- a/x/wasm/client/cli/genesis_msg_test.go
+++ b/x/wasm/client/cli/genesis_msg_test.go
@@ -142,6 +142,7 @@ func TestInstantiateContractCmd(t *testing.T) {
 				flagSet := cmd.Flags()
 				flagSet.Set("label", "testing")
 				flagSet.Set("run-as", myWellFundedAccount)
+				flagSet.Set("no-admin", "true")
 			},
 			expMsgCount: 1,
 		},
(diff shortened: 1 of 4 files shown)


x/wasm/client/cli/genesis_msg_test.go
```diff
@@ -202,6 +206,59 @@ func TestInstantiateContractCmd(t *testing.T) {
...
+					},
+				},
+			},
+			mutator: func(cmd *cobra.Command) {
+				cmd.SetArgs([]string{"1", `{}`})
+				flagSet := cmd.Flags()
+				flagSet.Set("label", "testing")
+				flagSet.Set("run-as", myWellFundedAccount)
+				flagSet.Set("no-admin", "true")
+				flagSet.Set("admin", myWellFundedAccount)
```
alpe, line 261: good tests

alpe (approved):
LGTM 🎸

ethanfrey:
Rebased and updated changelog. Should be good to merge

PR #728, opened 2022-01-21
feat: Starport config file

fadeev:
Many Starport users are asking for CosmWasm. Cosmos SDK is notorious for how hard it is to add modules to an existing chain `app.go` (either manually or programmatically). Until Cosmos SDK becomes better at handling modules, I propose adding a simple `config.yml`, so that users can do this:

```
git clone https://github.com/cosmwasm/wasmd
cd wasmd
starport chain serve
```

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -1,0 +1,16 @@
+accounts:
+  - name: alice
+    coins: ["20000token", "200000000stake"]
+  - name: bob
+    coins: ["10000token", "100000000stake"]
+validator:
+  name: alice
+  staked: "100000000stake"
+client:
+  openapi:
+    path: "docs/static/openapi.yml"
+faucet:
+  name: bob
+  coins: ["5token", "100000stake"]
+build:
+  main: "cmd/wasmd"


alpe:
Thanks for the PR. 👏 Nice to see this!

I have run `starport chain serve` but beside this, I can not confirm or decline that it is working as expected or is complete. For a proper peer review, I would appreciate it if somebody else with more "Starport experience" than me can take a 👀 .

Otherwise, I would keep this PR open until I have some capacity to investigate myself.

If you would provide some description in our "INTEGRATION.md" about Starport or link to the docs, that may be helpful, too.

faddat:
I have more starport experience, and have looked through the annotations in app.go on wasmd. 

I think this is a very good idea so that:

* wasmd is used as a template
* wasmd gets more exposure to developers
* cw integration and usage from starport is streamlined

....basically what @fadeev is suggesting can help with making this the canonical cosmwasm implementation.  

Having looked at many app.go's, I'll share a thought/observation....

It may be desirable to allow developers to import an app.go.  Likecoin did this with gaia's cmd:

https://github.com/likecoin/likecoin-chain/blob/32056fc4759d276a11180256474cc4ed5e3ed8d4/cmd/liked/cmd/cmd.go#L47

similarly, the orm module is interesting

PR #730, opened 2022-01-21
Add sudo and execute proposal types

### LINKED ISSUE #672: Add gov proposal to call sudo
Contracts may expose `sudo` method that may be called by go modules to expose admin actions (like migrate) that are validated by other components of the blockchain. 

We have some gov proposal types that grant privileged access to other calls (like migrate), so it would make sense to add a similar proposal to call sudo. 

(sudo came after the gov proposals, so this was am oversight when adding it)

### LINKED ISSUE #711: Add WasmExecute proposal
Like #703 but for MsgWasmExecute which we do not support yet. 

https://github.com/CosmWasm/wasmd/blob/master/proto/cosmwasm/wasm/v1/proposal.proto

This is for on chain governance to run any contract as any user (setting `run_as`). Not just about token transfers, but assuming gov_contract anywhere. We should be careful to enable the proposal and keep it off by default. But I think this can be useful in some cases.

ethanfrey:
Closes #672 
Closes #711

- [x] Define protobuf definitions
- [x] Wire up as proposals in wasm module
- [x] Implement proposal handlers
- [x] Test proposal handlers
- [x] Add cli commands for them
- [x] Add rest handlers for them
- [x] Wire up in cli/rest
- [x] Test cli tooling (manually?)


ethanfrey:
@alpe I think this is done and covered with tests. Can you please review and let me know how to test the cli better?

x/wasm/keeper/proposal_handler.go
```diff
@@ -115,6 +119,53 @@ func handleMigrateProposal(ctx sdk.Context, k types.ContractOpsKeeper, p types.M
...
+		return sdkerrors.Wrap(err, "contract")
+	}
+	data, err := k.Sudo(ctx, contractAddr, p.Msg)
+	if err != nil {
+		return err
+	}
+
+	ctx.EventManager().EmitEvent(sdk.NewEvent(
+		types.EventTypeGovContractResult,
+		sdk.NewAttribute(types.AttributeKeyResultDataHex, hex.EncodeToString(data)),
```
alpe, line 138: 👍

x/wasm/keeper/proposal_handler.go
```diff
@@ -115,6 +119,53 @@ func handleMigrateProposal(ctx sdk.Context, k types.ContractOpsKeeper, p types.M
...
+
+	contractAddr, err := sdk.AccAddressFromBech32(p.Contract)
+	if err != nil {
+		return sdkerrors.Wrap(err, "contract")
+	}
+	runAsAddr, err := sdk.AccAddressFromBech32(p.RunAs)
+	if err != nil {
+		return sdkerrors.Wrap(err, "run as address")
+	}
+	// we currently don't support sending tokens as part of an execute proposal. should we? from which account?
```
alpe, line 156: Good question! For me it would make sense to support this from the community pool for example.

x/wasm/keeper/proposal_integration_test.go
```diff
@@ -206,6 +206,119 @@ func TestMigrateProposal(t *testing.T) {
...
+	require.NoError(t, err)
+
+	// and proposal execute
+	handler := govKeeper.Router().GetRoute(storedProposal.ProposalRoute())
+	err = handler(ctx.WithEventManager(em), storedProposal.GetContent())
+	require.NoError(t, err)
+
+	// balance should be empty (proper release)
+	bal = bankKeeper.GetBalance(ctx, contractAddr, "denom")
+	require.Equal(t, bal.Amount, sdk.NewInt(0))
```
alpe, line 266: 👍 good test

alpe (approved):
LGTM 🎸 
Please update the method doc in https://github.com/CosmWasm/wasmd/blob/master/x/wasm/keeper/keeper.go#L441
as it would not be correct anymore.

If you can add the new types to `Governance.md` then this would be 🚀 

Disclaimer:
I have checked that the CLI commands are listed in `wasmd tx gov submit-proposal -h` but not executed a proposal

proto/cosmwasm/wasm/v1/proposal.proto
```diff
@@ -56,6 +56,7 @@ message MigrateContractProposal {
   string title = 1;
   // Description is a human readable text
   string description = 2;
+  // FIXME: I think this is unused? Migrate has no sender
```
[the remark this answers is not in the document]
ethanfrey: Done in https://github.com/CosmWasm/wasmd/pull/730/commits/0a96e26fbad8ff67cd130ad151c7f8e968c12d09

proto/cosmwasm/wasm/v1/proposal.proto
```diff
@@ -66,6 +67,33 @@ message MigrateContractProposal {
...
+message ExecuteContractProposal {
+  // Title is a short summary
+  string title = 1;
+  // Description is a human readable text
+  string description = 2;
+  // RunAs is the address that is passed to the contract's environment as sender
+  string run_as = 3;
+  // Contract is the address of the smart contract
+  string contract = 4;
+  // Msg json encoded message to be passed to the contract as sudo
```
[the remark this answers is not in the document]
ethanfrey: done in https://github.com/CosmWasm/wasmd/pull/730/commits/0a96e26fbad8ff67cd130ad151c7f8e968c12d09

x/wasm/keeper/proposal_handler.go
ethanfrey: Okay, I will add it to the proposal type

ethanfrey:
Thank you for the feedback. I have addressed all comments, and also added the funds field to the new execute contract proposal type. I also updated the changelog and this is ready to merge

PR #731, opened 2022-01-24
Remove unneeded version enforcement

faddat:
Basically, go.sum has nothing to do with what is used to compile.  It is checksums that ensure that you've got the actual package.  Whole thing still baffles me, but I think I've removed what's not needed :).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -121,17 +121,14 @@ require (
 )
 
 replace (
+	// Use the cosmos-flavored keyring library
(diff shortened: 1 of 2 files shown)


ethanfrey:
Makes sense.
Can you link to the other discussion/comment about the why (I don't have the issue at hand), so there is a better paper trail (and anyone on that issue and see this).

alpe:
Thanks for taking care @faddat  💯 
This issue looks related https://github.com/cosmos/gaia/pull/1099 and contains some good details

alpe (approved):
🥇

PR #732, opened 2022-01-24
Expose SetParams

ethanfrey:
as I will need that for migration code outside of this package


alpe (approved):
I tried to avoid making this public as long as possible but I guess it is time now 😢

PR #733, opened 2022-01-25
Upgrade docker images in dockerfile

alpe:
* Fixes failing docker build
* Enables docker build for all branches

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -143,7 +143,9 @@ jobs:
       - attach_workspace:
           at: /tmp/workspace
       - checkout
-      - setup_remote_docker
+      - setup_remote_docker:
+          # >= v20.10 https://wiki.alpinelinux.org/wiki/Release_Notes_for_Alpine_3.14.0#faccessat2
+          version: 20.10.11
       - run:
           name: Build Docker artifact
           command: docker build --pull -t "cosmwasm/wasmd:${CIRCLE_SHA1}" .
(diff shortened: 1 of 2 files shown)


ethanfrey:
Looking at the build failure... 

"Setup a remote docker engine"

```
Server Engine Details:
  Version:          17.09.0-ce
  API version:      1.32 (minimum version 1.12)
  Go version:       go1.8.3
  Git commit:       afdb6d4
  Built:            2017-09-26T22:40:56.000000000+00:00
  OS/Arch:          linux/amd64
  Experimental:     false
```

Think this is a version issue?

ethanfrey:
Bisecting when docker broke in master, I found this:

Screenshot 2022-01-25 at 13 22 30

ethanfrey:
The last working one had the same docker engine

alpe:
The "issue" was introduced with the alpine image upgrade. The remote docker version became incompatible. See https://wiki.alpinelinux.org/wiki/Release_Notes_for_Alpine_3.14.0#faccessat2 for details

.circleci/config.yml
```diff
@@ -182,10 +186,10 @@ workflows:
       - docker-image:
           requires:
             - setup-dependencies
-          filters:
-            branches:
-              only:
-                - master
+#          filters:
```
ethanfrey, line 189: We need these back!

ethanfrey (commented):
I think you merged without the filters.

We should get those back (one more PR before tagging)

PR #734, opened 2022-01-25
Run docker build only for master branch

alpe:
Enable filters to run docker build only on master branch.
(Disabled in #733)


ethanfrey (approved):
Looks good.

We can wait a bit to merge it if you want. I think it is fine to only run there, but we should look at the big red X on master sometime... I ignored this as well (it is on the main wasmd page)

PR #740, opened 2022-01-28
Minor integration support

alpe:
Open some internal methods to make it easier to integrate in tgrade and other projects


ethanfrey (approved):
Nice one

PR #752, opened 2022-02-10
Add missing colons in String of some proposals

loloicci:
Some proposals miss a colon after `Msg` in string yaml format.
This PR adds them.


alpe (approved):
Good findings! Thanks for your contribution 💐

PR #753, opened 2022-02-11
Add GH action to publish proto files to BSR

### LINKED ISSUE #706: Publish protobuf files to https://buf.build/
Inspired by https://github.com/CosmWasm/wasmd/issues/705

pinosu:
Resolves #706 

GH Action includes:
- check for lint errors
- check for backward compatibility breaking changes

Migrated buf config files from v1beta1 to v1 using command:
'buf config migrate-v1beta1'
(https://docs.buf.build/configuration/v1beta1-migration-guide)


.github/workflows/proto-buf-publisher.yml
```diff
@@ -0,0 +1,29 @@
...
+      - uses: actions/checkout@v2
+      - uses: bufbuild/buf-setup-action@v0.7.0
+
+      # lint checks
+      - uses: bufbuild/buf-lint-action@v1
+
+      # backward compatibility breaking checks
+      - uses: bufbuild/buf-breaking-action@v1
+        with:
+          against: 'https://github.com/CosmWasm/wasmd.git#branch=master'
```
alpe, line 24: Does this break the build?
This looks useful when we have a v1 branch. Currently this project is not stable and we may have to do breaking changes.

.github/workflows/proto-buf-publisher.yml
```diff
@@ -0,0 +1,29 @@
...
+
+jobs:
+  push:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: bufbuild/buf-setup-action@v0.7.0
+
+      # lint checks
+      - uses: bufbuild/buf-lint-action@v1
```
alpe, line 19: Does this break the build?
We have some linter in the `make proto-all`.

alpe (commented):
Nice start. 
I think the `buf-breaking-action` is very useful in the future.

.github/workflows/proto-buf-publisher.yml
pinosu: It will not fail the build, just adds comments to let you know that those are breaking changes.

.github/workflows/proto-buf-publisher.yml
pinosu: It might be redundant, but it just adds inline comments and does not break the build.

alpe (approved):
Good work. ⛹️‍♂️

PR #755, opened 2022-02-14
Add buf.md for proto API docs

pinosu:
In this file a module description could be added as exaplained here:
https://docs.buf.build/bsr/documentation


alpe (approved):
👍

PR #756, opened 2022-02-14
Add module name to buf.yml file

pinosu:
The module name is a required field as specified here:
https://docs.buf.build/bsr/overview


alpe (approved):
👍

PR #757, opened 2022-02-14
Add protobuf deps

pinosu:
Add dependencies to buf.yml
Remove third_party folder, since dependencies are fetched using the deps in buf.yml
Update Makefile
Update protocgen.sh script


alpe (commented):
Excellent work. 💯  
I expect that we need to pin the sdk protobuf versions to the sdk version that we use in go.mod. Can you :eye add the linked doc and add this to this PR, please? 
Bonus points for some minimal doc in proto/readme how to do it.

alpe:
This PR is pending as we need to find the "right" sdk protobufs versions to match our v0.45.x version

ethanfrey:
> This PR is pending as we need to find the "right" sdk protobufs versions to match our v0.45.x version

Can you explain this more? Seems like an issue with the buf publishing system as used by the cosmos-sdk?

Maybe @marbar3778 could help on this as he has been working to get more adoption of buf.build?

alpe:
Without "pinning" the protobufs to a fix version or tag, we would link to a moving target and could not have fully reproducible code-generations. The sdk publishes versions with the git hash as label (https://buf.build/cosmos/cosmos-sdk/history/main). This is great but seems to be for master only. We could not find the matching protobufs for the v0.45.1 hash [`2646b474c7beb0c93d4fafd395ef345f41afc251`](https://github.com/cosmos/cosmos-sdk/releases/tag/v0.45.1)

@marbar3778 it would be great if you could help with pushing the protobufs for this version and automating this step to have them available for future tags as well.

ethanfrey:
What is the state of this PR?

If it is done and makes better protobuf generation, let's get it updated and merged. Or close if not needed

pinosu:
We are still blocked for the same reason @alpe has explained above.

ethanfrey:
> We are still blocked for the same reason @alpe has explained above.

Ah, now I understand that comment.

ethanfrey:
@marbar3778 could you look into the buf.build publishing of the sdk (so they publish the actual releases, like v0.45.4) not just main branch?

Or pass this off to someone working on that in the core team?

marbar3778:
> @marbar3778 could you look into the buf.build publishing of the sdk (so they publish the actual releases, like v0.45.4) not just main branch?
> 
> Or pass this off to someone working on that in the core team?

we publish each time since proto version is not tied to software versioning. On buf you will see the sdk protofiles have v1beta1 and v1 tags.

pinosu:
Hi @marbar3778 , it was nice to meet you at gateway conference ! We didn't have the chance to discuss about this PR. I have now created a [PR](https://github.com/cosmos/cosmos-sdk/pull/11993) in cosmos-sdk, and I think this will solve the issue we are having when trying to include cosmos-sdk protobufs from BSR.
Please let me know if it makes sense for you.

pinosu:
Closing in favor of #1030

PR #760, opened 2022-02-21
Replace custom codec with sdk codec

### LINKED ISSUE #749: Custom client codec crashes with Rosetta enabled
We use a custom client [codec](https://github.com/CosmWasm/wasmd/blob/master/x/wasm/client/codec/marshaller.go) to not print empty elements. This unfortunately does not work with the current sdk (v0.45.1 + current master) and creates:

`panic: interface conversion: codec.Codec is *codec.ProtoCodec, not *codec.ProtoCodec (types from different packages)`

Root cause for this panic is the [type cast in the sdk](https://github.com/cosmos/cosmos-sdk/blob/v0.45.1/server/start.go#L355)

alpe:
Resolves #749


ethanfrey (approved):
The simplest fix.

Ideally the sdk uses this (IMO better) codec or makes it more extensible.
The code is in git history in any case.

Issue #759, opened 2022-02-20
Return minimal info on errors returned to contracts


ethanfrey:
When discussing what information that can be returned in Interchain Accounts, we realized that [error messages and events are not guaranteed to be the same on every system](https://github.com/cosmos/ibc-go/issues/701#issuecomment-1018484522) (they are not part of consensus).

We can safely return data and *error code*. They have changed events in the sdk in minor upgrades, but for the same version, it will be the same on all systems (regardless of build system).

I would start limiting what we return to the contract and for the error returned in the reply section, not include error strings, rather the error code. This will make it harder to debug, but it will remove an attack vector that has popped up unexpectedly time after time. (Not in attacks, but in issues with different build systems)

@webmaster128 what do you think about this? I know this was also a concern of yours for windows and arm64 builds.

@alpe I wonder if there is a subset of info we could safely whitelist, like messages generated from x/wasm itself that don't include strings that come from other parts of the sdk.

ethanfrey:
This would not change the contract interface, just the actual error strings returned by wasmd (which are not guaranteed in any case). It would be a breaking change for `wasmd` (no migration needed, but x/upgrade for sure) but no effects on CosmWasm.

alpe:
The error message is helpful for debugging issues but should not be used to build logic on. We could follow the ABCI model and reduce any error to their ABCI code id, indeed. That should be straight forward. Unknown errors would be redacted to "internal". With tracing flag set, we can still print the full error to the console for debugging.
If we want to be more verbose then the root error message could be used but it is likely not super helpful without more context.

ethanfrey:
When contract A sends a message to contract B and it fails, we usually get an error message that is quite helpful for debugging (e.g message missing some field, etc). If we just replace that with `ExecutionFailed`, this would make life for developers much harder. 

We should think about this and pick one of these:

A. Don't care, just lock everything down. Every error that gets sent to a contract in reply is redacted.
B. Allow some "debug" flag to show these messages. This would be set on a node and meant for testnets. Kind of like the `--trace` message. Not sure if this would really help.
C. Go through all error message sources, try to figure out which ones are surely deterministic and don't filter those (whitelist them) and redact the rest.

Note that in any case, we don't redact the error messages returned over the tendermint abci interface, this is just about the errors sent into the contracts in the reply block. Also, if we don't catch the error case, these messages would probably propagate the the users anyway without passing through reply.

I prefer (A) for its simplicity.

[...]

PR #765, opened 2022-02-23
Redact errors returned in reply

ethanfrey:
Part 1 of #759 

The reply callback after executing a submessage no longer contains the entire error string on the failure case.
We only provide the error code and code space, which is guaranteed to be consistent (and part of the app hash)


ethanfrey:
@webmaster128 you might be interested in this as well.
This is a minimal first step.
In existing contracts, the actual error messages returned In reply never make it to end users.

x/wasm/keeper/msg_dispatcher.go
```diff
@@ -155,6 +159,22 @@ func (d MessageDispatcher) DispatchSubmessages(ctx sdk.Context, contractAddr sdk
...
+	}
+}
+
+func redactError(err error) string {
+	// FIXME: do we want to hardcode some constant string mappings here as well?
+	// Or better document them? (SDK error string may change on a patch release to fix wording)
+	// sdk/11 is out of gas
+	// sdk/5 is insufficient funds (on bank send)
+	codespace, code, _ := sdkerrors.ABCIInfo(err, false)
+	return fmt.Sprintf("Error: %s/%d", codespace, code)
```
webmaster128, line 175: Do we need this "Error: " prefix because this string is embedded in the `Err` case of the result, so this would come from the containing structure?

I wonder if `codespace: %s, code: %d` would be more helpful to understand where the values comes from. Having the variable names available is how devs interact with those values at the Tendermint RPC and CosmJS level.

webmaster128 (commented):
LGTM in general

x/wasm/keeper/msg_dispatcher.go
alpe: I agree with @webmaster128 that `codespace: %s, code: %d` is more readable for humans but `codespace/code` is easier to parse. In both cases, I would expect people to ask what the data means. 

IMHO, let's start with `codespace/code` to optimize for parsing. We can help the devs with some doc page for common codes.
Hardcoding some constants would add complexity on the contract side as they may have to match for certain constants as well as the "codespace/code" types.

"Error: " as a prefix is a redundant information. A prefix may make sense though if we want to have different structures/ versions to be parseable, like
`e1/sdk/11`,`e2/sdk/11/some-more-info`, `e3/sdk/11/some-other-info/123`, `e4/sdk/11/{"some":"json"}`

Getting a bit crazy on this, we could also make the whole error message json so that it is easy to parse, better readable for humans and simpler to extend. `{"codespace":"sdk", "code: "11"}`

x/wasm/keeper/submsg_test.go
```diff
@@ -275,15 +275,15 @@ func TestDispatchSubMsgErrorHandling(t *testing.T) {
 			subMsgError: true,
 			gasLimit:    &subGasLimit,
 			// uses same gas as call without limit (note we do not charge the 40k on reply)
-			resultAssertions: []assertion{assertGasUsed(79000, 79040), assertErrorString("insufficient funds")},
+			resultAssertions: []assertion{assertGasUsed(77500, 77600), assertErrorString("sdk/5")},
```
alpe, line 278: New error stings are cheaper 👍

alpe (approved):
This is small but effective change that solves one part of the issue. 💐 
I added some comments for improvements but the "error message text" one may got a bit beyond real requirements.

x/wasm/keeper/msg_dispatcher.go
webmaster128: Shouldn't parsability be a non-goal if we say there is no stability guarantee for the content of those error strings? The more we make them look like easy to parse, the mode users make assumptions on the format.

x/wasm/keeper/msg_dispatcher.go
```diff
@@ -155,6 +159,22 @@ func (d MessageDispatcher) DispatchSubmessages(ctx sdk.Context, contractAddr sdk
 	return rsp, nil
 }
 
+func logError(ctx sdk.Context, msg string, err error) {
+	logger := ctx.Logger()
+	if logger != nil {
```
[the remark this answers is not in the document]
ethanfrey: Sure, but without this it panic'd in several tests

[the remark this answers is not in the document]
ethanfrey: I did the moduleLogger(ctx) as you requested, without the check, and got:

```
--- FAIL: TestDispatchSubmessages (0.00s)
    --- FAIL: TestDispatchSubmessages/multiple_msg_-_last_reply_returned (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x1bffec1]

goroutine 1336 [running]:
testing.tRunner.func1.2({0x1e4c100, 0x30cfc50})
	/usr/local/go/src/testing/testing.go:1209 +0x36c
testing.tRunner.func1()
	/usr/local/go/src/testing/testing.go:1212 +0x3b6
panic({0x1e4c100, 0x30cfc50})
	/usr/local/go/src/runtime/panic.go:1047 +0x266
github.com/CosmWasm/wasmd/x/wasm/keeper.moduleLogger({{0x0, 0x0}, {0x2514dc0, 0xc0011f67b0}, {{0x0, 0x0}, {0x0, 0x0}, 0x0, {0x0, ...}, ...}, ...})
	/go/src/github.com/cosmwasm/wasmd/x/wasm/keeper/keeper.go:958 +0x181
github.com/CosmWasm/wasmd/x/wasm/keeper.MessageDispatcher.DispatchSubmessages({{_, _}, {_, _}}, {{0x0, 0x0}, {0x2514dc0, 0xc0011f67b0}, {{0x0, 0x0}, ...}, ...}, ...)
	/go/src/github.com/cosmwasm/wasmd/x/wasm/keeper/msg_dispatcher.go:137 +0xb18
github.com/CosmWasm/wasmd/x/wasm/keeper.TestDispatchSubmessages.func29(0xc0018ac820)
	/go/src/github.com/cosmwasm/wasmd/x/wasm/keeper/msg_dispatcher_test.go:352 +0x495
testing.tRunner(0xc0018ac820, 0xc00056ce80)
	/usr/local/go/src/testing/testing.go:1259 +0x230
created by testing.(*T).Run
	/usr/local/go/src/testing/testing.go:1306 +0x727
FAIL	github.com/CosmWasm/wasmd/x/wasm/keeper	15.259s
FAIL
```

x/wasm/keeper/msg_dispatcher.go
```diff
@@ -155,6 +159,15 @@ func (d MessageDispatcher) DispatchSubmessages(ctx sdk.Context, contractAddr sdk
 	return rsp, nil
 }
 
+func redactError(err error) string {
+	// FIXME: do we want to hardcode some constant string mappings here as well?
+	// Or better document them? (SDK error string may change on a patch release to fix wording)
+	// sdk/11 is out of gas
+	// sdk/5 is insufficient funds (on bank send)
+	codespace, code, _ := sdkerrors.ABCIInfo(err, false)
+	return fmt.Sprintf("codespace: %s, code: %d", codespace, code)
```
webmaster128, line 168: I think this is a pretty nice error string. Can be parsed if really needed but does not screem PARSE ME.

PR #770, opened 2022-03-02
Fix test setup and assertions

alpe:
Update for the module logger


ethanfrey (approved):
thanks. just saw this a bit late.

Issue #758, opened 2022-02-15
During migration from legacy 20 byte address a 'no history for address' error occurs 


iramiller:
In testing 0.23 on a testnet with existing contract instance an error is thrown due to the way kvstore addresses are generated.  Addresses are based on the current 32 byte length form, unfortunately some contracts were established with the prior 20 byte address form which is incompatible with this approach.

https://github.com/CosmWasm/wasmd/blob/de38ea008b07f6537e077039469e64648e6e9747/x/wasm/types/keys.go#L90

The above line always presumes the length will be 32 and it fails because it can't delete the history  here:

https://github.com/CosmWasm/wasmd/blob/de38ea008b07f6537e077039469e64648e6e9747/x/wasm/keeper/keeper.go#L578

This occurred during testing on Provenance (using wasmd 0.23) with the following:
```
provenanced tx wasm migrate \
    tp15qkf84uddc0zjpkjrn64rln75w40plrtcq4kp3 \
    111 \
    '{}' \
    --chain-id pio-testnet-1 \
    --node tcp://rpc-0.test.provenance.io:26657 \
    [ ... ]

Error: rpc error: code = InvalidArgument desc = recovered: no history for tp15qkf84uddc0zjpkjrn64rln75w40plrtcq4kp3
stack:
goroutine 764913 [running]:
runtime/debug.Stack()
    runtime/debug/stack.go:24 +0x65
```

iramiller:
One potential solution here is to use:
```
	addrLen := len(contractAddr)
	r := make([]byte, prefixLen+addrLen)
```
Length assertions could be inserted to require 20 or 32 byte address forms (which are the only two allowed) however if this is done additional refactoring will be required as the methods do not currently return errors and a panic isn't really desirable.

alpe:
Can you give more context why you need 20 byte contract addresses in your scenario?
A mix of both address sizes adds some complexity that I would avoid. You can either migrate the 20 byte addresses to 32 bytes or set the `ContractAddrLen` to 20 bytes.
The later one is not possible currently without changing code as it is a constant. Just open an issue/ PR for this.

The contract address is deterministic and based on codeID and contractID. Unfortunately the contractID is not persisted but can be found with some smarter than brute force algorithm for a migration.

the-frey:
By migration you mean at the contract level or at the go/binary level? We will run into this issue on Juno. as well (as I think we have 3-4 length 20 contracts instantiated when the chain first launched.

[...]

PR #772, opened 2022-03-02
Relaxed address strength restrictions for legacy 20 byte addresses

iramiller:
This PR upstreams the changes required for addressing #758 on the Provenance Blockchain.

By removing the assumption that a valid address would always be 32 bytes long per the current standard, the previous shorter 20 byte addresses can be used for accessing existing instances.  All new instances will conform to the new 32 byte form.

(PR opened by request of @ethanfrey, also CC @channa-figure )

### DIFF
--- a/x/wasm/types/keys.go
+++ b/x/wasm/types/keys.go
@@ -57,7 +57,8 @@ func GetContractStorePrefix(addr sdk.AccAddress) []byte {
 func GetContractByCreatedSecondaryIndexKey(contractAddr sdk.AccAddress, c ContractCodeHistoryEntry) []byte {
 	prefix := GetContractByCodeIDSecondaryIndexPrefix(c.CodeID)
 	prefixLen := len(prefix)
-	r := make([]byte, prefixLen+AbsoluteTxPositionLen+ContractAddrLen)
+	contractAddrLen := len(contractAddr)
+	r := make([]byte, prefixLen+AbsoluteTxPositionLen+contractAddrLen)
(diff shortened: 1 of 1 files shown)


the-frey:
Thanks @iramiller - keen to get this tested at our end :)

faddat:
Many thanks @iramiller

iramiller:
We have been running with this change on our testnet... it was all that we required based on our testing... we have contracts with both length addresses now.

alpe:
LGTM. 
In the tests we use 32 byte addresses. It is an important design decision that 20 byte addresses should work, too. I am afraid that without code comments and/ or tests it would be easy to break in the future. If you have some capacity, please add some tests so that we are on the safe side. 🙏

ethanfrey:
Any more testing before we merge this?

iramiller:
Some more comments/tests coverage makes sense.  

When I opened this PR I only wanted the two initial commits that we are actually running on Provenance `pio-testnet-1` for clarity so as to not imply there were more pieces that we have verified directly than what we are actually using.

@channa-figure and I discussed the unit test angle this morning and it should be straight forward to add some extra comments and a test that asserts the 20 byte form is accepted here as well as the 32 byte version in `keys_test.go`)

channa-figure:
I added a new test and a test case to the keys_test.go file.  I also left a comment with in the test for a little clarification.

alpe (approved):
Very nice. Thanks for the contribution and the additional test. 🥇

PR #773, opened 2022-03-02
Make utils methods public

pinosu:
Make GetAllCodes and GetAllContracts methods public


alpe (commented):
Looks good. Just curious, why are there changes to go.mod/sum ?

pinosu:
My ide was saying that vendor folder is out of sync so I had to run `go mod vendor` and `go mod tidy` . So they got updated.

PR #775, opened 2022-03-07
Redact query errors

### LINKED ISSUE #759: Return minimal info on errors returned to contracts

ethanfrey:
Follow up from #765 

Closes #759

### DIFF
--- a/x/wasm/keeper/query_plugins.go
+++ b/x/wasm/keeper/query_plugins.go
@@ -60,6 +60,12 @@ func (q QueryHandler) Query(request wasmvmtypes.QueryRequest, gasLimit uint64) (
 	if ok := errors.As(err, &noSuchContract); ok {
 		return res, wasmvmtypes.NoSuchContract{Addr: noSuchContract.Addr}
 	}
+	// Otherwise redact all (we can theoretically redact less in the future)
+	if err != nil {
+		// Issue #759 - we don't return error string for worries of non-determinism
+		// moduleLogger(ctx).Info("Redacting submessage error", "cause", err)
+		err = errors.New(redactError(err))
+	}
 	return res, err
 }
 


alpe (approved):
LGTM 👍 
The error path was covered by a test already

x/wasm/keeper/query_plugins.go
```diff
@@ -60,6 +60,12 @@ func (q QueryHandler) Query(request wasmvmtypes.QueryRequest, gasLimit uint64) (
 	if ok := errors.As(err, &noSuchContract); ok {
 		return res, wasmvmtypes.NoSuchContract{Addr: noSuchContract.Addr}
 	}
+	// Otherwise redact all (we can theoretically redact less in the future)
+	if err != nil {
+		// Issue #759 - we don't return error string for worries of non-determinism
+		// moduleLogger(ctx).Info("Redacting submessage error", "cause", err)
```
larry0x, line 66: @alpe @webmaster128 Is it possible to have this error being logged?

alpe: No problem. I have opened an issue to track the work

webmaster128: https://github.com/CosmWasm/wasmd/issues/1593

larry0x: Thank you

PR #776, opened 2022-03-07
Verify Stargate Queries handles properly

### LINKED ISSUE #762: Better tests for StargateQuery
People seem to be using this feature more recently and it has very limited testing in our code.

It is such a generic process, but we should ensure it works as expected for more than one example. Let's provide better test cases and ensure this works for various cases, such as app queries as well as raw queries (and maybe even getting Merkle Proofs for ics store).

ethanfrey:
Closes #762

### DIFF
--- a/x/wasm/keeper/query_plugins.go
+++ b/x/wasm/keeper/query_plugins.go
@@ -265,15 +265,15 @@ func IBCQuerier(wasm contractMetaDataSource, channelKeeper types.ChannelKeeper) 
 }
 
 var queryDenyList = []string{
-	"/cosmos.tx",
-	"/cosmos.base.tendermint",
+	"/cosmos.tx.",
+	"/cosmos.base.tendermint.",
 }
 
 func StargateQuerier(queryRouter GRPCQueryRouter) func(ctx sdk.Context, request *wasmvmtypes.StargateQuery) ([]byte, error) {
 	return func(ctx sdk.Context, msg *wasmvmtypes.StargateQuery) ([]byte, error) {
 		for _, b := range queryDenyList {
-			if strings.Contains(msg.Path, b) {
-				return nil, wasmvmtypes.UnsupportedRequest{Kind: fmt.Sprintf("'%s' path is not allowed from the contract", msg.Path)}
+			if strings.HasPrefix(msg.Path, b) {
+				return nil, wasmvmtypes.UnsupportedRequest{Kind: "path is not allowed from the contract"}
 			}
 		}
 
--- a/x/wasm/keeper/query_plugins.go
+++ b/x/wasm/keeper/query_plugins.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"strings"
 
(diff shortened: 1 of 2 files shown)


x/wasm/keeper/query_plugins.go
```diff
@@ -263,8 +264,19 @@ func IBCQuerier(wasm contractMetaDataSource, channelKeeper types.ChannelKeeper)
...
+var queryDenyList = []string{
+	"/cosmos.tx",
+	"/cosmos.base.tendermint",
+}
+
 func StargateQuerier(queryRouter GRPCQueryRouter) func(ctx sdk.Context, request *wasmvmtypes.StargateQuery) ([]byte, error) {
 	return func(ctx sdk.Context, msg *wasmvmtypes.StargateQuery) ([]byte, error) {
+		for _, b := range queryDenyList {
+			if strings.Contains(msg.Path, b) {
+				return nil, wasmvmtypes.UnsupportedRequest{Kind: fmt.Sprintf("'%s' path is not allowed from the contract", msg.Path)}
```
alpe, line 276: I want to be more sensible to not return user input in errors. Would 'path is not allowed from the contract' be good enough? The caller knows the path already.

x/wasm/keeper/reflect_test.go
```diff
@@ -382,6 +382,62 @@ func TestReflectStargateQuery(t *testing.T) {
...
+		Chain: &ChainQuery{Request: &protoRequest},
+	})
+	require.NoError(t, err)
+
+	// make a query on the chain, should be blacklisted
+	_, err = keeper.QuerySmart(ctx, contractAddr, protoQueryBz)
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "path is not allowed from the contract")
+}
+
```
alpe, line 440: To cover the deny list completely we should have a  `/cosmos.tx` query as well. A table test structure may be more readable but this is ok as well.

alpe (approved):
This solves the issue already. 
@webmaster128  had some good points that would improve the solution. I added some comments for bonus points.

x/wasm/keeper/query_plugins.go
ethanfrey: I updated this. The caller (user) will not know by default, as this is called from a contract. But I did redact.

Also, 3 lines lower we already had this code:
`return nil, wasmvmtypes.UnsupportedRequest{Kind: fmt.Sprintf("No route to query '%s'", msg.Path)}`

Which I wrote March 2021. Should we change that as well?

x/wasm/keeper/reflect_test.go
ethanfrey: We do on line 409.

```go
	// now, try to build a protobuf query
	protoRequest := wasmvmtypes.QueryRequest{
		Stargate: &wasmvmtypes.StargateQuery{
			Path: "/cosmos.tx.v1beta1.Service/GetTx",
			Data: []byte{},
		},
	}
```

ethanfrey:
Addressed comments. Waiting for CI

PR #779, opened 2022-03-09
Prepare 0.24

ethanfrey:
- [x] Fixes broken test on main (due to merging 3 PRs together)
- [x] Updates CHANGELOG


x/wasm/keeper/query_plugins.go
```diff
@@ -56,18 +56,26 @@ func (q QueryHandler) Query(request wasmvmtypes.QueryRequest, gasLimit uint64) (
...
-		return res, wasmvmtypes.NoSuchContract{Addr: noSuchContract.Addr}
+		err = wasmvmtypes.NoSuchContract{Addr: noSuchContract.Addr}
 	}
-	// Otherwise redact all (we can theoretically redact less in the future)
-	if err != nil {
+
+	// Redact all that are not system errors
+	// SystemErrors must be created in x/wasm and we can ensure determinism
+	// (we can theoretically redact less in the future, but this is a first step to safety)
+	if wasmvmtypes.ToSystemError(err) == nil {
```
alpe, line 73: Would it make sense to exclude system errors in https://github.com/CosmWasm/wasmd/blob/master/x/wasm/keeper/msg_dispatcher.go#L139 as well?

alpe (approved):
LGTM, one question in the comments.

❤️  for the changelog update

x/wasm/keeper/query_plugins.go
ethanfrey: Good point.

I can actually generalise the redactError logic

PR #792, opened 2022-03-22
fix: close iterators

albertchon:
Iterators should be closed after use. 

I'd also potentially recommend renaming `GetContractState(ctx sdk.Context, contractAddress sdk.AccAddress) sdk.Iterator` to `GetContractStateIterator(ctx sdk.Context, contractAddress sdk.AccAddress) sdk.Iterator` so it's clear that clients should close the iterator themselves or simply change the function signature to take a callback so the usage pattern is standardized.

### DIFF
--- a/x/wasm/keeper/genesis.go
+++ b/x/wasm/keeper/genesis.go
@@ -119,6 +119,8 @@ func ExportGenesis(ctx sdk.Context, keeper *Keeper) *types.GenesisState {
 			ContractState:   state,
 		})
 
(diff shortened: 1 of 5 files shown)


alpe:
Thanks a lot for your contribution! 💪 . Good 👀 !
I agree with your proposal to modify `GetContractState` and standardize using callbacks. I will do a PR

alpe (approved):
Nice 💯

PR #793, opened 2022-03-23
interchain accounts

faddat:
Hi, this PR upgrades wasmd to ibcv3


faddat:
I think this one is:

* up to date
* ready to go

Could you let me know if this is something that you're interested in reviewing / accepting?

Thanks,

-Jacob

ethanfrey:
Awesome! I was wondering about pulling this out of the 0.46 upgrade and just noticed this PR.
I was off since the end of March, but I will look into this PR this week and give you full feedback

ethanfrey:
Note: Resolves #806

ethanfrey (commented):
Thank you for this PR.

I would pull it out into a few different ones. I see:

1. some osmosis-specific stuff that should be removed
2. Renaming lots of WasmApp members to make them public
3. Importing ibcv3 and exposing SetChannel

I see (2) and (3) as independent and not necessarily both needed.

I will try to cut out the ibcv3 commits alone and see if I can make a new PR.

You could also pull out the "make all app members public" code into a separate PR. I have no strong opinion on this either way, but it does make some testing easier. I would like @alpe to give feedback on that part.

ethanfrey:
Closing in favor of #824 and #837

PR #801, opened 2022-04-11
Create link to SECURITY.md in other repo

webmaster128:


alpe (commented):
IMHO it makes sense to keep the `Supported Versions` section here. WDYT?

webmaster128:
Good question. We can also create an overview like this:

Bildschirmfoto 2022-04-11 um 13 19 54

(taken from https://github.com/CosmWasm/advisories#supported-projects-and-version)

If it's not in one place but spread across repos, it will get outdated very quickly.

In https://github.com/CosmWasm/cw-plus/blob/main/SECURITY.md it's structured as link + supported version. But all those supported versions sections seem to be very unrealistic anyways as people build on this software no matter which version name we use for it.

webmaster128:
I brought back the Supported versions section (see last 2 commits). We can try it this way.

alpe (approved):
LGTM 👏

PR #802, opened 2022-04-11
Unpack contract details from genesis

alpe:
On state import the contract `ContractInfoExtension` was not unpacked proper. This PR adds some unpacker callbacks to solve this.


pinosu (approved):
lgtm 💯

PR #808, opened 2022-04-17
Fix SudoContractProposal and ExecuteContractProposal

the-frey:
Follow up fix for #730

Add missing `SudoContractProposal` and `ExecuteContractProposal` where it was missinng in types/codec and basic validation unit test.

Could be some more to this that I've missed.

### DIFF
--- a/x/wasm/types/codec.go
+++ b/x/wasm/types/codec.go
@@ -17,12 +17,14 @@ func RegisterLegacyAminoCodec(cdc *codec.LegacyAmino) { //nolint:staticcheck
 	cdc.RegisterConcrete(&MsgMigrateContract{}, "wasm/MsgMigrateContract", nil)
 	cdc.RegisterConcrete(&MsgUpdateAdmin{}, "wasm/MsgUpdateAdmin", nil)
 	cdc.RegisterConcrete(&MsgClearAdmin{}, "wasm/MsgClearAdmin", nil)
+
 	cdc.RegisterConcrete(&PinCodesProposal{}, "wasm/PinCodesProposal", nil)
 	cdc.RegisterConcrete(&UnpinCodesProposal{}, "wasm/UnpinCodesProposal", nil)
-
 	cdc.RegisterConcrete(&StoreCodeProposal{}, "wasm/StoreCodeProposal", nil)
 	cdc.RegisterConcrete(&InstantiateContractProposal{}, "wasm/InstantiateContractProposal", nil)
 	cdc.RegisterConcrete(&MigrateContractProposal{}, "wasm/MigrateContractProposal", nil)
+	cdc.RegisterConcrete(&SudoContractProposal{}, "wasm/SudoContractProposal", nil)
+	cdc.RegisterConcrete(&ExecuteContractProposal{}, "wasm/ExecuteContractProposal", nil)
 	cdc.RegisterConcrete(&UpdateAdminProposal{}, "wasm/UpdateAdminProposal", nil)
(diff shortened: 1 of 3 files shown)


the-frey:
Will also be running some manual tests tomorrow to check if this resolves the issues we've seen with gov/sudo.

the-frey:
Update: we have run this on a testnet with a sudo contract and successfully moved funds via a gov prop. Our auto-CI hook that was failing with 0.25.0 also passes when we substitute in this branch, so we're pretty confident it's the fix.

maurolacy (approved):
LGTM. Perhaps @alpe wants to take a look.

x/wasm/types/codec.go
```diff
@@ -44,6 +46,8 @@ func RegisterInterfaces(registry types.InterfaceRegistry) {
 		&StoreCodeProposal{},
 		&InstantiateContractProposal{},
 		&MigrateContractProposal{},
+		&SudoContractProposal{},
+		&ExecuteContractProposal{},
```
pinosu, line 50: 👍

x/wasm/types/proposal_test.go
```diff
@@ -317,6 +317,61 @@ func TestValidateMigrateContractProposal(t *testing.T) {
 	}
 }
 
+func TestValidateSudoContractProposal(t *testing.T) {
```
pinosu, line 320: nice test!

x/wasm/types/proposal_test.go
```diff
@@ -317,6 +317,61 @@ func TestValidateMigrateContractProposal(t *testing.T) {
...
+			err := spec.src.ValidateBasic()
+			if spec.expErr {
+				require.Error(t, err)
+			} else {
+				require.NoError(t, err)
+			}
+		})
+	}
+}
+
```
pinosu, line 374: Could you add some unit tests also for **ExecuteContractProposal** ?

x/wasm/types/test_fixtures.go
```diff
@@ -257,6 +257,25 @@ func MigrateContractProposalFixture(mutators ...func(p *MigrateContractProposal)
 	return p
 }
 
+func SudoContractProposalFixture(mutators ...func(p *SudoContractProposal)) *SudoContractProposal {
+	const (
+		contractAddr = "cosmos14hj2tavq8fpesdwxxcu44rty3hh90vhujrvcmstl4zr3txmfvw9s4hmalr"
+		anyAddress   = "cosmos1qyqszqgpqyqszqgpqyqszqgpqyqszqgpqyqszqgpqyqszqgpqyqs2m6sx4"
```
pinosu, line 263: unused

pinosu (commented):
lgtm. 👍  
I added just some minor comments.
Also it would be nice if you could update the description mentioning that also ExecuteContractProposal was added.

x/wasm/types/test_fixtures.go
the-frey: SudoContractProposalFixture is used in the test above?

x/wasm/types/proposal_test.go
the-frey: Added a unit test for it 👍

the-frey:
@pinosu have updated the description and some additional test cases as you suggested.

webmaster128 (approved):

ethanfrey:
Good find. I have a mental block with amino. Can't believe we **still** need to register all this stuff.

PR #809, opened 2022-04-19
Remove MaxWasmSize restriction in ValidateBasic

bigs:
Spoke with Simon @ Confio on Discord about this. In the process of developing a Cosmos chain with CosmWasm, we found that `wasmd`'s `MaxWasmSize` is constantly defined and thus, potentially at odds with the module params that a chain might provide. It seems as though we should simply rely on the param and not on this constant. 

This change simply removes the constant and corresponding check. I am definitely up to other solutions, but wanted to surface it here so it gets tracked/discussed!

### DIFF
--- a/x/wasm/types/validation.go
+++ b/x/wasm/types/validation.go
@@ -5,17 +5,18 @@ import (
 )
 
 const (
-	MaxWasmSize = 500 * 1024
-
 	// MaxLabelSize is the longest label that can be used when Instantiating a contract
 	MaxLabelSize = 128
 )
+
+// MaxWasmSize is the largest a compiled contract code can be when storing code on chain
+var MaxWasmSize uint64 = 800 * 1024
 
 func validateWasmCode(s []byte) error {
 	if len(s) == 0 {
 		return sdkerrors.Wrap(ErrEmpty, "is required")
 	}
-	if len(s) > MaxWasmSize {
+	if uint64(len(s)) > MaxWasmSize {
 		return sdkerrors.Wrapf(ErrLimit, "cannot be longer than %d bytes", MaxWasmSize)
 	}
 	return nil
--- a/x/wasm/alias.go
+++ b/x/wasm/alias.go
@@ -17,7 +17,6 @@ const (
 	TStoreKey                       = types.TStoreKey
 	QuerierRoute                    = types.QuerierRoute
 	RouterKey                       = types.RouterKey
-	MaxWasmSize                     = types.MaxWasmSize
 	MaxLabelSize                    = types.MaxLabelSize
 	WasmModuleEventType             = types.WasmModuleEventType
 	AttributeKeyContractAddr        = types.AttributeKeyContractAddr
--- a/x/wasm/keeper/keeper.go
+++ b/x/wasm/keeper/keeper.go
@@ -139,12 +139,6 @@ func (k Keeper) getInstantiateAccessConfig(ctx sdk.Context) types.AccessType {
 	return a
 }
 
-func (k Keeper) GetMaxWasmCodeSize(ctx sdk.Context) uint64 {
-	var a uint64
-	k.paramSpace.Get(ctx, types.ParamStoreKeyMaxWasmCodeSize, &a)
-	return a
-}
-
 // GetParams returns the total set of wasm parameters.
 func (k Keeper) GetParams(ctx sdk.Context) types.Params {
 	var params types.Params
@@ -164,7 +158,7 @@ func (k Keeper) create(ctx sdk.Context, creator sdk.AccAddress, wasmCode []byte,
 	if !authZ.CanCreateCode(k.getUploadAccessConfig(ctx), creator) {
 		return 0, sdkerrors.Wrap(sdkerrors.ErrUnauthorized, "can not create code")
 	}
-	wasmCode, err = uncompress(wasmCode, k.GetMaxWasmCodeSize(ctx))
+	wasmCode, err = uncompress(wasmCode, types.MaxWasmSize)
 	if err != nil {
 		return 0, sdkerrors.Wrap(types.ErrCreateFailed, err.Error())
 	}
@@ -206,7 +200,7 @@ func (k Keeper) storeCodeInfo(ctx sdk.Context, codeID uint64, codeInfo types.Cod
 }
 
 func (k Keeper) importCode(ctx sdk.Context, codeID uint64, codeInfo types.CodeInfo, wasmCode []byte) error {
-	wasmCode, err := uncompress(wasmCode, k.GetMaxWasmCodeSize(ctx))
+	wasmCode, err := uncompress(wasmCode, types.MaxWasmSize)
 	if err != nil {
 		return sdkerrors.Wrap(types.ErrCreateFailed, err.Error())
 	}
--- a/x/wasm/simulation/params.go
+++ b/x/wasm/simulation/params.go
@@ -28,11 +28,6 @@ func ParamChanges(r *rand.Rand, cdc codec.Codec) []simtypes.ParamChange {
 				return fmt.Sprintf("%q", params.CodeUploadAccess.Permission.String())
 			},
 		),
-		simulation.NewSimParamChange(types.ModuleName, string(types.ParamStoreKeyMaxWasmCodeSize),
-			func(r *rand.Rand) string {
-				return fmt.Sprintf(`"%d"`, params.MaxWasmCodeSize)
-			},
-		),
 	}
 }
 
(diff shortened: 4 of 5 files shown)


ethanfrey:
Good eye.

ValidateBasic does not have access to state, so we cannot check against params. If we remove this check here, I would like to ensure that it will actually be checked somewhere in CheckTx. Otherwise, you can get honest validators to fill up the chain with arbitrary large garbage data.

Removing it is not sufficient, we need an alternative. One idea is using a var rather than const and making this definable compile time (eg in app.go or using a go linker flag)

x/wasm/types/validation.go
```diff
@@ -5,8 +5,6 @@ import (
 )
 
 const (
-	MaxWasmSize = 500 * 1024
```
ethanfrey, line 8: Removing is dangerous.
Making this `var` could allow an app importing this to override the value at compile time, yet maintain a safety guarantee in CheckTX.

This is not the same as a module param, but better than a const.

alpe:
>  Otherwise, you can get honest validators to fill up the chain with arbitrary large garbage data.

Good point! 
A custom ante handler that reads the params would be an option, too.

bigs:
The param is ultimately read in the [uncompress](https://github.com/CosmWasm/wasmd/blob/main/x/wasm/keeper/ioutil.go#L18) function which is used in the Keeper's [create](https://github.com/CosmWasm/wasmd/blob/main/x/wasm/keeper/keeper.go#L167) and [importCode](https://github.com/CosmWasm/wasmd/blob/main/x/wasm/keeper/keeper.go#L209) functions.

ash-burnt:
The weird thing here is that there is a hardcoded variable that exists, and also a param for the same value. The value is exposed to governance, but not actually honored by governance. It seems like this should be one or the other, not both.

ethanfrey:
> The param is ultimately read in the [uncompress](https://github.com/CosmWasm/wasmd/blob/main/x/wasm/keeper/ioutil.go#L18) function which is used in the Keeper's [create](https://github.com/CosmWasm/wasmd/blob/main/x/wasm/keeper/keeper.go#L167) and [importCode](https://github.com/CosmWasm/wasmd/blob/main/x/wasm/keeper/keeper.go#L209) functions. This (more configurable) redundancy was my motivation to remove it from the message itself. I noticed my transaction was failing client side and never even sending to a validator to be rejected.

Yes, and eg. `create` is only called in DeliverTx, not CheckTx. We want a sanity check before including 5 MB invalid wasm tx in the blockchain in the first place (which ValidateBasic will do, but nothing else, as not x/wasm handler logic is called in CheckTx)

I agree we should only have one value. I would make it a compile-time variable and remove the param completely. (And document this along with migration for chains when they upgrade)

bigs:
Ah, makes sense. I think I had my wires a bit crossed, there! Compile time seems reasonable as does a configuration object passed to `NewAppModule`, perhaps? Happy to take this on when we come to something, sensible.

bigs:
@alpe @ethanfrey @ash-burnt Digging in a bit further, it seems like there are a few options:

1. Change the const to a `var`, eliminate the conflicting module param, and respect the `var` in `ValidateBasic`. It would probably also entail removing the check in the `Keeper`, which seems redundant, but I'm not entirely sure on that.
2. Implement an additional `AnteDecorator` in `x/wasm/keeper` that accepts the wasmd module's param `Subspace` as an argument and validates messages storing contracts in a transaction accordingly. We would then update `app`'s `NewAnteHandler` to optionally accept a `Subspace` in `HandlerOptions`. When present, it can add our decorator to the stack. Similarly, my gut tells me we could then do away with the `Keeper` validation.

I'm relatively new to the ecosystem, so I won't claim to know which is best. Option 1 is certainly simplest and requires minimal modification to anything. Option 2 allows us to modulate this parameter via governance. For example, the `wasm-opt` optimized byte code for cw721-metadata-onchain comes out to 253KB which is already pushing half of the statically defined maximum. I could see a chain starting with a maximum around 500KB and then revising it upwards. Doing so via governance rather than cutting a completely new version of the application could save some headaches!

ethanfrey:
>  I could see a chain starting with a maximum around 500KB and then revising it upwards. 

I think we advise 600KB or 1200KB as default parameters (unaware the const was still there).

I would have chains just start with something like 800KB-1000KB to start, and then if this really is an issue and some contract creator convinces them this is important, they just include this in the next upgrade. A contract creator should realise this well before production (when building code, running on testnet, etc) and I don't see a quick turnaround as critical.

Most chains do upgrades every 1-3 months, so that seems reasonable.

TL;DR all for the simple solution - using `var` and updating default to 800 KB or so

bigs:
@ethanfrey Pushed up a change that should address it the way we discussed (force-pushed for cleaner history, which auto-closed the branch.) Let me know what you think! Also, for posterity's sake [here is a diff](https://gist.github.com/bigs/1340f1c69ba2464527b69d02e7b6e1c4) that implements the params & AnteHandler based approach.

x/wasm/types/validation.go
```diff
@@ -5,17 +5,18 @@ import (
...
-
 	// MaxLabelSize is the longest label that can be used when Instantiating a contract
 	MaxLabelSize = 128
 )
 
+// MaxWasmSize is the largest a compiled contract code can be when storing code on chain
+var MaxWasmSize uint64 = 800 * 1024
```
ethanfrey, line 13: This is good.
Can also be larger if you wish (param was upped to 1200 * 1024 earlier).

But I think 800 or 900 is a good default

ethanfrey: But... why make uint64 and not int? Since we compare it to length which returns an int

ethanfrey (commented):
Good stuff. I agree with this.

Can you add a CHANGELOG entry? And update the description to add:

"Addresses #813"

and

"The WasmCodeSize parameter is no longer used. You can set it by overriding `x/wasm/types/MaxWasmSize` at compile time (in app.go). The default is now set to 800kB but any chain can easily configure when doing a binary upgrade."

You can expand the text, but that is the rough idea that should be there, and linked from the CHANGELOG. I will merge once those cleanup tasks are done. Thanks for the code contribution.

x/wasm/types/validation.go
bigs: I made this call because one of its uses requires a `uint64` and with split use cases, I went unsigned as a length cannot be negative. I truly don't mind changing it, though!

bigs:
@ethanfrey Should be sorted!

x/wasm/types/validation.go
```diff
@@ -5,12 +5,13 @@ import (
...
 
 const (
-	MaxWasmSize = 500 * 1024
-
 	// MaxLabelSize is the longest label that can be used when Instantiating a contract
 	MaxLabelSize = 128
 )
 
+// MaxWasmSize is the largest a compiled contract code can be when storing code on chain
+var MaxWasmSize int = 800 * 1024
```
alpe, line 13: Just curious why you increased the value to 800kb? Do you have a contract example with this size? We had some internal discussion for [tgrade](https://github.com/confio/tgrade/issues/239#issuecomment-1109603114) yesterday where I looked at some of our contracts and max was 555kb. 

```suggestion
var MaxWasmSize int = 777 * 1024
```

alpe (changes requested):
Nice work! Some tests are failing. Can you run `make test`, please?

Would be great if can add some doc to
https://github.com/CosmWasm/wasmd#runtime-flags
for this new flag

x/wasm/types/validation.go
alpe: ah, Ethan suggested 800kb. We should have coordinated better. 🤷  It would help with tgrade if you can make it 777kb

bigs:
Ah wow -- sorry for moving too quickly. I need to update the protobuf definitions for the params. That broke all the tests!

x/wasm/types/validation.go
ethanfrey: 777kb is an odd number. 
But tgrade can easily override this default, so maybe best to use 800 and verify we can change it in the imported one

ethanfrey:
Hmmm... is it possible to enable CI runs on this? I didn't realise the tests hadn't all run until @alpe pointed it out.

ethanfrey:
BTW, you can look at https://github.com/CosmWasm/wasmd/pull/822 for example how to update the README and CHANGELOG. (Just add one more line to the README section I created)

alpe:
@bigs: we have some dependencies on this feature, so I jumped in this morning to rebased your work on top of current main in a new PR (#826) and fixed tests and protobuf. I hope you don't mind. 

Can you take a look at the new PR, if this still makes sense for you? Thanks a lot for your work!

bigs:
Totally! I was held up the last two days with other things (least of which
is not the fact that my work laptop has an Arm processor which the docker
images don’t work with.) I’ll review and we can get this out the door.

bigs:
Closing in favor of #826

PR #816, opened 2022-04-25
Add WASM files to state sync

assafmo:
This helps state sync to track the WASM files too, as they're not in the IAVL.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -682,6 +682,11 @@ func NewWasmApp(
 	app.scopedIBCKeeper = scopedIBCKeeper
 	app.scopedTransferKeeper = scopedTransferKeeper
 	app.scopedWasmKeeper = scopedWasmKeeper
+
+	app.SnapshotManager().RegisterExtensions(
+		wasm.NewWasmSnapshotter(filepath.Join(wasmDir, "wasm", "state", "wasm")),
+	)
+
 	return app
 }
 
--- a/x/wasm/wasm_snapshotter.go
+++ b/x/wasm/wasm_snapshotter.go
@@ -1,0 +1,150 @@
+package wasm
+
+import (
+	"io"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"regexp"
+	"sort"
+	"strings"
+
+	snapshot "github.com/cosmos/cosmos-sdk/snapshots/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	protoio "github.com/gogo/protobuf/io"
+)
+
+/*
+API to implement:
+
+// Snapshotter is something that can create and restore snapshots, consisting of streamed binary
+// chunks - all of which must be read from the channel and closed. If an unsupported format is
+// given, it must return ErrUnknownFormat (possibly wrapped with fmt.Errorf).
+type Snapshotter interface {
+	// Snapshot writes snapshot items into the protobuf writer.
+	Snapshot(height uint64, protoWriter protoio.Writer) error
+
+	// Restore restores a state snapshot from the protobuf items read from the reader.
+	// If the ready channel is non-nil, it returns a ready signal (by being closed) once the
+	// restorer is ready to accept chunks.
+	Restore(height uint64, format uint32, protoReader protoio.Reader) (SnapshotItem, error)
+}
+
+// ExtensionSnapshotter is an extension Snapshotter that is appended to the snapshot stream.
+// ExtensionSnapshotter has an unique name and manages it's own internal formats.
+type ExtensionSnapshotter interface {
+	Snapshotter
+
+	// SnapshotName returns the name of snapshotter, it should be unique in the manager.
+	SnapshotName() string
+
+	// SnapshotFormat returns the default format the extension snapshotter use to encode the
+	// payloads when taking a snapshot.
+	// It's defined within the extension, different from the global format for the whole state-sync snapshot.
+	SnapshotFormat() uint32
+
+	// SupportedFormats returns a list of formats it can restore from.
+	SupportedFormats() []uint32
+}
+*/
+
+type WasmSnapshotter struct {
+	wasmDirectory string
+}
+
+func NewWasmSnapshotter(wasmDirectory string) *WasmSnapshotter {
+	return &WasmSnapshotter{
+		wasmDirectory,
+	}
+}
+
+func (ws *WasmSnapshotter) SnapshotName() string {
+	return "WASM Files Snapshot"
+}
+
+func (ws *WasmSnapshotter) SnapshotFormat() uint32 {
+	return 1
+}
+
+func (ws *WasmSnapshotter) SupportedFormats() []uint32 {
+	return []uint32{1}
+}
+
+var wasmFileNameRegex = regexp.MustCompile(`^[a-f0-9]{64}$`)
+
+func (ws *WasmSnapshotter) Snapshot(height uint64, protoWriter protoio.Writer) error {
+	wasmFiles, err := ioutil.ReadDir(ws.wasmDirectory)
+	if err != nil {
+		return err
+	}
+
+	// In case snapshotting needs to be deterministic
+	sort.SliceStable(wasmFiles, func(i, j int) bool {
+		return strings.Compare(wasmFiles[i].Name(), wasmFiles[j].Name()) < 0
+	})
+
+	for _, wasmFile := range wasmFiles {
+		if !wasmFileNameRegex.MatchString(wasmFile.Name()) {
+			continue
+		}
+
+		wasmFilePath := filepath.Join(ws.wasmDirectory, wasmFile.Name())
+		wasmBytes, err := ioutil.ReadFile(wasmFilePath)
+		if err != nil {
+			return err
+		}
+
+		// snapshotItem is 64 bytes of the file name, then the actual WASM bytes
+		snapshotItem := append([]byte(wasmFile.Name()), wasmBytes...)
+
+		snapshot.WriteExtensionItem(protoWriter, snapshotItem)
+	}
+
+	return nil
+}
+
+func (ws *WasmSnapshotter) Restore(
+	height uint64, format uint32, protoReader protoio.Reader,
+) (snapshot.SnapshotItem, error) {
+	if format != 1 {
+		return snapshot.SnapshotItem{}, snapshot.ErrUnknownFormat
+	}
+
+	// Create .compute directory if it doesn't exist already
+	err := os.MkdirAll(ws.wasmDirectory, os.ModePerm)
+	if err != nil {
+		return snapshot.SnapshotItem{}, sdkerrors.Wrapf(err, "failed to create directory '%s'", ws.wasmDirectory)
+	}
+
+	for {
+		item := &snapshot.SnapshotItem{}
+		err = protoReader.ReadMsg(item)
+		if err == io.EOF {
+			break
+		} else if err != nil {
+			return snapshot.SnapshotItem{}, sdkerrors.Wrap(err, "invalid protobuf message")
+		}
+
+		payload := item.GetExtensionPayload()
+		if payload == nil {
+			return snapshot.SnapshotItem{}, sdkerrors.Wrap(err, "invalid protobuf message")
+		}
+
+		// snapshotItem is 64 bytes of the file name, then the actual WASM bytes
+		if len(payload.Payload) < 64 {
+			return snapshot.SnapshotItem{}, sdkerrors.Wrapf(err, "wasm snapshot must be at least 64 bytes, got %v bytes", len(payload.Payload))
+		}
+
+		wasmFileName := string(payload.Payload[0:64])
+		wasmBytes := payload.Payload[64:]
+
+		wasmFilePath := filepath.Join(ws.wasmDirectory, wasmFileName)
+
+		err = ioutil.WriteFile(wasmFilePath, wasmBytes, 0664 /* -rw-rw-r-- */)
+		if err != nil {
+			return snapshot.SnapshotItem{}, sdkerrors.Wrapf(err, "failed to write wasm file '%v' to disk", wasmFilePath)
+		}
+	}
+
+	return snapshot.SnapshotItem{}, nil
+}


app/app.go
```diff
@@ -682,6 +682,11 @@ func NewWasmApp(
 	app.scopedIBCKeeper = scopedIBCKeeper
 	app.scopedTransferKeeper = scopedTransferKeeper
 	app.scopedWasmKeeper = scopedWasmKeeper
+
+	app.SnapshotManager().RegisterExtensions(
+		wasm.NewWasmSnapshotter(filepath.Join(wasmDir, "wasm", "state", "wasm")),
```
assafmo, line 687: There's probably a nicer way of doing this, but honestly this directory structure is buried so deep inside https://github.com/CosmWasm/cosmwasm/blob/be4d94a8d73ceab359226b8ea0a5b4be032a07e6/packages/vm/src/cache.rs#L105 so maybe it's better to leave it that way for now. Just a thought.

ethanfrey: Yeah, that top level wasm comes from wasmd from long ago.
I have thought of changing it, but it would be a very painful breaking change to all running nodes.

ethanfrey: Actually, notice `wasmDir := filepath.Join(homePath, "wasm")` on line 448 or so. That already contains the first wasm.

It should be `filepath.Join(wasmDir, "state", "wasm")`

x/wasm/wasm_snapshotter.go
```diff
@@ -0,0 +1,150 @@
...
+
+var wasmFileNameRegex = regexp.MustCompile(`^[a-f0-9]{64}$`)
+
+func (ws *WasmSnapshotter) Snapshot(height uint64, protoWriter protoio.Writer) error {
+	wasmFiles, err := ioutil.ReadDir(ws.wasmDirectory)
+	if err != nil {
+		return err
+	}
+
+	// In case snapshotting needs to be deterministic
```
ethanfrey, line 81: Nice call.

x/wasm/wasm_snapshotter.go
```diff
@@ -0,0 +1,150 @@
...
+			continue
+		}
+
+		wasmFilePath := filepath.Join(ws.wasmDirectory, wasmFile.Name())
+		wasmBytes, err := ioutil.ReadFile(wasmFilePath)
+		if err != nil {
+			return err
+		}
+
+		// snapshotItem is 64 bytes of the file name, then the actual WASM bytes
```
ethanfrey, line 97: This works. But maybe we want a real (eg. more future-proof) format.

Eg. like a Protobuf object with two fields and a type. Especially if we ever want to sync anything else

x/wasm/wasm_snapshotter.go
```diff
@@ -0,0 +1,150 @@
...
+		snapshotItem := append([]byte(wasmFile.Name()), wasmBytes...)
+
+		snapshot.WriteExtensionItem(protoWriter, snapshotItem)
+	}
+
+	return nil
+}
+
+func (ws *WasmSnapshotter) Restore(
```
ethanfrey, line 106: Is it possible to do a unit test with some simple data, by feeding the output of snapshot into restore (or another instance with different directory) and ensuring the proper files are copied verbatim (and misnamed files ignored)?

x/wasm/wasm_snapshotter.go
```diff
@@ -0,0 +1,150 @@
...
+	height uint64, format uint32, protoReader protoio.Reader,
+) (snapshot.SnapshotItem, error) {
```
ethanfrey, line 108: This seems to read the entire data and return nothing ever. Why does it return `(snapshot.SnapshotItem, error)` rather than `error`? (eg. Is there something I/we are missing?)

ethanfrey (commented):
Thank you for your contribution.

It looks solid in general. Added a few comments, also an idea how to make a simple unit test just copying arbitrary files based on names between two dirs. That would be a nice sanity check before manual testing.

What do you think of a custom SnapshotItem payload format (not just appending bytes)?

app/app.go
webmaster128: The file structure is private and does not provide any stability guarantees. There is not even a guarantee Wasm is stored as file on disk in the next version of cosmwasm-vm. The file system should not be accessed from wasmd. Instead there is `GetCode` and `Create` in wasmvm to load and store a Wasm blob.

x/wasm/wasm_snapshotter.go
ethanfrey: I know this is not your design, but wonder why it is there

x/wasm/wasm_snapshotter.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+func (ws *WasmSnapshotter) SupportedFormats() []uint32 {
+	return []uint32{1}
+}
+
+var wasmFileNameRegex = regexp.MustCompile(`^[a-f0-9]{64}$`)
+
+func (ws *WasmSnapshotter) Snapshot(height uint64, protoWriter protoio.Writer) error {
+	wasmFiles, err := ioutil.ReadDir(ws.wasmDirectory)
```
webmaster128, line 76: `wasmFiles`, i.e. the list of checksums needs to come from the main Tendermint database. Otherwise a node can inject all kind of data here bypassing consensus.

x/wasm/wasm_snapshotter.go
```diff
@@ -0,0 +1,150 @@
...
+
+func (ws *WasmSnapshotter) SnapshotFormat() uint32 {
+	return 1
+}
+
+func (ws *WasmSnapshotter) SupportedFormats() []uint32 {
+	return []uint32{1}
+}
+
+var wasmFileNameRegex = regexp.MustCompile(`^[a-f0-9]{64}$`)
```
webmaster128, line 73: Those file names can change any time. I'm actually thinking about adding a .wasm file extension for node operator friendlyness.

app/app.go
assafmo: There's another "wasm" along the way
https://github.com/CosmWasm/wasmd/blob/3fb737147bff3358bd7f028c96cde8f6c191ace5/x/wasm/keeper/keeper.go#L99

webmaster128 (commented):
Thank you Assaf

x/wasm/wasm_snapshotter.go
assafmo: Yeah although I'm still not sure whether that's needed.

x/wasm/wasm_snapshotter.go
assafmo: Yeah, I took this from the sdk tests. Not sure if the return value goes anywhere here, but I can confirm that this works without any issue (tested manually on Secret). Also might be irrelevant for this case as we're dealing with files and not IAVL.

x/wasm/wasm_snapshotter.go
ethanfrey: This is a good point, but for a given wasmd version, there will be only one storage format (and realistically, changing this will be a major state-breaking change needing an internal wasmvm migration, so unlikely to happen, or at least very infrequently).

If it does change, we make a new wasmd (eg. 2.0) with snapshot format 2. Upon receiving that the data is parsed in a different way (eg the new internal format). The snapshotter itself (they one reading), will just need to implement the new format when we break this internal storage format.

assafmo:
@webmaster128 is right, it's better to go through the existing API instead of accessing the filesystem directly.

app/app.go
ethanfrey: > The file system should not be accessed from wasmd. Instead there is GetCode and Create in wasmvm to load and store a Wasm blob.

This is a good point, we can iterate over the values in the iavl tree rather than filenames. That is cleaner. And it will also "force" the node restoring from snapshotting to do the precompiling of the wasm files upon syncing, not later, providing more stable execution times later.

x/wasm/wasm_snapshotter.go
ethanfrey: That said, I do agree with using the GetCode and Create calls

assafmo: I agree that it's better to do it through the proper API, and let future migrations less to worry about WRT snapshotting.

assafmo: This also "solves" the "wasm/wasm/state/wasm` issue here.

webmaster128:
> @webmaster128 is right, it's better to go through the existing API instead of accessing the filesystem directly.

Thank you. Let me know if you need additional APIs or more advances features than the two mentioned above. But those are the two basic operations.

x/wasm/wasm_snapshotter.go
```diff
@@ -0,0 +1,150 @@
...
+		wasmFilePath := filepath.Join(ws.wasmDirectory, wasmFile.Name())
+		wasmBytes, err := ioutil.ReadFile(wasmFilePath)
+		if err != nil {
+			return err
+		}
+
+		// snapshotItem is 64 bytes of the file name, then the actual WASM bytes
+		snapshotItem := append([]byte(wasmFile.Name()), wasmBytes...)
+
+		snapshot.WriteExtensionItem(protoWriter, snapshotItem)
```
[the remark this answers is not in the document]
assafmo: I didn't consider it, not sure if it's a problem. The wasm files are usually far smaller than the rest of the blockchain's state.

It can be a nice addition though.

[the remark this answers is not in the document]
assafmo: Also not sure if it's taken care of by the sdk.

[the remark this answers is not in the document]
webmaster128: Okay, no priority or blocker for sure. I was just curious since Wasm blobs can be compressed nicely. But we can always make that an optional extension later.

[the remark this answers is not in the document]
assafmo: Interesting observation nonetheless. 💪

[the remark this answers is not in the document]
jhernandezb: wasmd already gzips wasm blobs when you submit a tx or through governance.
https://github.com/CosmWasm/wasmd/blob/d5ef3ba2de3c9c87adb2d8826da35f4c5b07bc3c/x/wasm/client/cli/tx.go#L88

https://github.com/CosmWasm/wasmd/blob/d5ef3ba2de3c9c87adb2d8826da35f4c5b07bc3c/x/wasm/client/utils/utils.go#L24


and then uncompressed when the `keeper.Create` is called 

https://github.com/CosmWasm/wasmd/blob/d5ef3ba2de3c9c87adb2d8826da35f4c5b07bc3c/x/wasm/keeper/ioutil.go#L18

x/wasm/wasm_snapshotter.go
jhernandezb: All this logic is in a loop where it seems like its the extension responsibility to handle the next `SnapshotItem` to the next one, which doesn't make sense because a plugin can break another one without knowing.

See https://github.com/cosmos/cosmos-sdk/blob/6f070623741fe0d6851d79ada41e6e2b1c67e236/snapshots/manager.go#L333

giansalex:
I was working on this feature, but tests are pending... I have updated with master branch, maybe it can be used as a reference.

Snapshot:
- Iterate  codes
- Remove duplicate hash
- Save CodeHash, WasmBytes using [SnapshotWasmItem](https://github.com/disperze/wasmd/blob/state-sync-sdk-45.2/proto/cosmwasm/wasm/v1/types.proto#L143)

Restore:
- Get codehash, wasmbytes from SnapshotPayload
- Create wasm using vm
- Verify hash

https://github.com/disperze/wasmd/blob/state-sync-sdk-45.2/x/wasm/keeper/snapshot.go
[ [full diff](https://github.com/CosmWasm/wasmd/compare/main...disperze:state-sync-sdk-45.2?expand=1) ]

webmaster128:
@giansalex Thank you! Could you add your work as a second PR to provide better visibility and being able to comment inline?

alpe (commented):
Thanks for the PR and all the good comments here. Nice collaboration!
I think @assafmo solution was a very pragmatic approach and automates what people already do with sharing zipped wasm folders. 
I completely agree with @webmaster128 suggestions for decoupling and a more resilient solution that uses existing apis. Upstream implementations may change and can break this easily otherwise. 
Proper tests can be built with our [mock](https://github.com/CosmWasm/wasmd/blob/main/x/wasm/keeper/wasmtesting/mock_engine.go#L19)

ethanfrey:
I got a anon review via DM:

When restoring all the wasm files, we must ensure to tell the wasmvm to pin the proper codes.

If we use the Create method, this will just be one more check on CodeInfo and possibly another call to wasmvm

alpe:
[InitializePinnedCodes](https://github.com/CosmWasm/wasmd/blob/main/x/wasm/keeper/keeper.go#L811) can be used to pin all codes again.

assafmo:
> I got a anon review via DM:
> 
> When restoring all the wasm files, we must ensure to tell the wasmvm to pin the proper codes.
> 
> If we use the Create method, this will just be one more check on CodeInfo and possibly another call to wasmvm

I'm not sure it that's necessary. Codes info are already stored in the IAVL.

webmaster128:
> I'm not sure it that's necessary. Codes info are already stored in the IAVL.

[InitializePinnedCodes](https://github.com/CosmWasm/wasmd/blob/main/x/wasm/keeper/keeper.go#L811) is basically the answer to that DM message. It's really about lifting contracts into cache after a node restart or state sync.

tomtau:
copy-pasting from Discord:
> fyi we have this simple test on CI for state sync: https://github.com/crypto-org-chain/chain-main/blob/master/integration_tests/test_basic.py#L96 via this tool: https://github.com/crypto-com/pystarport ... so I imagine a test on CI for wasmd may be an extension of it: create a network, deploy a contract there, create a new node with state sync to join the network, submit a tx to invoke the contract-related functionality, assert that query responses from the state sync'ed and the original nodes are the same

x/wasm/wasm_snapshotter.go
ethanfrey: > My guess is that is possible that in the protobuf stream there might be more items and if the format is not the one that wasmd can understand then it should return it.

Yes, very good idea. It seems like we should not error on some unknown extension, but just return it, as there may in the future be multiple extensions than need to cooperate somehow. Not sure how this all works together.

Maybe @yihuang could clarify how we should handle this properly

ethanfrey:
@assafmo Very helpful PR showing how to implement such an extension.

There have been a number of wonderful comments explaining how to improve the process, and add more features (pinning, compression, etc). 

I would like to get a version of this in incorporating all the input. Do you want to update this PR with all feedback? Or shall I write one? I'm happy either day. I would try to write it or review it tomorrow (Thursday) to get this in soon, and let people test  manually on some tag.

assafmo:
I'm not sure if I can get to it that soon, so go for it, no ego here. ❤️
BTW I'm not sure that pinning is necessary here.

x/wasm/wasm_snapshotter.go
yihuang: > Yes, very good idea. It seems like we should not error on some unknown extension, but just return it, as there may in the future be multiple extensions than need to cooperate somehow. Not sure how this all works together.

exactly, we should return the item if it's unknown to us.

yihuang: > which doesn't make sense because a plugin can break another one without knowing.

yeah, that's possible unfortunately, maybe we can create some helpers at module side without changing the cosmos-sdk api.

ethanfrey: ah, I got it... we *either* end on EOF, or when we hit something where `GetExtensionPayload()` returns nil (another type we don't expect)

ethanfrey (commented):
Building on this, I have made a new PR that attempts to incorporate all the discussion.

Please check out #823 

It should be complete now, "just" missing tests, but good for a first review.

x/wasm/wasm_snapshotter.go
yihuang: > > which doesn't make sense because a plugin can break another one without knowing.
> 
> yeah, that's possible unfortunately, maybe we can create some helpers at module side to retain the last item automatically, without changing the api in cosmos-sdk.

```

// ExtensionPayloadReader only reads the payloads of current extension.
type ExtensionPayloadReader struct {
	protoReader protoio.Reader
	LastItem    types.SnapshotItem
}

func NewExtensionPayloadReader(protoReader protoio.Reader) *ExtensionPayloadReader {
	return &ExtensionPayloadReader{
		protoReader: protoReader,
		LastItem:    types.SnapshotItem{},
	}
}

// ReadPayload reads a payload item from the stream.
func (epr *ExtensionPayloadReader) ReadPayload() ([]byte, error) {
	epr.LastItem = types.SnapshotItem{}
	if err := epr.protoReader.ReadMsg(&epr.LastItem); err != nil {
		return nil, err
	}
	payload := epr.LastItem.GetExtensionPayload()
	if payload == nil {
		// end of payload stream of current extension
		return nil, io.EOF
	}
	return payload.Payload, nil
}
```

I think of sth like this could help, the manager pass the `ExtensionPayloadReader` extension's `Restore` method, but it's api breaking, so maybe change this in next version.

ethanfrey:
> I was working on this feature, but tests are pending... I have updated with master branch, maybe it can be used as a reference.

@giansalex sorry I didn't see your code earlier. I was looking at open PRs... even draft.

I will look now, definitely would have saved me time yesterday and some things I can learn from it still

assafmo:
Closing in favor of #823.

app/app.go
sascha1337: sounds legit!

PR #817, opened 2022-04-25
More Granular Authz perms for cosmwasm contracts

jackzampolin:


x/wasm/types/types.go
```diff
@@ -22,6 +25,109 @@ const (
...
+// NewExecuteContractAuthorization creates a new ExecuteContractAuthorization object.
+func NewExecuteContractAuthorization(allowedContracts []*AllowedContract) *ExecuteContractAuthorization {
+	return &ExecuteContractAuthorization{
+		AllowedContracts: allowedContracts,
+	}
+}
+
+// MsgTypeURL implements Authorization.MsgTypeURL.
+func (a ExecuteContractAuthorization) MsgTypeURL() string {
+	return "/cosmos.wasm.base.MsgExecuteContract"
```
giansalex, line 48:
```suggestion
	return sdk.MsgTypeURL(&MsgExecuteContract{})
```
should be `/cosmwasm.wasm.v1.MsgExecuteContract` , ref [sendAuthz](https://github.com/cosmos/cosmos-sdk/blob/v0.45.3/x/bank/types/send_authorization.go#L22)

x/wasm/types/types.go
```diff
@@ -22,6 +25,109 @@ const (
...
+				continue
+			} else if len(allowedContract.AllowedMessages) > 0 {
+				// otherwise we exclude it from the update array
+				continue
+			}
+
+			// NOTE: exec.Msg is base64 encoded, so we need to decode it first
+			// If there are allowedMessages then decode the call and check if the message is allowed
+			callBz := []byte{}
+			if _, err := base64.RawStdEncoding.Decode(exec.Msg.Bytes(), callBz); err != nil {
```
giansalex, line 76: exec.Msg is json data

x/wasm/types/types.go
```diff
@@ -22,6 +25,109 @@ const (
...
+			}
+
+			messageName := map[string]interface{}{}
+			if err := json.Unmarshal(callBz, &messageName); err != nil {
+				return authztypes.AcceptResponse{}, sdkerrors.Wrap(err, "failed to unmarshal json")
+			}
+
+			for _, allowedFunction := range allowedContract.AllowedMessages {
+				if len(messageName) != 1 {
+					return authztypes.AcceptResponse{}, sdkerrors.ErrInvalidRequest.Wrap("too many message calls in the same transaction")
```
webmaster128, line 87: This can be checked before the loop

x/wasm/types/types.go
```diff
@@ -22,6 +25,109 @@ const (
...
+
+			// NOTE: exec.Msg is base64 encoded, so we need to decode it first
+			// If there are allowedMessages then decode the call and check if the message is allowed
+			callBz := []byte{}
+			if _, err := base64.RawStdEncoding.Decode(exec.Msg.Bytes(), callBz); err != nil {
+				return authztypes.AcceptResponse{}, sdkerrors.Wrap(err, "failed to decode base64")
+			}
+
+			messageName := map[string]interface{}{}
+			if err := json.Unmarshal(callBz, &messageName); err != nil {
```
webmaster128, line 81: Looks good for a start

webmaster128 (commented):
Could you extract the whole JSON matching logic into a separate function that can easily be unit tested? Like `func isJSONObjectWithTopLevelKey(jsonBytes, allowedKeys)`.

webmaster128:
I extracted and implemented the JSON object matching in https://github.com/CosmWasm/wasmd/pull/819 (last commit). I wanted to create a PR into your PR but chan't easily choose your branch as the base branch. Can you check the last commit and cherry-pick it?

giansalex:
A version using `IsJSONObjectWithTopLevelKey` function.

```go
func (a ExecuteContractAuthorization) Accept(ctx sdk.Context, msg sdk.Msg) (authztypes.AcceptResponse, error) {
	exec, ok := msg.(*MsgExecuteContract)
	if !ok {
		return authztypes.AcceptResponse{}, sdkerrors.ErrInvalidType.Wrap("type mismatch")
	}

	var allowed *AllowedContract
	for _, allowedContract := range a.AllowedContracts {
		// ctx.GasMeter().ConsumeGas(gasCostPerIteration, "wasm execute authorization")
		if allowedContract.ContractAddress == exec.Contract {
			allowed = allowedContract
			break
		}
	}

        // verify contract is allowed
	if allowed == nil {
		return authztypes.AcceptResponse{}, sdkerrors.ErrUnauthorized.Wrapf("cannot run %s contract", exec.Contract)
	}

        // verify msg is allowed
	err := IsJSONObjectWithTopLevelKey(exec.Msg, allowed.AllowedMessages)
	if err != nil {
		return authztypes.AcceptResponse{}, sdkerrors.ErrUnauthorized.Wrapf("no allowed msg: %s", err.Error())
	}

	if allowed.Once {
		return authztypes.AcceptResponse{Accept: true, Delete: true}, nil
	}

	return authztypes.AcceptResponse{Accept: true}, nil
}
```

giansalex:
[staking authz](https://github.com/cosmos/cosmos-sdk/blob/v0.45.3/x/staking/types/authz.go#L82) consumes gas for iterations, in this case contract list or msgs list.

giansalex:
yes, json parsing.  I think it might also be necessary to add a command `tx wasm grant  ....`

ethanfrey:
This is a great idea.
Work has continued in #966 and #978 and we will use an implementation based on those in the next v0.30 release.

Thank you for pushing it forward with this spike.

Issue #813, opened 2022-04-25
Replace many const with var


ethanfrey:
This would address many of the points raised in #202  and also be a cleaner solution than fully removing checks as in #809 

Using a var for these values will allow a chain importing wasmd (or compiling a custom version) to easily override the values compile time. It will not allow a param change proposal to fix it, but any chain upgrade could update these values as well, so it does make them somewhat dynamic (or at least adjustable by governance without forking wasmd).

This should be a minimally intrusive change that will remove a number of headaches for integrations.

alpe:
With the [gas register](https://github.com/CosmWasm/wasmd/blob/main/x/wasm/keeper/gas_register.go#L58) interface and extension point, custom values or logic can be integrated easily. Switching to vars for them with compiler flags may cause more confusion as we have 2 ways now to set values.

`MaxWasmSize` and `MaxLabelSize` are a different story though. They are not customizable for chains, currently. Either we make them both vars or use a custom ante handler that can read from params.
The ante handler would be a cleaner solution for me as it removes duplication but adds on gas cost or requires caching. 
Wondering if there is a use case for custom `MaxLabelSize` 🤷

ethanfrey:
Great. If it is just those two left to do, let's make them vars and document that can call this done.

I don't see the need to customize `MaxLabelSize`, but someone will surely come up with one. If we just make it var and keep the same default, it should be fine.

PR #822, opened 2022-04-27
Make MaxLabelSize a var not const

ethanfrey:
Addresses #813 along with #809

MaxLabelSize is a now var, which can be overridden compile-time by any chain that imports `x/wasm`. 
The 128 char limit, while reasonable, was an arbitrary choice by Confio developers and a blockchain may wish to customize it.

To override, set this variable to a new value in some initialisation code (either `app.go` or `cmd/../main.go`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 [Full Changelog](https://github.com/CosmWasm/wasmd/compare/v0.26.0...HEAD)
 
+**Implemented Enhancements**
+
+- Make MaxLabelSize a var not const [\#822](https://github.com/CosmWasm/wasmd/pull/822)
+
 ## [v0.26.0](https://github.com/CosmWasm/wasmd/tree/v0.26.0) (2022-04-21)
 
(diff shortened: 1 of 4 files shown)


pinosu (approved):
lgtm 👍

maurolacy (approved):
LGTM.

alpe (approved):
LGTM 👍 
Bonus points, if you could add some doc to
https://github.com/CosmWasm/wasmd#runtime-flags
for this new flag

ethanfrey:
I did that and then tested like:

`go test -count=1 -ldflags '-X github.com/CosmWasm/wasmd/x/wasm/types.MaxLabelSize=64' ./keeper`

 But that fails with:
 
 `github.com/CosmWasm/wasmd/x/wasm/types.MaxLabelSize: cannot set with -X: not a var of type string (type.int)`

I would try to make this a string and parse it and all, but I think it is easier to just say to override in app.go. People modifying this should have their own repo importing it anyway

PR #823, opened 2022-04-27
Add state sync support

ethanfrey:
This replaces #816 incorporating much of the feedback.

Main changes:

* Iterates over codes in the iavl tree to get proper list
* compresses wasm before sending over snapshot item
* skips duplicates (multiple codes with same byte code)
* uses wasmvm.Create() on restore, which will precompile the code
* pins all proper codes in the cache, just like on normal startup
* Play [better with other extensions](https://github.com/CosmWasm/wasmd/pull/816#discussion_r859703173) - done in https://github.com/CosmWasm/wasmd/pull/823/commits/7a29351463072a8daeb855b726fb933b355454cc

TODO:

* unit tests

As a side, I pull the compress logic (in cli/utils) and the uncompress logic (from keeper) into a common package. That could be it's own PR, if someone really cares, I guess I could cherry pick https://github.com/CosmWasm/wasmd/pull/823/commits/71ff81edae9cf9d1435b9e2086316908a45bf46b out


x/wasm/keeper/snapshotter.go
```diff
@@ -0,0 +1,172 @@
...
+			return snapshot.SnapshotItem{}, finalize(ctx, ws.wasm)
+		} else if err != nil {
+			return snapshot.SnapshotItem{}, sdkerrors.Wrap(err, "invalid protobuf message")
+		}
+
+		// if it is not another ExtensionPayload message, then it is not for us.
+		// we should return it an let the manager handle this one
+		payload := item.GetExtensionPayload()
+		if payload == nil {
+			return item, nil
```
ethanfrey, line 165: @yihuang If I understand https://github.com/CosmWasm/wasmd/pull/816#discussion_r859703173 correctly, then this is the correct behavior. Can you confirm or correct me?

x/wasm/keeper/snapshotter.go
```diff
@@ -0,0 +1,172 @@
...
+	return 1
+}
+
+func (ws *WasmSnapshotter) SupportedFormats() []uint32 {
+	return []uint32{1}
+}
+
+func (ws *WasmSnapshotter) Snapshot(height uint64, protoWriter protoio.Writer) error {
+	var rerr error
+	ctx := sdk.NewContext(ws.cms, tmproto.Header{}, false, log.NewNopLogger())
```
assafmo, line 78: Cool!

x/wasm/keeper/snapshotter.go
```diff
@@ -0,0 +1,172 @@
...
+	cb func(sdk.Context, Keeper, []byte) error,
+	finalize func(sdk.Context, Keeper) error,
+) (snapshot.SnapshotItem, error) {
+	ctx := sdk.NewContext(ws.cms, tmproto.Header{}, false, log.NewNopLogger())
+
+	for {
+		item := snapshot.SnapshotItem{}
+		err := protoReader.ReadMsg(&item)
+		if err == io.EOF {
+			return snapshot.SnapshotItem{}, finalize(ctx, ws.wasm)
```
giansalex, line 156: there may be other modules after wasm snapshotter, and this condition may not occur.

giansalex:
this is the place where I think the pin codes should be initialized
image

x/wasm/keeper/snapshotter.go
```diff
@@ -0,0 +1,172 @@
...
+
+func NewWasmSnapshotter(cms sdk.CommitMultiStore, wasm Keeper) *WasmSnapshotter {
+	return &WasmSnapshotter{
+		wasm: wasm,
+		cms:  cms,
+	}
+}
+
+func (ws *WasmSnapshotter) SnapshotName() string {
+	return "WASM Code Snapshot"
```
giansalex, line 65: I think it could be `types.ModuleName`

x/wasm/keeper/snapshotter.go
```diff
@@ -0,0 +1,172 @@
...
+	return rerr
+}
+
+func (ws *WasmSnapshotter) Restore(
+	height uint64, format uint32, protoReader protoio.Reader,
+) (snapshot.SnapshotItem, error) {
+	if format == 1 {
+		return ws.processAllItems(height, protoReader, restoreV1, finalizeV1)
+	}
+	return snapshot.SnapshotItem{}, snapshot.ErrUnknownFormat
```
giansalex, line 122: this validation is performed by SnapshotManager using Snaphotter.SupportFormats , but I guess it's not a problem

x/wasm/keeper/snapshotter.go
assafmo: So then how do you exit the loop?

giansalex: when SnaphostItem is not a payload

assafmo: Oh, okay, so this condition is still needed and it was just a general comment.

giansalex: yes, is necessary, but `finalize(ctx, ws.wasm)` will not run in some scenarios

x/wasm/keeper/snapshotter.go
ethanfrey: Got you, I need to run finalise here as well.

x/wasm/keeper/snapshotter.go
ethanfrey: I think I addressed this for the other case we end without error.
Please confirm

ethanfrey:
Thanks for the reviews, I made some updated

x/wasm/keeper/snapshotter.go
giansalex: @ethanfrey yes, ok

yihuang: Nit: multitstore handle the finalization at the end of loop, and use break here and when `payload == nil`.

ethanfrey:
I used this line from @giansalex to properly load from a historical height: https://github.com/disperze/wasmd/blob/state-sync-sdk-45.2/x/wasm/keeper/snapshot.go#L49

However, all the tests panic when I try to commit the state before snapshotting (so I have a height to read from). It keeps saying the version was already committed. I spent 30 minutes digging in the sdk, but remain lost. Maybe someone has insights.

See https://github.com/CosmWasm/wasmd/pull/823/commits/7663d6ffdd700dbf48b8474ac7b956f1429a7bb5 and error messages

ethanfrey:
I think this is about as good as I can do.

Happy for reviews. And if anyone wants to make a PR on top to address the historical height issue (or add app level tests), I would be happy to merge it in. Otherwise, I would revisit this next week Tues/Wed and try to get this in main.

Also, very happy for any manual testing on this. I definitely need that before tagging the release, but would merge before that if possible.

x/wasm/ioutils/utils.go
```diff
@@ -1,12 +1,16 @@
...
 import (
 	"bytes"
 	"compress/gzip"
 )
 
 var (
+	// magic bytes to identify gzip.
+	// See https://www.ietf.org/rfc/rfc1952.txt
+	// and https://github.com/golang/go/blob/master/src/net/http/sniff.go#L186
 	gzipIdent = []byte("\x1F\x8B\x08")
```
webmaster128, line 12: Nice dedup. But it should be const, right?

x/wasm/keeper/snapshotter.go
```diff
@@ -0,0 +1,190 @@
...
+	// user requests
+	// ------
+	// cacheMS, err := ws.cms.CacheMultiStoreWithVersion(int64(height))
+	// if err != nil {
+	// 	return err
+	// }
+	cacheMS := ws.cms.CacheMultiStore()
+
+	ctx := sdk.NewContext(cacheMS, tmproto.Header{}, false, log.NewNopLogger())
+	uniqueHashes := make(map[string]bool)
```
webmaster128, line 92: I think `seenBefore` would be a better name as this is the content map values: for each hash, has it been seen before.

x/wasm/keeper/snapshotter.go
```diff
@@ -0,0 +1,190 @@
...
+	return snapshot.SnapshotItem{}, snapshot.ErrUnknownFormat
+}
+
+func restoreV1(ctx sdk.Context, k *Keeper, compressedCode []byte) error {
+	wasmCode, err := ioutils.Uncompress(compressedCode, k.GetMaxWasmCodeSize(ctx))
+	if err != nil {
+		return sdkerrors.Wrap(types.ErrCreateFailed, err.Error())
+	}
+
+	// FIXME: check which codeIDs the checksum matches??
```
webmaster128, line 144: Which code ID is not important. But ensuring the checksum exists in the codes list at all makes sense. Otherwise a malicious node could send all kind of unrelated Wasm that gets stored and compiled, right?

webmaster128 (commented):
nice

x/wasm/keeper/snapshotter.go
ethanfrey: I mean, they can send us anything.
And at the point we have the checksum to compare it has been stored and compiled.

I guess this check would only allow one useless wasm.

I am more interested in asserting that all needed wasm has been synced. The biggest problem is a sync where some wasm file is missing and the node crashes later when that contract is called.

x/wasm/keeper/snapshotter.go
```diff
@@ -0,0 +1,190 @@
...
+	// If we support older formats, add them here and handle them in Restore
+	return []uint32{SnapshotFormat}
+}
+
+func (ws *WasmSnapshotter) Snapshot(height uint64, protoWriter protoio.Writer) error {
+	// TODO: This seems more correct (historical info), but kills my tests
+	// Since codeIDs and wasm are immutible, it is never wrong to return new wasm data than the
+	// user requests
+	// ------
+	// cacheMS, err := ws.cms.CacheMultiStoreWithVersion(int64(height))
```
yihuang, line 85: If we use the snapshot storage at the height, would different nodes could generate different snapshots?

assafmo: Yes that might make result in 2 nodes creating a different snapshot for the same height, which will make a client reject the snapshot.

alpe (commented):
I need to make myself more familiar with the internals of the snapshot package. Beside some minor comments, your code looks very good.

x/wasm/keeper/snapshotter.go
```diff
@@ -0,0 +1,190 @@
...
+		seenBefore[hexHash] = true
+
+		// load code and abort on error
+		wasmBytes, err := ws.wasm.GetByteCode(ctx, id)
+		if err != nil {
+			rerr = err
+			return true
+		}
+
+		compressedWasm, err := ioutils.GzipIt(wasmBytes)
```
alpe, line 111: Just found: 

> The current version 1 snapshot format is a zlib-compressed, length-prefixed Protobuf stream of cosmos.base.store.v1beta1.SnapshotItem messages, split into chunks at exact 10 MB byte boundaries.

https://github.com/cosmos/cosmos-sdk/tree/main/snapshots#snapshot-format

x/wasm/keeper/snapshotter.go
```diff
@@ -0,0 +1,190 @@
...
+	protoio "github.com/gogo/protobuf/io"
+	"github.com/tendermint/tendermint/libs/log"
+	tmproto "github.com/tendermint/tendermint/proto/tendermint/types"
+
+	"github.com/CosmWasm/wasmd/x/wasm/ioutils"
+	"github.com/CosmWasm/wasmd/x/wasm/types"
+)
+
+/*
+API to implement:
```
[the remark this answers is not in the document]
alpe: `ExtensionSnapshotter` inherits  from `Snapshotter` so that would be enough as a ref

assafmo:
Glad to see this merged! Did you find what caused `ws.cms.CacheMultiStoreWithVersion(int64(height))` not to work at first?

ethanfrey:
It worked. Just not with our test code.

Alex used a different test setup (the full app setup) and it works now

PR #826, opened 2022-04-29
Consolidate MaxWasmSize constraints into a single var

### LINKED ISSUE #813: Replace many const with var

alpe:
Resolves #813 

This PR contains all work started in #809 by @bigs rebased on top of current main. 
Big thanks for the contribution and driving this forward! 💐 

Note: with this PR the max code size was changed from 500kb to [800kb](https://github.com/CosmWasm/wasmd/pull/826/files#diff-97926945f9085be2d68111c06136885b7031717ee4c629b0ed3bbb9a0cd4d4e8R12)


pinosu (approved):
lgtm 👍

bigs:
@jhernandezb Good questions! As far as breaking changes go, I think the only time you'd experience breaking behavior is if your param was set to a value lower than the 500KB that was defined as a constant in the validation section.

The old param not being present shouldn't break anything, but I do agree it would be tidier to write a migration to remove it from the store, as it certainly won't be used anymore.

jhernandezb:
I'll add an upgrade test to our chain code targeting this branch and see how it goes.

PR #827, opened 2022-04-29
Add instantiate-nobody flag to cli store cmd

jhernandezb:
adds `instantiate-nobody` flag to  regular tx, store-code gov proposal, genesis store commands.

### DIFF
--- a/x/wasm/client/cli/genesis_msg.go
+++ b/x/wasm/client/cli/genesis_msg.go
@@ -71,6 +71,7 @@ func GenesisStoreCodeCmd(defaultNodeHome string, genesisMutator GenesisMutator) 
 	}
 	cmd.Flags().String(flagRunAs, "", "The address that is stored as code creator")
 	cmd.Flags().String(flagInstantiateByEverybody, "", "Everybody can instantiate a contract from the code, optional")
+	cmd.Flags().String(flagInstantiateNobody, "", "Nobody can instantiate a contract from the code, optional")
 	cmd.Flags().String(flagInstantiateByAddress, "", "Only this address can instantiate a contract instance from the code, optional")
 
 	cmd.Flags().String(flags.FlagHome, defaultNodeHome, "The application home directory")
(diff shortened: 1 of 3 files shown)


x/wasm/client/cli/genesis_msg.go
```diff
@@ -71,6 +71,7 @@ func GenesisStoreCodeCmd(defaultNodeHome string, genesisMutator GenesisMutator)
 	}
 	cmd.Flags().String(flagRunAs, "", "The address that is stored as code creator")
 	cmd.Flags().String(flagInstantiateByEverybody, "", "Everybody can instantiate a contract from the code, optional")
+	cmd.Flags().String(flagInstantiateNobody, "", "Nobody can instantiate a contract from the code, optional")
```
JakeHartnell, line 74: Would almost want to call it `flagInstantiateGovernace`, `"Only governance can instantiate a contract from the code, optional"`. That might be more clear?

Otherwise this is awesome!

jhernandezb: this follows the naming convention of  `AcessConfig` while it's true only governance will be able to instantiate it with a default wasmd setup, might not be the case for other chains using the keeper directly with different `AuthorizationPolicy`.  Happy to change it tho to reflect the most common case.

jhernandezb: Maybe, `instantiate-governance-only` ?

shanev: The convention so far has been to use `nobody` no?

alpe: I like being verbose in the description. Without more context "nobody" sounds odd. 
How about "Nobody except the governance process can instantiate a contract from the code, optional" 🤷

alpe (approved):
Thanks for the PR 💐 
LGTM with the description as it is or mentioning the gov process!

x/wasm/client/cli/genesis_msg.go
jhernandezb: updated description

PR #828, opened 2022-04-30
fix: allow no admin in instantiation through proposal

jhernandezb:


x/wasm/keeper/proposal_handler.go
```diff
@@ -77,9 +77,11 @@ func handleInstantiateProposal(ctx sdk.Context, k types.ContractOpsKeeper, p typ
 	if err != nil {
 		return sdkerrors.Wrap(err, "run as address")
 	}
-	adminAddr, err := sdk.AccAddressFromBech32(p.Admin)
-	if err != nil {
-		return sdkerrors.Wrap(err, "admin")
+	var adminAddr sdk.AccAddress
+	if p.Admin != "" {
```
alpe, line 81: 👍 this was a bug

x/wasm/keeper/proposal_integration_test.go
```diff
@@ -123,6 +123,80 @@ func TestInstantiateProposal(t *testing.T) {
 	require.NotEmpty(t, em.Events()[2].Attributes[0])
 }
 
+func TestInstantiateProposal_NoAdmin(t *testing.T) {
```
alpe, line 126: Thanks for covering the additional cases! If you feel table tests are a better fit, then please feel free to refactor tests as well.

alpe (approved):
Thanks for finding and fixing this 🐛 ! 
LGTM 💯

PR #832, opened 2022-05-03
formatting: gofumpt 

faddat:
This PR should change nothing functionally, but instead nicely organizes the code and makes it easier to read.

It is easy to verify:

```bash
go install mvdan.cc/gofumpt@latest
gofumpt -l -w .
```

### DIFF
--- a/app/ante.go
+++ b/app/ante.go
@@ -38,7 +38,7 @@ func NewAnteHandler(options HandlerOptions) (sdk.AnteHandler, error) {
 		return nil, sdkerrors.Wrap(sdkerrors.ErrLogic, "tx counter key is required for ante builder")
 	}
 
-	var sigGasConsumer = options.SigGasConsumer
+	sigGasConsumer := options.SigGasConsumer
 	if sigGasConsumer == nil {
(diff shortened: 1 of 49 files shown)


alpe:
Nice one! I like the changes and appreciate your contribution. 🏅 
This touches a lot of files while there are some important feature PRs in progress that cover IBC and statesync. Let's run the  command again when they got merged or after the v0.27.0 final tag at the latest. I will open an issue to keep track of this.

ethanfrey:
Nice addition.

I would like to merge the large outstanding PRs first, before this, to avoid conflicts.
Then happy if you regenerate the cleanup PR and we can merge quickly

ethanfrey:
It would be good to make a Makefile entry (as a separate commit) that does the formatting. So devs can easily run this.

That commit could then easily be cherry-picked once we close most of the other open PRs (to avoid conflicts) and then run it

ethanfrey:
Closing for #851 which adds the commands from the comment to the Makefile.

We will likely need to run that again, but easier to reproduce (and tagged to a fixed version, not latest)

PR #836, opened 2022-05-04
Add instantiate permission to CodeInfoResponse

### LINKED ISSUE #829: Add AccessConfig to CodeInfo query response
It would be nice to have AccessConfig as part of the CodeInfo query in 
https://github.com/CosmWasm/wasmd/blob/main/proto/cosmwasm/wasm/v1/query.proto#L173

Currently can't find a way to get the current configuration of a code id.

jhernandezb:
closes #829


alpe (commented):
👍 Very nice! thanks for the contribution.

I was surprised to see no tests were touched nor failing. It looks like none is actually covering the full response object. As this is part of the public API we should ensure the expected data is returned. 
The `legacy querier` is deprecated already but adding 2 tests for the grpc querier would be a great contribution and fit into the scope of this PR.
Please let me know if you have capacity otherwise I can merge the PR as it is.

jhernandezb:
> 

yes I can't add the tests actually I already had one in draft will complete them and push the update.

x/wasm/keeper/querier_test.go
```diff
@@ -711,6 +711,72 @@ func TestQueryCodeInfo(t *testing.T) {
...
+	allCodesResponse := make([]types.CodeInfoResponse, 0)
+	for _, code := range codes {
+		t.Run(fmt.Sprintf("import_%s", code.name), func(t *testing.T) {
+			require.NoError(t, keeper.importCode(ctx, code.codeId,
+				code.codeInfo,
+				wasmCode),
+			)
+		})
+
+		allCodesResponse = append(allCodesResponse, types.CodeInfoResponse{
```
alpe, line 761: When I am strict, then there is a little smell as the test cases are not independent.
When you import with a "cached store" , the modified state would not be persisted
```go
ctx, _ := parentCtx.CacheContext()
```

x/wasm/keeper/querier_test.go
```diff
@@ -711,6 +711,72 @@ func TestQueryCodeInfo(t *testing.T) {
...
+
+	anyAddress, err := sdk.AccAddressFromBech32("cosmos100dejzacpanrldpjjwksjm62shqhyss44jf5xz")
+	require.NoError(t, err)
+	codeInfoWithConfig := func(accessConfig types.AccessConfig) types.CodeInfo {
+		codeInfo := types.CodeInfoFixture(types.WithSHA256CodeHash(wasmCode))
+		codeInfo.InstantiateConfig = accessConfig
+		return codeInfo
+	}
+
+	codes := []struct {
```
alpe, line 730: We mostly use map based table tests. I have this macro for Goland:
```
	specs := map[string]struct{
        $START$
	}{
		"$SECOND$": {
		$THIRD$
		},
		
	}
	for name, spec := range specs {
		t.Run(name, func(t *testing.T) {
            $FORTH$
		})
	}
```

alpe (approved):
Well done and thanks for the additional tests! 💪 
I commented some nits as I hope you can add more with your contributions in the future.

PR #838, opened 2022-05-04
Fix contract path in simulations

alpe:
* Move wasm file loading into function to avoid global var
* Fix contract path
* Configurable contract path

Can be tested with:
```sh
make test-sim-multi-seed-short
```


ethanfrey (approved):
Ah, didn't realise they run in different folders.
Good one.

pinosu (approved):
Very nice! I tested and it works fine 💯

PR #839, opened 2022-05-04
Add mauth for real authorization

ethanfrey:
Copied from Gaia branch.
I had to update from v3.0.0-alpha2. I think I made all needed API changes, but something fails now.


ethanfrey:
Probably better just to import a recent version once https://github.com/cosmos/interchain-accounts-demo/issues/91 gets resolved

PR #845, opened 2022-05-06
Add Debian docker build to test libwasmvm.aarch64.so

webmaster128:
wasmvm 1.0.0-rc.0 and wasmd v0.27.0-rc0 now allow running on glibc Linux on an ARM64 machine. With this Dockerfile this can be tested on an M1 dev machine.

Creating this draft for demonstration purposes. Not sure if and to what degree we want to maintain the docker file or turn it into documentation. Let's see.


alpe:
There was not any feedback or support on this spike. Let's close this now. 

Personally, I prefer statically built binaries as they are work nicely with tools like cosmovisor. On dynamically linked binaries, there is extra work to keep the libs in place on an upgrade.

PR #846, opened 2022-05-06
Add libwasm-version query command

jhernandezb:
New wasmvm introduces `LibwasmvmVersion`  api method that can help to get the version of the library that is being resolved specially for dynamic linked binaries. 

I'm adding it as a query because it will be automatically wired to apps using wasmd but can be just a helper command and requires manual wiring.

### DIFF
--- a/x/wasm/client/cli/query.go
+++ b/x/wasm/client/cli/query.go
@@ -36,18 +36,19 @@ func GetQueryCmd() *cobra.Command {
 		GetCmdGetContractHistory(),
 		GetCmdGetContractState(),
 		GetCmdListPinnedCode(),
-		GetCmdVMVersion(),
+		GetCmdLibVersion(),
 	)
 	return queryCmd
 }
 
-// GetCmdVMVersion gets current libwasmvm version.
-func GetCmdVMVersion() *cobra.Command {
+// GetCmdLibVersion gets current libwasmvm version.
+func GetCmdLibVersion() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "vm-version",
-		Short: "Get libwasmvm version",
-		Long:  "Get libwasmvm version",
-		Args:  cobra.ExactArgs(0),
+		Use:     "libwasmvm-version",
+		Short:   "Get libwasmvm version",
+		Long:    "Get libwasmvm version",
+		Aliases: []string{"lib-version"},
+		Args:    cobra.ExactArgs(0),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			version, err := wasmvmapi.LibwasmvmVersion()
 			if err != nil {
--- a/x/wasm/client/cli/query.go
+++ b/x/wasm/client/cli/query.go
@@ -17,6 +17,7 @@ import (
 	flag "github.com/spf13/pflag"
 
 	"github.com/CosmWasm/wasmd/x/wasm/types"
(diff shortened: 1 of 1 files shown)


x/wasm/client/cli/query.go
```diff
@@ -35,10 +36,30 @@ func GetQueryCmd() *cobra.Command {
...
 		GetCmdListPinnedCode(),
+		GetCmdVMVersion(),
 	)
 	return queryCmd
 }
 
+// GetCmdVMVersion gets current libwasmvm version.
+func GetCmdVMVersion() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "vm-version",
```
webmaster128, line 47: The term "VM" is a bit overused and `wasmvm` is the Go project that contains code for both sides of the FFI interface. To avoid confusion, I'd go with "libwasmvm-version". A better abbreviation would be "lib-version". If we need something short, let's use that as alias.

x/wasm/client/cli/query.go
```diff
@@ -17,6 +17,7 @@ import (
 	flag "github.com/spf13/pflag"
 
 	"github.com/CosmWasm/wasmd/x/wasm/types"
+	wasmvmapi "github.com/CosmWasm/wasmvm/api"
```
webmaster128, line 20: Question to the Go pros: is the export via the `/api` folder good that way? Or should I make this available in the root project `github.com/CosmWasm/wasmvm`?

webmaster128 (commented):
Nice

webmaster128:
For some more context: this specific command is temporary and can be removed once the Cosmos SDK version command is extensible (see https://github.com/cosmos/cosmos-sdk/issues/11690).

x/wasm/client/cli/query.go
jhernandezb: sounds good to me

x/wasm/client/cli/query.go
jhernandezb: seems like all the imports in wasmd are from the root project. 
maybe just set an alias at root level `var LibwasmvmVersion = api.LibwasmvmVersion`

x/wasm/client/cli/query.go
jhernandezb: renamed and added the alias as well.

webmaster128 (approved):
Great, thanks.

Let's wait for Alex's review next week.

x/wasm/client/cli/query.go
alpe: IMHO root level would be the first place to look for this kind of functionality

x/wasm/client/cli/query.go
```diff
@@ -35,10 +36,31 @@ func GetQueryCmd() *cobra.Command {
 		GetCmdGetContractHistory(),
 		GetCmdGetContractState(),
 		GetCmdListPinnedCode(),
+		GetCmdLibVersion(),
```
alpe, line 39: I was expecting this at the command root level, too but I can see that it is easier to maintain this way 👍

alpe (approved):
Very nice! Thanks for the PR 💐

x/wasm/client/cli/query.go
webmaster128: Thank you! Will move to the root. The `./api` package will become internal soon (https://github.com/CosmWasm/wasmvm/issues/333).

PR #847, opened 2022-05-09
Proto updates

alpe:
* Update protobuf files for [ibc v3](https://github.com/cosmos/ibc-go/tree/v3.0.0/proto) 
* Update protobuf files for [Tendermint v0.34.19](https://github.com/tendermint/tendermint/tree/v0.34.19/proto/tendermint)
* Fixed proofs.proto path
* Ran `make proto-all` which includes formatting


ethanfrey (approved):
Looks good.

PR #848, opened 2022-05-09
Ica with demo controller

ethanfrey:
Full implementation of an ICA controller. Thanks to IBC team.

Replaces #839


alpe (approved):
👍 very nice!
disclaimer: I have looked at integration for app.go but not into intertx details

PR #850, opened 2022-05-09
Enforce default code permissions

### LINKED ISSUE #761: Reduce gas when using governance proposals to store wasm
wasm contracts tend to be 200-300kb which is large and expensive.

In order to reduce gas, we already zip the wasm on `wasmd tx wasm store` in order to reduce the tx size by 3x. We never write the wasm to the iavl tree, but to disk directly, where it is handled by Rust code later.

However, I do not believe we perform these optimisations on `wasmd tx gov submit-proposal wasm-store`. We should check.

Also, even if we reduce the tx size, I am almost certain that when we create the proposal, we store the unzipped wasm. Once we validate the syntax (in memory), we can continue to store the original zip data in the proposal (in the IAVL tree) and unzip it when processing a passed proposal before sending it to the Rust library.

Context: when testing this on osmosis, I got surprisingly high gas costs for creating this proposal and for depositing on it (which read and wrote the proposal)

### LINKED ISSUE #840: Simpler upload/instantiate flow for permissioned chains
A proposed solution to solve https://github.com/CosmWasm/wasmd/issues/761

Once #796 is merged, governance can adjust the instantiation config after the fact.

Since voting on the actual "store wasm" is a huge blob and lots of gas, I could imagine using that to achieve a much simpler process for permissioned contracts.

1. Anyone can upload a contract. (7 million gas of so)
2. Default InstantiatePermission is Nobody, so the code cannot be run
3. Make a proposal to allow everyone (or just you) to instantiate the contract (with normal gas costs)
4. If passed, you can now instantiate the contract, which is the key point "permissioned cosmwasm" seeks to control

This would replace the current flow:

1. Default CodeUpload permission is nobody
2. Make a proposal to upload some code (voting is very expensive... several million gas), along with the instantiateconfig
3. If passed, everyone (or just you) can now use this code

There is one last little gotcha... We currently allow the user to override the DefaultInstantiatePermssion upon code upload, so they could easily bypass control 2. However, if we make one change, this would work:

**If InstantiatePermission is set on CodeUpload, it must be equal to or stricter than the default**

* If the default is everyone, you can set anything there.
* If the default is OnlyAddress(x), you can either set the same OnlyAddress (but not changing the address), or Nobody
* If the default is Nobody, you can only set nobody.
* If not set, it will use the default.

This is a very minor change, but coupled with #796 we should solve many of the usability issues around uploading contracts on permsisioned CosmWasm chains.

ethanfrey:
Closes #840 
Closes #761

We now define the `InstantiateDefaultPermission` as a restriction on the contracts, not a "helpful default".
That means, it is used if you do not set a permission explicitly, but it also provide a "maximum bound" on how permissive you can be.

That is, if `InstantiateDefaultPermission` is `Everybody`, you can set any permission you wish (permissionless cosmwasm)
But if `InstantiateDefaultPermission` is `Nobody`, you can only set permission to Nobody on upload and must request governance to update that to allow instantiation (permissioned cosmwasm)

This minor logic change now allows us to open up code upload to everybody (which cost $$$ gas when it was a proposal) and just have a vote on updating the instantiate permissions (implemented in #796), which is much cheaper to vote on. Thus resolving the issue of very expensive votes on permssioned chains #761 while still allowing them to control what code is executed.

@jhernandezb Happy for feedback here as well

### DIFF
--- a/x/wasm/keeper/keeper.go
+++ b/x/wasm/keeper/keeper.go
@@ -159,6 +159,15 @@ func (k Keeper) create(ctx sdk.Context, creator sdk.AccAddress, wasmCode []byte,
 	if !authZ.CanCreateCode(k.getUploadAccessConfig(ctx), creator) {
 		return 0, sdkerrors.Wrap(sdkerrors.ErrUnauthorized, "can not create code")
 	}
+	// figure out proper instantiate access
+	defaultAccessConfig := k.getInstantiateAccessConfig(ctx).With(creator)
+	if instantiateAccess == nil {
+		instantiateAccess = &defaultAccessConfig
+	} else if !instantiateAccess.IsSubset(defaultAccessConfig) {
+		// we enforce this must be subset of default upload access
+		return 0, sdkerrors.Wrap(sdkerrors.ErrUnauthorized, "instantiate access must be subset of default upload access")
+	}
+
 	wasmCode, err = ioutils.Uncompress(wasmCode, uint64(types.MaxWasmSize))
 	if err != nil {
 		return 0, sdkerrors.Wrap(types.ErrCreateFailed, err.Error())
@@ -175,10 +184,6 @@ func (k Keeper) create(ctx sdk.Context, creator sdk.AccAddress, wasmCode []byte,
 	}
 	codeID = k.autoIncrementID(ctx, types.KeyLastCodeID)
 	k.Logger(ctx).Debug("storing new contract", "features", report.RequiredFeatures, "code_id", codeID)
-	if instantiateAccess == nil {
-		defaultAccessConfig := k.getInstantiateAccessConfig(ctx).With(creator)
-		instantiateAccess = &defaultAccessConfig
-	}
 	codeInfo := types.NewCodeInfo(checksum, creator, *instantiateAccess)
 	k.storeCodeInfo(ctx, codeID, codeInfo)
 
--- a/x/wasm/keeper/keeper_test.go
+++ b/x/wasm/keeper/keeper_test.go
@@ -192,8 +192,10 @@ func TestEnforceValidPermissionsOnCreate(t *testing.T) {
 
 	deposit := sdk.NewCoins(sdk.NewInt64Coin("denom", 100000))
 	creator := keepers.Faucet.NewFundedAccount(ctx, deposit...)
+	other := keepers.Faucet.NewFundedAccount(ctx, deposit...)
 
 	onlyCreator := types.AccessTypeOnlyAddress.With(creator)
+	onlyOther := types.AccessTypeOnlyAddress.With(other)
 
 	specs := map[string]struct {
 		defaultPermssion    types.AccessType
@@ -213,6 +215,11 @@ func TestEnforceValidPermissionsOnCreate(t *testing.T) {
 			requestedPermission: nil,
 			grantedPermission:   types.AccessConfig{Permission: types.AccessTypeEverybody},
 		},
+		"explicitly set everybody": {
+			defaultPermssion:    types.AccessTypeEverybody,
+			requestedPermission: &types.AccessConfig{Permission: types.AccessTypeEverybody},
+			grantedPermission:   types.AccessConfig{Permission: types.AccessTypeEverybody},
+		},
 		"cannot override nobody": {
 			defaultPermssion:    types.AccessTypeNobody,
 			requestedPermission: &onlyCreator,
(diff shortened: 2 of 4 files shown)


jhernandezb:
This looks good to me, will solve current issues with the gov system. We may need to add some migration notes besides Stargaze and Osmosis I'm not aware of which other chains have permissioned setup. But when upgrading chains will have to add a migration that:

```
StoreCodePermissions > Nobody-> Everybody
InstantiatePermissions > Everybody -> Nobody
```

ethanfrey:
Yes, we should use some migrate notes.

That said, the old settings will continue to work, we just provide a second (more gas-efficient) way to achieve this. And the chain can adjust when they are ready

x/wasm/types/types.go
```diff
@@ -330,3 +330,21 @@ func VerifyAddressLen() func(addr []byte) error {
 		return nil
 	}
 }
+
+// IsSubset will return true if the caller is the same as the superset,
+// or if the caller is more restrictive than the superset.
+func (a AccessConfig) IsSubset(superSet AccessConfig) bool {
```
alpe, line 336: 👍  good design to add this to AccessConfig

x/wasm/types/types_test.go
```diff
@@ -372,3 +372,79 @@ func TestVerifyAddressLen(t *testing.T) {
...
+		"only > nobody": {
+			superSet: AccessConfig{Permission: AccessTypeNobody},
+			check:    AccessConfig{Permission: AccessTypeOnlyAddress, Address: "foobar"},
+			isSubSet: false,
+		},
+		"everybody > nobody": {
+			superSet: AccessConfig{Permission: AccessTypeNobody},
+			check:    AccessConfig{Permission: AccessTypeEverybody},
+			isSubSet: false,
+		},
```
alpe, line 396: nit: to be complete "undefined !< nobody". For other target types as well

x/wasm/types/types_test.go
```diff
@@ -372,3 +372,79 @@ func TestVerifyAddressLen(t *testing.T) {
 		})
 	}
 }
+
+func TestAccesConfigSubset(t *testing.T) {
```
alpe, line 376: 🏅 for all the testing

x/wasm/keeper/keeper.go
```diff
@@ -159,6 +159,15 @@ func (k Keeper) create(ctx sdk.Context, creator sdk.AccAddress, wasmCode []byte,
 	if !authZ.CanCreateCode(k.getUploadAccessConfig(ctx), creator) {
 		return 0, sdkerrors.Wrap(sdkerrors.ErrUnauthorized, "can not create code")
 	}
+	// figure out proper instantiate access
```
alpe, line 162: Small change big impact 👏

x/wasm/keeper/keeper_test.go
```diff
@@ -183,6 +183,62 @@ func TestCreateWithParamPermissions(t *testing.T) {
...
+		"cannot override nobody": {
+			defaultPermssion:    types.AccessTypeNobody,
+			requestedPermission: &onlyCreator,
+			expError:            sdkerrors.ErrUnauthorized,
+		},
+		"default to nobody": {
+			defaultPermssion:    types.AccessTypeNobody,
+			requestedPermission: nil,
+			grantedPermission:   types.AccessConfig{Permission: types.AccessTypeNobody},
+		},
```
alpe, line 225: nit: you don't have a case where requestedPermission < defaultPermission .

x/wasm/keeper/keeper_test.go
```diff
@@ -183,6 +183,62 @@ func TestCreateWithParamPermissions(t *testing.T) {
...
+	keeper := keepers.WasmKeeper
+	contractKeeper := keepers.ContractKeeper
+
+	deposit := sdk.NewCoins(sdk.NewInt64Coin("denom", 100000))
+	creator := keepers.Faucet.NewFundedAccount(ctx, deposit...)
+
+	onlyCreator := types.AccessTypeOnlyAddress.With(creator)
+
+	specs := map[string]struct {
+		defaultPermssion    types.AccessType
```
alpe, line 199: nit: 
```suggestion
		defaultPermission    types.AccessType
```

x/wasm/keeper/keeper_test.go
```diff
@@ -183,6 +183,62 @@ func TestCreateWithParamPermissions(t *testing.T) {
...
+			defaultPermssion:    types.AccessTypeEverybody,
+			requestedPermission: &onlyCreator,
+			grantedPermission:   onlyCreator,
+		},
+		"default to everybody": {
+			defaultPermssion:    types.AccessTypeEverybody,
+			requestedPermission: nil,
+			grantedPermission:   types.AccessConfig{Permission: types.AccessTypeEverybody},
+		},
+		"cannot override nobody": {
```
alpe, line 216: 👍  very important test

alpe (approved):
Excellent work! 
I appreciate the effort you spent in testing this. I added some nits for bonus points but feel free to merge as is

x/wasm/keeper/keeper_test.go
ethanfrey: I do... that is what "override everyone" tests (request more restrictive)
and "cannot override nobody" (request less restrictive)

I can add more, but these cover the general cases, and the actual compare is heavily tested elsewhere.

If you have a concrete default/requested I should add, let me know. I will just add one for equals

PR #851, opened 2022-05-09
Gofumpt

ethanfrey:
Replaces #832 (which was a great idea, thanks @faddat )

This has 2 commits to install the proper tooling and use it with the `make format` and `make lint` commands.

The last commit actually runs the updated `make format` command, and `make lint` will help enforce this as a requirement for merging.


alpe (approved):
LGTM! 
I got the same output when running this against main

Issue #815, opened 2022-04-25
Upgrade to Go v1.18


alpe:
Tracking issue
Cosmos-SDK main branch is already upgraded to use [Go 1.18](https://tip.golang.org/doc/go1.18)

> Generics ... 
These new language changes required a large amount of new code that has not had significant testing in production settings. That will only happen as more people write and use generic code. We believe that this feature is well implemented and high quality. However, unlike most aspects of Go, we can't back up that belief with real world experience. Therefore, while we encourage the use of generics where it makes sense, please use appropriate caution when deploying generic code in production.

alpe:
Done in https://github.com/CosmWasm/wasmd/pull/866

PR #866, opened 2022-05-18
close 815: upgrade to go 1.18

faddat:
This PR upgrades go to version 1.18


alpe (commented):
Thanks for the PR. You are bleeding edge as always and I appreciate your energy and effort!

When we discussed this internally, the plan was to be a bit more conservative and follow the sdk roadmap on this.
The sdk [v0.46.x](https://github.com/cosmos/cosmos-sdk/blob/release/v0.46.x/go.mod#L1) and v0.45.x are still on Go 1.17.
 
IMHO this upgrade needs more time to settle but please comment on #815 if you have any requirements or see other chains upgrading to Go 1.18. This will help to get more confidence.

Other than this, your PR looks good. Just update the circle-ci image and Readme.

jhernandezb:
We have been running stargaze with go 1.18 without issues so far for a month probalby, we have not introduced any of the new features (generics) to the codebase tho.

faddat:
Given (whatever this is)

https://app.circleci.com/pipelines/github/CosmWasm/wasmd/2111/workflows/7a47baf1-f484-4803-891d-eac1de26f0a5/jobs/7124

could we consider moving to plain github actions?

alpe:
The new circleCI image is unfortunately not a drop in replacement. I have opened #872 and will 👀  when I get a chance

faddat:
Kinda my point about migrating to actions.  

Also, if we use actions, the whole repo becomes more easily forkable.

faddat:
got it :)

ethanfrey:
@alpe what is the status here? I was happy not to merge before v0.27.0 (there was plenty in there already), but what about merging this now, to include in v0.28.0.

Cosmos-sdk v0.46.0-rc1 us on Go 1.18 and Stargaze's experience is also positive (as they use CosmWasm and the whole stack)

scripts/protocgen.sh
```diff
@@ -8,7 +8,7 @@ protoc_gen_gocosmos() {
     return 1
   fi
 
-  go get github.com/regen-network/cosmos-proto/protoc-gen-gocosmos@latest 2>/dev/null
+  go install github.com/regen-network/cosmos-proto/protoc-gen-gocosmos@latest 2>/dev/null
```
alpe, line 11: This is not working with the docker image (tendermintdev/sdk-proto-gen) that we use. Can you revert this change/ test with `make proto-gen` ?

alpe (changes requested):
Thank you very much for the PR! 🌻 
Everything looks great except that `make proto-gen` is broken in this PR. If you can revert that change in protocgen.sh, I would merge this asap, now.

alpe (approved):
I can not expect you to work on this old PR to fix one line. It is also blocking the sdk upgrade, so I will merge it as is and provide a patch myself. Thanks again for your work!

PR #869, opened 2022-05-19
Publish protobufs on release tags

pinosu:
Publish protobufs when:
- A new commit is pushed into main (only if files in proto folder are modified)
- A new release tag event is triggered

### DIFF
--- a/.github/workflows/proto-buf-publisher.yml
+++ b/.github/workflows/proto-buf-publisher.yml
@@ -4,9 +4,13 @@ name: Proto Buf Publishing - Action
 on:
   push:
     branches:
-      - master
+      - main
     paths:
       - 'proto/**'
+
+    # Sequence of patterns matched against refs/tags
+    tags:
+      - "v[0-9]+.[0-9]+.[0-9]+" # Push events to matching v*, i.e. v1.0, v20.15.10
 
 jobs:
   push:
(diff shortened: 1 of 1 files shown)


pinosu:
@alpe [this](https://github.com/cosmos/cosmos-sdk/issues/12003#issue-1242322013) is how cosmos-sdk will approach protobuf publishing. Should we follow the same path or we go with this solution since we might have backwards compatibility breaking changes and later on upgrade to the same approach adopted by cosmos-sdk?

.github/workflows/proto-buf-publisher.yml
```diff
@@ -4,10 +4,14 @@ name: Proto Buf Publishing - Action
...
   push:
     branches:
-      - master
+      - main
     paths:
       - 'proto/**'
 
+    # Sequence of patterns matched against refs/tags
+    tags:
+      - "v[0-9]+.[0-9]+.[0-9]+" # Push events to matching v*, i.e. v1.0, v20.15.10
```
alpe, line 13: nit: let's publish for RC versions as well, so that people can integrate/ test early
```suggestion
      - "v[0-9]+.[0-9]+.[0-9]+.*" # Push events to matching v*, //i.e. v20.15.10, v0.27.0-rc1
```

alpe (approved):
Looks good. Minor comment for RC versions

PR #873, opened 2022-05-27
Sdk46 no global

faddat:
This is just a review tool.

I write app.go with 

app.BankKeeper

but here we do app.bankKeeper

going to have a look see around to see how that affects things


ethanfrey:
> I write app.go with
> app.BankKeeper
> but here we do app.bankKeeper

We currently have all members of WasmApp lowercase, other chains have them Capitalised (public). This seems to cause most of the merge conflicts with your PRs (between wasmd and Juno / Osmosis).

I am happy to reconsider that design. If you want to raise an issue just on that point, we can talk, and if we then merge a PR just to do the rename, this will help with other merge issues in the future. I would love to hear Alex's thoughts on making them public as well, so probably best to have a focused issue, then we decide to change or not.

x/wasm/types/codec.go
```diff
@@ -70,6 +70,7 @@ var (
 )
 
 func init() {
+	RegisterLegacyAminoCodec(legacy.Cdc)
```
giansalex, line 73: this seems necessary for authz + ledger

https://github.com/cosmos/cosmos-sdk/blob/v0.46.0-rc1/x/bank/types/codec.go#L47

faddat:
@ethanfrey I'll make that issue now, and thanks very much for your openness to discussion on the matter.

❤️

PR #875, opened 2022-05-29
feat: setup and run script in dockerfile

j0nl1:
The reason to introduce this script is allow you to configure some tests account and run wasmd from a single docker command. There wouldn't need to create a docker volume and it will improve CI tests.


alpe (approved):
LGTM 🌻

PR #879, opened 2022-06-01
Remove unused NegotiateAppVersion ibc callback

alpe:
> Now applications will perform this version negotiation during the channel handshake, thus removing the need for NegotiateAppVersion.

https://github.com/cosmos/ibc-go/blob/main/docs/migrations/v2-to-v3.md#negotiateappversion-removed-from-ibcmodule-interface


ethanfrey (approved):
Good catch

PR #863, opened 2022-05-17
Cosmos SDK v46

faddat:
Hi, so here's another CW for 46.

I am mainly interested in seeing how much better or worse we can get CW-enabled chains to run in high traffic scenarios.

I think this contributed to the fall of terra.  

After some debugging and refinement, hope to get this out on Juno and Osmosis.

PR #882, opened 2022-06-02
use ecosystem convention for keeper names in app.go

faddat:
This PR changes keeper names in app.go to match the commonly used ecosystem conventional style, which:

* Makes wasmd easier to reference
* Makes upgrading wasmd easier, because there will be fewer merge conflicts

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -243,33 +243,33 @@ type WasmApp struct {
 	memKeys map[string]*sdk.MemoryStoreKey
 
 	// keepers
-	accountKeeper       authkeeper.AccountKeeper
-	bankKeeper          bankkeeper.Keeper
-	capabilityKeeper    *capabilitykeeper.Keeper
-	stakingKeeper       stakingkeeper.Keeper
-	slashingKeeper      slashingkeeper.Keeper
-	mintKeeper          mintkeeper.Keeper
-	distrKeeper         distrkeeper.Keeper
-	govKeeper           govkeeper.Keeper
-	crisisKeeper        crisiskeeper.Keeper
-	upgradeKeeper       upgradekeeper.Keeper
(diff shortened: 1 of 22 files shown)


faddat:
My preference here is to first:

* merge #883 
* merge #866 

The specific reason is that go 1.18 will change some of the supported linters.  We could then dial in on the ones that we think are valuable, and also exclude anything that isn't.

alpe:
Thanks for bringing this up! 🌻 

I have seen public fields in other apps, too. I was always wondering why break encapsulation without a need.
I can see public fields being helpful with the test suite pattern used in sdk and elsewhere. Do you have other examples?

> Makes wasmd easier to reference

what do you mean with this?

faddat:
@alpe - the practice that I openly recommend to other developers when constructing app.go is to very literally copy and paste.

re: encapsulation, and having a standard on these-- I have wondered same as yourself.

ethanfrey:
> from the perspective of someone working on Juno, who'd like to move changes between wasmd and juno, adopting this standard would make upstreaming work into wasmd easier.

I agree :100: with you that we should have a standard and the same format in wasmd, osmosis and Juno.

However, Alex's question was why Juno's approach is better than the wasmd approach. Maybe you or another Juno dev can explain why these are needed to be public. (I have some ideas, but I prefer feedback from actual users)

faddat:
@ethanfrey honestly I don't see a reason in either direction, 🤣

faddat:
OK, this one is ready to rock.  After merging this and #815, the review scope for #863 will drop a great deal.

alpe:
Thanks for the PR and bringing this topic up! I was pending for a while now due to my other priorities but I polished your commits in #951.

PR #883, opened 2022-06-02
Update config.yml

faddat:
This is a micro-PR that only brings the golangci-lint binary that we use in CI current with their latest release.

I think it may be helpful with:

* #882 
* #815
* #866


alpe (commented):
👍 Thanks

alpe (approved):

PR #892, opened 2022-06-21
Update on README.md

miguelgargallo:
Correction on the first step of, clone git, in case someone needs the correct git clone command, or just want to make it easier, git clone command is written here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,13 @@ your use case.
 
 ## Quick Start
 
+## Clone securely the repo
+
+```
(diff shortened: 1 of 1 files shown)


alpe:
Thanks for the contribution! 🌻 
With git clone the user is not in the right directory. Would you mind to add a `cd wasmd` ?

alpe:
Closing this now as the doc starting on checkout is still incomplete.
IMHO it is fair to assume that people can checkout a project from github and we build our doc on top of that.

PR #898, opened 2022-07-07
Charge gas to unzip wasm code

alpe:
Uncompressing packed wasm code should be paid in gas to compensate for CPU.

~~Early version for feedback. Left to do:
**- Set proper cost per byte value**~~

Not included is an ante handler to verify gas. But we don't do this for costs per bytes either.


x/wasm/keeper/gas_register.go
```diff
@@ -52,6 +52,8 @@ const (
 	DefaultPerCustomEventCost uint64 = 20
 	// DefaultEventAttributeDataFreeTier number of bytes of total attribute data we do not charge.
 	DefaultEventAttributeDataFreeTier = 100
+	// DefaultPerByteUncompressCost is how much SDK gas we charge per source byte to unpack
+	DefaultPerByteUncompressCost uint64 = 2
```
webmaster128, line 56: You can make this of type `wasmvmtypes.UFraction` and change to 15/100 (=0.15).

1 gas or more per byte is very expensive for gzip decompression.

x/wasm/ioutils/utils.go
```diff
@@ -17,7 +17,7 @@ var (
 
 // IsGzip returns checks if the file contents are gzip compressed
 func IsGzip(input []byte) bool {
-	return bytes.Equal(input[:3], gzipIdent)
+	return len(input) > 3 && bytes.Equal(gzipIdent, input[0:3])
```
webmaster128, line 20:
```suggestion
	return len(input) >= 3 && bytes.Equal(gzipIdent, input[0:3])
```

Before we accepted a length of 3 as valid gzip. I'd keep it that way and leave it up to the uncompressor to check everything else. Otherwise we should add a comment how much data we require after the `gzipIdent` and why.

x/wasm/keeper/keeper.go
```diff
@@ -178,11 +178,14 @@ func (k Keeper) create(ctx sdk.Context, creator sdk.AccAddress, wasmCode []byte,
...
+		ctx.GasMeter().ConsumeGas(k.gasRegister.UncompressCosts(len(wasmCode)), "Uncompress gzip bytecode")
+		wasmCode, err = ioutils.Uncompress(wasmCode, uint64(types.MaxWasmSize))
+		if err != nil {
+			return 0, sdkerrors.Wrap(types.ErrCreateFailed, err.Error())
+		}
 	}
-	ctx.GasMeter().ConsumeGas(k.gasRegister.CompileCosts(len(wasmCode)), "Compiling WASM Bytecode")
+	ctx.GasMeter().ConsumeGas(k.gasRegister.CompileCosts(len(wasmCode)), "Compiling wasm bytecode")
 
 	checksum, err := k.wasmVM.Create(wasmCode)
```
webmaster128, line 190:
```suggestion

	ctx.GasMeter().ConsumeGas(k.gasRegister.CompileCosts(len(wasmCode)), "Compiling wasm bytecode")
	checksum, err := k.wasmVM.Create(wasmCode)
```

Just a whitespace diff. Let's move this to the block below where it belongs to (the compiling step)

x/wasm/keeper/snapshotter.go
```diff
@@ -99,6 +99,9 @@ func (ws *WasmSnapshotter) Restore(
 }
 
 func restoreV1(ctx sdk.Context, k *Keeper, compressedCode []byte) error {
+	if !ioutils.IsGzip(compressedCode) {
+		return types.ErrInvalid.Wrap("not a gzip")
+	}
```
webmaster128, line 104: Before `ioutils.Uncompress` accepted uncompressed inputs. This diff makes gzip mandatory. Is this intended?

webmaster128 (commented):
Looks good. Some questions and suggestions.

faddat:
The reasoning behind this one is super-solid.

x/wasm/keeper/snapshotter.go
alpe: yes as in L80 the snapshotter writes only gzipped data to the stream. Any non gzipped data is not from us and should be rejected.

alpe:
I have applied the feedback and rebased the branch on top of main.

x/wasm/ioutils/ioutil.go
```diff
@@ -8,18 +8,12 @@ import (
...
 
-// Uncompress returns gzip uncompressed content if input was gzip, or original src otherwise
-func Uncompress(src []byte, limit uint64) ([]byte, error) {
-	switch n := uint64(len(src)); {
-	case n < 3:
-		return src, nil
-	case n > limit:
+// Uncompress expects a valid gzip source to unpack or fails. See IsGzip
+func Uncompress(gzipSrc []byte, limit uint64) ([]byte, error) {
+	if uint64(len(gzipSrc)) > limit {
```
pinosu, line 13: 👍

x/wasm/keeper/gas_register.go
```diff
@@ -54,12 +54,26 @@ const (
 	DefaultEventAttributeDataFreeTier = 100
 )
 
+// default: 0.15 gas.
+// see https://github.com/CosmWasm/wasmd/pull/898#discussion_r937727200
+var defaultPerByteUncompressCost = wasmvmtypes.UFraction{
```
pinosu, line 59: 👍

x/wasm/keeper/gas_register_test.go
```diff
@@ -430,3 +432,32 @@ func TestFromWasmVMGasConversion(t *testing.T) {
 		})
 	}
 }
+
+func TestUncompressCosts(t *testing.T) {
```
pinosu, line 436: nice!

x/wasm/keeper/keeper.go
```diff
@@ -178,12 +178,15 @@ func (k Keeper) create(ctx sdk.Context, creator sdk.AccAddress, wasmCode []byte,
 		return 0, sdkerrors.Wrap(sdkerrors.ErrUnauthorized, "instantiate access must be subset of default upload access")
 	}
 
-	wasmCode, err = ioutils.Uncompress(wasmCode, uint64(types.MaxWasmSize))
-	if err != nil {
-		return 0, sdkerrors.Wrap(types.ErrCreateFailed, err.Error())
+	if ioutils.IsGzip(wasmCode) {
```
pinosu, line 181: 👍

pinosu (approved):
Good work! 💯

webmaster128 (approved):
🐎

PR #899, opened 2022-07-08
Remove unnecessary string

shiki-tak:
The error message contains an unnecessary string, so correct contract address is not output. 
I have removed it.


alpe (approved):
Very nice! Thanks 👍

PR #900, opened 2022-07-08
Prevent migration to a restricted code

alpe:
On migration, the permissions for the target contract were not checked proper.


pinosu (approved):
very nice! lgtm 👍

PR #903, opened 2022-07-12
Add query for the total supply of a coin

larry0x:
currently it is not possible to query the total supply of a native SDK coin. this is something that protocols often need to do, especially that now contract are able to mint native coins thanks to the `tokenfactory` module. this PR adds the binding for this query type.

see also:
* https://github.com/CosmWasm/wasmvm/pull/337
* https://github.com/CosmWasm/cosmwasm/pull/1356

**NOTE: the CI tests fail because they depend on the PR for `wasmvm` being merged first**

### DIFF
--- a/x/wasm/keeper/query_plugins.go
+++ b/x/wasm/keeper/query_plugins.go
@@ -189,6 +189,16 @@ func BankQuerier(bankKeeper types.BankViewKeeper) func(ctx sdk.Context, request 
 			}
 			return json.Marshal(res)
 		}
+		if request.Supply != nil {
+			coin := bankKeeper.GetSupply(ctx, request.Supply.Denom)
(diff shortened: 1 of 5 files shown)


larry0x:
the CI tests fail because they depend on the PR for `wasmvm` being merged first

faddat:
I love this.

x/wasm/keeper/query_plugins.go
```diff
@@ -189,6 +189,16 @@ func BankQuerier(bankKeeper types.BankViewKeeper) func(ctx sdk.Context, request
 			}
 			return json.Marshal(res)
 		}
+		if request.Supply != nil {
```
alpe, line 192: This looks good. Some happy path test with the mock would be good for regression

PR #905, opened 2022-07-19
Remove unnecessary error check

shiki-tak:
The runAsAddr variable does not exist in the latest version, but only the error check remains.
I think it is unnecessary, right?

### DIFF
--- a/x/wasm/keeper/proposal_handler.go
+++ b/x/wasm/keeper/proposal_handler.go
@@ -107,9 +107,7 @@ func handleMigrateProposal(ctx sdk.Context, k types.ContractOpsKeeper, p types.M
 	if err != nil {
 		return sdkerrors.Wrap(err, "contract")
 	}
-	if err != nil {
-		return sdkerrors.Wrap(err, "run as address")
-	}
+
 	// runAs is not used if this is permissioned, so just put any valid address there (second contractAddr)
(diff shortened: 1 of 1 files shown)


x/wasm/keeper/proposal_handler.go
```diff
@@ -107,9 +107,7 @@ func handleMigrateProposal(ctx sdk.Context, k types.ContractOpsKeeper, p types.M
 	if err != nil {
 		return sdkerrors.Wrap(err, "contract")
 	}
-	if err != nil {
-		return sdkerrors.Wrap(err, "run as address")
-	}
+
```
peterbourgon, line 110: If a function returns an error, that error must be checked.

peterbourgon (changes requested):

x/wasm/keeper/proposal_handler.go
shiki-tak: Function indicates `sdk.AccAddressFromBech32`?
If so, wouldn't the error check above suffice?
(https://github.com/CosmWasm/wasmd/pull/905/files#diff-a5c1aab521b573562de4814c157ce3c9b7564672c24f0acafffdb003de9c16fbR107-R109)

peterbourgon: Ah, there are duplicate error checks here — all good 👍

alpe (approved):
Good 👁️ . Thanks for the patch! 👍

PR #909, opened 2022-07-21
Standardized Development Environment

faddat:
## Description

Yesterday I made some changes to the standard dev env that we're using and to my surprise, they "just worked".  That includes adding docker to the image so that we could do protocol buffer builds.  It also has all the protobuf tooling installed in the filesystem for those (like myself in fact!) with container allergies. 

```yaml
image: ghcr.io/notional-labs/cosmos
```

That image is built here daily using cron via github actions and each build is observable. 

https://github.com/notional-labs/containers/blob/master/cosmos/Dockerfile

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,0 +1,1 @@
+image: github.com/notional-labs/cosmos


alpe:
Just curious, can you elaborate a bit why this adds value and how you are using it?
I have just logged into gitpod.io and the `make proto-all` works fine on their shell without this.

alpe:
I found the same at [gaia](https://github.com/cosmos/gaia/blob/main/.gitpod.yml#L4) and [juno](https://github.com/CosmosContracts/juno/blob/main/.gitpod.yml#L5). 🏅 
I guess, I need to try gitpod one day to see how this works.

PR #914, opened 2022-07-28
emergency patch to prevent non-determinisitic reply input

dimiandre:
emergency patch to prevent non-determinisitic reply input


angelorc:
you should use the tag `v0.27.0-reply-patch`

PR #916, opened 2022-07-28
Add sorting fix from terra devs

the-frey:
It may be that @assafmo's fix is enough, but also contributing this upstream as we are going to roll with it on a fork. Credit of course goes to the Terra folks for spotting the non-determinism issue there.

### DIFF
--- a/x/wasm/keeper/keeper.go
+++ b/x/wasm/keeper/keeper.go
@@ -500,6 +500,16 @@ func (k Keeper) reply(ctx sdk.Context, contractAddress sdk.AccAddress, reply was
 	// prepare querier
 	querier := k.newQueryHandler(ctx, contractAddress)
 	gas := k.runtimeGasForContract(ctx)
+	if reply.Result.Ok != nil {
+		events := reply.Result.Ok.Events
+		for _, e := range events {
+			attributes := e.Attributes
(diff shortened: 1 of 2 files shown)


x/wasm/keeper/keeper.go
```diff
@@ -500,6 +501,16 @@ func (k Keeper) reply(ctx sdk.Context, contractAddress sdk.AccAddress, reply was
 	// prepare querier
 	querier := k.newQueryHandler(ctx, contractAddress)
 	gas := k.runtimeGasForContract(ctx)
+	if reply.Result.Ok != nil {
+		events := reply.Result.Ok.Events
+		for _, e := range events {
+			attributes := e.Attributes
+			sort.SliceStable(attributes, func(i, j int) bool {
+				return bytes.Compare([]byte(attributes[i].Key), []byte(attributes[j].Key)) == -1
```
peterbourgon, line 509:
```suggestion
				return attributes[i].Key < attributes[j].Key
```

ethanfrey:
I am unsure this is needed. Please DM me where any possible reordering is possible one we restrict to just wasm messages. Otherwise, I would avoid this complexity.

ethanfrey:
Replacing this with #918 

Thank you very much for the contribution

PR #918, opened 2022-07-29
Sorting fix

ethanfrey:
Replaces #916 (Thank you [the-frey](https://github.com/the-frey))

Rebased and using suggestion. Made mods myself to get this out faster. And unified it with #917 so both changes are together


ueco-jb:
https://github.com/CosmWasm/wasmd#compatibility also needs a new entry if you're going to release a new version.

PR #919, opened 2022-08-02
Upgrade ibc-go to v4.0.0-rc2 and go to 1.18

devashishdxt:


faddat:
hey, :) -- this is great.

I hope you don't mind:

I am going to take my 457 branch, and apply your changes to it (then it'll be v4.0.0)

faddat:
Hey David, just one issue in this code that I saw:

instead of the *rand.Rand, you actually want 

txCtx.R

teacher: pain

alpe:
Thank you for your contribution and apologies for keeping this pending for such a long time! We follow a more conservative strategy with core dependency upgrades like IBC or SDK. Therefore upgrades of these modules can take longer until they are proven to be stable on production chains. This can be frustrating but our focus is on stability and focusing our energy on wasmd. Nevertheless your contribution was valuable and inspired others to work with this version!

Meanwhile there were more IBC version upgrades and I opened https://github.com/CosmWasm/wasmd/pull/1088 to catch up.

PR #933, opened 2022-08-09
tendermint v0.34.20

faddat:
I'm learning to make smaller PR's.

This **only** bumps tendermint to v0.34.20


alpe:
Thanks for taking care! Small PRs are are very much appreciated. In this case, I would follow the SDK do the v0.45.7 upgrade , which includes this

PR #934, opened 2022-08-09
remove io/ioutil

faddat:
ioutil was deprecated in go 1.16 and is an error in go 1.19


faddat:
this is also ready to go.

alpe (approved):
Very nice! Thank you for doing the 🧹.  💪

PR #937, opened 2022-08-10
ibc-go v3.2.0

faddat:
OK I think this one does it :)

* upgrades ibc-go to v3.2.x

PR #940, opened 2022-08-15
ibc-go v4.0.0

faddat:
This PR riffs on these PR's:

* https://github.com/CosmWasm/wasmd/pull/937
* https://github.com/CosmWasm/wasmd/pull/919

To create a v0.45.7 wasmd that also uses ibc-go v4.0.0 and would be most suited to Juno's v10 upgrade.

Thanks to David, author of #919 

Note: if we merge this, then #937 should be closed.   This is my preference for merging, so I've updated it, but please let me know if there's a preference for #937.  Thanks!

## Update

Since main is updated to v45.7, this branch now serves only as an upgrade to ibc-go v4.0.0

It also changes the version of golangci-lint from v1.47.0 to v1.47 so that we catch all go 1.18 compatible updates in golangci-lint (eg 1.47.*)


faddat:
We should not do this.

larry0x:
> We should not do this.

@faddat why not?

Issue #913, opened 2022-07-25
wasmd, make install fails on MacOSX 11.6.7


4coffeeihadbreakfast:
Hi, wasmd make install fails for me on MacOSX 11.6.7

Following along in [dev-academy setup](https://docs.cosmwasm.com/dev-academy/basics/environment) docs -> dev-academy -> Basics -> Environment Setup -> wasmd -> make install

I am unable to install wasmd

**env:**
```
$ go version
go version go1.18.4 darwin/amd64

$ uname -a
Darwin MacBook-Pro.local 20.6.0 Darwin Kernel Version 20.6.0: Tue Apr 19 21:04:45 PDT 2022; root:xnu-7195.141.29~1/RELEASE_X86_64 x86_64
```

**steps:**
```
git clone https://github.com/CosmWasm/wasmd.git
cd wasmd
git fetch --tags
git checkout v0.27.0
make install
```
```
$ make install
go install -mod=readonly -tags "netgo,ledger" -ldflags '-X github.com/cosmos/cosmos-sdk/version.Name=wasm -X github.com/cosmos/cosmos-sdk/version.AppName=wasmd -X github.com/cosmos/cosmos-sdk/version.Version=0.27.0 -X github.com/cosmos/cosmos-sdk/version.Commit=d63bea442bedf5b3055f3821472c7e6cafc3d813 -X github.com/CosmWasm/wasmd/app.Bech32Prefix=wasm -X "github.com/cosmos/cosmos-sdk/version.BuildTags=netgo,ledger" -L/usr/local/opt/libpq/lib' -trimpath ./cmd/wasmd
go: downloading github.com/cosmos/cosmos-sdk v0.45.4
go: downloading github.com/prometheus/client_golang v1.12.1
go: downloading github.com/tendermint/tendermint v0.34.19
...
...
...
# github.com/keybase/go-keychain
cgo-gcc-prolog:203:11: warning: 'SecTrustedApplicationCreateFromPath' is deprecated: first deprecated in macOS 10.15 - No longer supported [-Wdeprecated-declarations]
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/Frameworks/Security.framework/Headers/SecTrustedApplication.h:59:10: note: 'SecTrustedApplicationCreateFromPath' has been explicitly marked deprecated here
# github.com/CosmWasm/wasmd/cmd/wasmd
[...]

PR #950, opened 2022-08-21
FIX: wasmd, make install fails on MacOSX 11.6.7

4coffeeihadbreakfast:
This PR resolves my [issue](https://github.com/CosmWasm/wasmd/issues/913). Install as (`LDFLAGS= make install`)


alpe:
Thanks a lot for digging into this! As the keyring/keychain dependencies provide access to the account keys, I am very sensible to apply any upgrades. I have opened #957 to discuss and find a version/ fork that fits best.
Please bare with us for some time to get feedback.

alpe:
Thanks for your help again. I am closing this PR in favour of #985

PR #951, opened 2022-08-23
Make app keepers public

### LINKED ISSUE #881: Should we expose Keepers in app.go?
Currently, in app.go we do like:

app.bankKeeper

the most common practice is to do like:

app.BankKeeper

Should we change the way that CosmWasm presents these?

alpe:
Resolves #881

This PR is based on the work from @faddat in PR #882 (💐) with the variable names renaming reverted.
Added deprecation annotation and formatting.

Note to reviewer: please squash and merge

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -243,33 +243,33 @@ type WasmApp struct {
 	memKeys map[string]*sdk.MemoryStoreKey
 
 	// keepers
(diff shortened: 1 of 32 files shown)


pinosu (approved):
LGTM 👍

Issue #957, opened 2022-08-24
Upgrade keyring / go-keychain dependencies


alpe:
We were following the SDK on the keyring + go-keychain dependencies. There had been odd compiler warnings for some time already but not it seems that people have real issues with the versions. 
See #950 and linked issue for context.

in our go.mod
> 	github.com/keybase/go-keychain v0.0.0-20190712205309-48d3d31d256d // indirect
> replace(
>	// Use the cosmos-flavored keyring library
>	github.com/99designs/keyring => github.com/cosmos/keyring v1.1.7-0.20210622111912-ef00f8ac3d76

I did some a quick research what other versions are used. See below. 

## go-keychain
`github.com/99designs/go-keychain v0.0.0-20191008050251-8e49817e8af4` [gaia-main](https://github.com/cosmos/gaia/blob/main/go.mod#L36), [osmosis v11](https://github.com/osmosis-labs/osmosis/blob/v11.0.1/go.mod#L40), [sdk main](https://github.com/cosmos/cosmos-sdk/blob/main/go.mod#L75), #950

`github.com/keybase/go-keychain v0.0.0-20190712205309-48d3d31d256d` [terra v2](https://github.com/terra-money/core/blob/v2.1.1/go.mod#L80), [juno v9](https://github.com/CosmosContracts/juno/blob/v9.0.0/go.mod#L108), [sdk v0.45.7](https://github.com/cosmos/cosmos-sdk/blob/v0.45.7/go.mod#L88), [stargaze v6](https://github.com/public-awesome/stargaze/blob/v6.0.1/go.mod#L80)

## keyring

`github.com/cosmos/keyring v1.2.0` [sdk main](https://github.com/cosmos/cosmos-sdk/blob/main/go.mod#L290)
`github.com/99designs/keyring v1.2.1` [gaia main](https://github.com/cosmos/gaia/blob/main/go.mod#L36), [osmosis v11](https://github.com/osmosis-labs/osmosis/blob/v11.0.1/go.mod#L41)

`github.com/99designs/keyring v1.1.6` [juno v9](https://github.com/CosmosContracts/juno/blob/v9.0.0/go.mod#L27), [terra v2.1](https://github.com/terra-money/core/blob/v2.1.1/go.mod#L26)

[...]

PR #967, opened 2022-08-29
updated go.mod to make install work on Apple M1

sajanjswl:
make install fails on Apple M1 due to the older version used for

github.com/keybase/go-keychain
github.com/zondax/hid

This Pr fixes this.


alpe:
Thanks for the PR! 
We has some similar one recently where I opened https://github.com/CosmWasm/wasmd/issues/957 for discussion.

alpe:
I am closing this in favour of https://github.com/CosmWasm/wasmd/pull/985. Can you confirm that it works for you, when merged? 

I saw that gaia has pinned `github.com/zondax/hid` to the [`v0.9.0`](https://github.com/cosmos/gaia/blob/main/go.mod#L287) release. I don't feel comfortable to upgrade this dependency without more discussion.

PR #968, opened 2022-08-31
Remove obsolete cli gov params

### LINKED ISSUE #849: Remove unused flag in gov proposals
`flagProposalType` is being set in all gov proposals but only used in the sdk for the main `submit-proposal` command https://github.com/cosmos/cosmos-sdk/blob/v0.45.4/x/gov/client/cli/parse.go#L13 

We should consider removing it if it's not actually used. 

Context: https://github.com/CosmWasm/wasmd/pull/796#discussion_r868125317

alpe:
Resolves #849 

I have looked at other proposal types in the sdk and ibc-go and "proposal" + "type" are not required at the subcommand level. Happy to remove them

For example: https://github.com/cosmos/cosmos-sdk/blob/v0.45.8/x/upgrade/client/cli/tx.go#L69-L73


pinosu (approved):
LGTM 👍

PR #971, opened 2022-09-01
Enable whitelisted Stargate Queries

### LINKED ISSUE #904: Use an Allow list for StargateQuery instead of completely disabling it
I was discussing the need to disable StargateQuery in the discord channel to better understand why they're completely disabled (https://discord.com/channels/737637324434833438/737643344712171600/998624457188987030). 

I would like to suggest replacing the current handler with one that contains an allow list. This way any chain that uses wasmd and has the stargate feature enabled, can just customize the allow list for the queries. If the chain takes no action, the behaviour would be equivalent to the current one (https://github.com/CosmWasm/wasmd/blob/7c7840c32a65c79e9b9525ec9488de464cbf78a0/x/wasm/keeper/query_plugins.go#L275). While chains can customize this manually by replacing the stargate querier, I think it feels inconsistent to have `StargateQuery` on cosmwasm and have it be useless in wasmd. 

As a reference, here is the previous DenyList implementation: https://github.com/CosmWasm/wasmd/blob/v0.25.0/x/wasm/keeper/query_plugins.go#L281 (thanks @jhernandezb!)

My interest in this comes from the decision from osmosis to use StargateMsg/StargateQuery for cosmwasm<->osmosis communication instead of custom bindings (https://github.com/osmosis-labs/osmosis/pull/1484#issuecomment-1176960176).

If this is of interest, I can provide a PR.

mattverse:
Resolves #904

We have a history of disabling stargate queries, as some of the queries were non-deterministic to run on the state machine. (cref: https://github.com/CosmWasm/wasmd/pull/812).

This PR introduces re-enabling stargate queries whilst ensuring determinism using proto marshalling, json marshalling and whitelists.

Note that this PR does not simply revert #812, this PR includes converting the state machine response from a proto response to a json marsahlled response so that clients can use it directly by json unmarshalling. (This is also why we now feed codec as a parameter in StargateQuerier in this PR)


x/wasm/keeper/query_plugins_test.go
```diff
@@ -528,3 +539,192 @@ func (m bankKeeperMock) GetAllBalances(ctx sdk.Context, addr sdk.AccAddress) sdk
...
+
+			wasmApp.AppCodec().MustUnmarshalJSON(jsonMarshalledOriginalBz, protoResponse)
+
+			newVersionBz, err := hex.DecodeString(tc.updatedResponse)
+			require.NoError(t, err)
+			jsonMarshalledUpdatedBz, err := keeper.ConvertProtoToJSONMarshal(protoResponse, newVersionBz, wasmApp.AppCodec())
+			require.NoError(t, err)
+
+			// json marshalled bytes should be the same since we use the same proto sturct for unmarshalling
+			require.Equal(t, jsonMarshalledOriginalBz, jsonMarshalledUpdatedBz)
```
alpe, line 722: better output
```suggestion
			require.JSONEq(t, string(jsonMarshalledOriginalBz), string(jsonMarshalledUpdatedBz))
```

x/wasm/keeper/query_plugins_test.go
```diff
@@ -528,3 +539,192 @@ func (m bankKeeperMock) GetAllBalances(ctx sdk.Context, addr sdk.AccAddress) sdk
...
+			if tc.expectedError {
+				require.Error(t, err)
+				return
+			}
+			require.NoError(t, err)
+
+			// check response by json marshalling proto response into json response manually
+			jsonMarshalExpectedResponse, err := wasmApp.AppCodec().MarshalJSON(tc.expectedProtoResponse)
+			require.NoError(t, err)
+			require.Equal(t, jsonMarshalledResponse, jsonMarshalExpectedResponse)
```
alpe, line 591: you are unmarshalling an object twice into json and compare the result. It may use the same code under the hood. When you have the expected output defined as json, it would make the test more readable and ensures the result is what you want.

For better diffs:
```suggestion
			require.JSONEq(t, string(jsonMarshalledResponse), string(jsonMarshalExpectedResponse))
```

x/wasm/keeper/query_plugins_test.go
```diff
@@ -528,3 +539,192 @@ func (m bankKeeperMock) GetAllBalances(ctx sdk.Context, addr sdk.AccAddress) sdk
...
+	for _, tc := range testCases {
+		t.Run(fmt.Sprintf("Case %s", tc.name), func(t *testing.T) {
+			// set up app for testing
+			wasmApp := app.SetupWithEmptyStore(t)
+			// wasmApp := app.SetupWithEmptyStore(t)
+
+			// set up whitelist manually for testing
+			protoResponse, ok := tc.responseProtoStruct.(proto.Message)
+			require.True(t, ok)
+			keeper.StargateWhitelist.Store(tc.queryPath, protoResponse)
```
alpe, line 700: I don't think that the whitelist adds much value to this test. You may want to call `StargateQuerier` to cover it

x/wasm/keeper/stargate_bindings.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	"sync"
+)
+
+// StargateWhitelist keeps whitelist and its deterministic
+// response binding for stargate queries.
+//
+// The query can be multi-thread, so we have to use
+// thread safe sync.Map.
+var StargateWhitelist sync.Map
```
alpe, line 12: Let's make this an AcceptList to be sensible with terms

x/wasm/keeper/stargate_bindings.go
```diff
@@ -0,0 +1,18 @@
...
+
+// Define whitelists here as maps using 'StargateWhitelist'
+// e.x) StargateWhitelist.Store("/cosmos.auth.v1beta1.Query/Account", &authtypes.QueryAccountResponse{})
+func init() {
```
alpe, line 16: Any suggestions what should go in here? I would prefer to not use the init() method but set/ call this from app.go.

alpe (commented):
Thanks a lot for your contribution and driving this forward!
I still feel not very confident with enabling Stargate queries due to their history. I have commented a bit on the code but would ask @webmaster128 and @ethanfrey to comment on the status with the SDK and steps needed to enabled them again.

x/wasm/keeper/stargate_bindings.go
jhernandezb: I'll also prefer to be something to be set by app.go or similar, since this will go unnoticed for projects that may not want to enable some of the items in this potential list.

x/wasm/keeper/query_plugins_test.go
mattverse: ah nice!

x/wasm/keeper/stargate_bindings.go
mattverse: Agree with you on moving towards being more sensible with terms here, while I still feel like having `Stargate` as a part of the variable name is an important part. What do you think about `StargateAcceptList`?

x/wasm/keeper/query_plugins.go
```diff
@@ -517,6 +545,30 @@ func ConvertSdkCoinToWasmCoin(coin sdk.Coin) wasmvmtypes.Coin {
 	}
 }
 
+// ConvertProtoToJSONMarshal  unmarshals the given bytes into a proto message and then marshals it to json.
+// This is done so that clients calling stargate queries do not need to define their own proto unmarshalers,
+// being able to use response directly by json marshalling, which is supported in cosmwasm.
+func ConvertProtoToJSONMarshal(protoResponse interface{}, bz []byte, codec codec.Codec) ([]byte, error) {
+	// all values are proto message
+	message, ok := protoResponse.(proto.Message)
```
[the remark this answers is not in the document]
mattverse: Good catch!

x/wasm/keeper/query_plugins_test.go
mattverse: I'm thinking of a separate case for `StargateQuerier`, this is just a test case to ensure we have the same marshalled value and response upon proto struct change

x/wasm/keeper/stargate_bindings.go
mattverse: Should we simply remove this init method here? And then projects can simply call AcceptList.Store() or AcceptList.Load() for setter and getter for the whitelists

x/wasm/keeper/query_plugins_test.go
mattverse: Or, if you think this testing does not have much value in testing, we can simply replace this with a test for StargateQuerier.

https://github.com/osmosis-labs/osmosis/pull/2353/files#diff-b1c91ccd0545c457c9470ecc7dd0fb906193e8986db10316c9f14a1d7d3d4f26R43-R164 This is a test I'm trying to get on this PR(we can also do it on a separate PR, but having hard time in doing so due to different testing settings)

ethanfrey:
@alpe 

I feel strongly against enabling any StargateQueries by default in x/wasm or wasmd.

I have talked with the sdk team and they will be working to provide stability guarantees in the future, so we will be able to safely whitelist some queries in the future.

Some chains like Osmosis want to whitelist both their custom queries (from modules they control, can guarantee) and some standard sdk modules they feel comfortable handling.

Since there is a strong push for this, I think providing a standard way for a chain to safely enable a subset of queries makes sense. As long as the implementation seems solid, and the default behavior is same as now (no StargateQueries) unless the chain explicitly enables them, I think this is safe and happy to roll this out.

Each chain can decide how much risk they want to take and knowingly take it.

ethanfrey:
Once the above code comments are addressed, I would be happy to see this merged (and know a few chains are also excited by that)

x/wasm/keeper/query_plugins_test.go
pinosu: @mattverse I think it would be nice to have a test where you setup AcceptList and then you call StargateQuerier to verify that whithelisted queries are executed and the others return an error.

x/wasm/keeper/stargate_bindings.go
pinosu: @mattverse I agree you can simply remove the init method

alpe:
I have opened a new PR based on this to complete the feature integration. @mattverse please 👀 https://github.com/CosmWasm/wasmd/pull/1069

PR #974, opened 2022-09-02
Introduce new AccessType to allow a set of addresses

### LINKED ISSUE #945: Allow AccessConfig to use a list of addresses instead of just a single address
Currently the [AccessConfig](https://github.com/CosmWasm/wasmd/blob/main/proto/cosmwasm/wasm/v1/types.proto#L37) allows for a single address to be given for use with `ACCESS_TYPE_ONLY_ADDRESS`.  However, often times we might want to have a set of addresses be given a certain access control.   For example, in a chain with permissioned code upload, we may want to give a list of trusted actors the ability to upload code without going through the governance process.

We should make a new AccessType called `ACCESS_TYPE_ONLY_ADDRESSES` that allows for a list of addresses instead of just a single one.

We could achieve something like this with authz, but a proper whitelist is cleaner and can be fully controlled by governance rather than the original grantee.

alpe:
Resolves #945

Introduces a new Access type to allow a set of addresses to accept.

This PR includes a small but breaking change to the CLI. The separator char between codeID and permission is `:` now:
Before:
```sh
wasmd tx gov submit-proposal update-instantiate-config 1,nobody 2,everybody 3,wasm1l2rsakp388kuv9k8qzq6lrm9taddae7fpx59wm
```
Now:
```sh
wasmd tx gov submit-proposal update-instantiate-config 1:nobody 2:everybody 3:wasm1l2rsakp388kuv9k8qzq6lrm9taddae7fpx59wm,wasm1vx8knpllrj7n963p9ttd80w47kpacrhuts497x`
```

- [x] keep or migrate `AccessConfig.Address` into `AccessConfig.Addresses` see [proto](https://github.com/CosmWasm/wasmd/pull/974/files#diff-afea929694a2920f57dcfab61a08130539f5402a9f0c46e9c6265f8122e31b6fR48)? - keep and remove in a future version


proto/cosmwasm/wasm/v1/types.proto
```diff
@@ -19,12 +19,16 @@ enum AccessType {
   // AccessTypeNobody forbidden
   ACCESS_TYPE_NOBODY = 1
       [ (gogoproto.enumvalue_customname) = "AccessTypeNobody" ];
-  // AccessTypeOnlyAddress restricted to an address
+  // AccessTypeOnlyAddress restricted to a single address
+  // Deprecated: use AccessTypeAnyOfAddresses instead
```
ethanfrey, line 23: Why deprecate?
I think it is fine to use the simpler version for the cases where it works

proto/cosmwasm/wasm/v1/types.proto
```diff
@@ -19,12 +19,16 @@ enum AccessType {
...
   ACCESS_TYPE_ONLY_ADDRESS = 2
       [ (gogoproto.enumvalue_customname) = "AccessTypeOnlyAddress" ];
   // AccessTypeEverybody unrestricted
   ACCESS_TYPE_EVERYBODY = 3
       [ (gogoproto.enumvalue_customname) = "AccessTypeEverybody" ];
+  // AccessTypeAnyOfAddresses allow any of the addresses
```
ethanfrey, line 29: Looks good

proto/cosmwasm/wasm/v1/types.proto
```diff
@@ -37,7 +41,11 @@ message AccessTypeParam {
 message AccessConfig {
   option (gogoproto.goproto_stringer) = true;
   AccessType permission = 1 [ (gogoproto.moretags) = "yaml:\"permission\"" ];
+
+  // Address
+  // Deprecated: replaced by addresses
```
ethanfrey, line 46: The decision to deprecate is to not maintain two fields here?
Maybe worth it 🤷

x/wasm/types/params.go
```diff
@@ -171,6 +208,13 @@ func (a AccessConfig) Allowed(actor sdk.AccAddress) bool {
 		return true
 	case AccessTypeOnlyAddress:
 		return a.Address == actor.String()
+	case AccessTypeAnyOfAddresses:
+		for _, v := range a.Addresses {
```
ethanfrey, line 212: Looks good logic.
"Not covered by tests" comments means this is not used?
I guess this is why it is draft

x/wasm/types/params_test.go
```diff
@@ -89,6 +110,27 @@ func TestValidateParams(t *testing.T) {
 			},
 			expErr: true,
 		},
+		"reject invalid address in any of addresses": {
```
ethanfrey, line 113: Can you add test case for AnyOf with empty list? That should also fail (it makes no sense, likely typo)

x/wasm/types/types.go
```diff
@@ -356,8 +356,32 @@ func (a AccessConfig) IsSubset(superSet AccessConfig) bool {
 	switch superSet.Permission {
 	case AccessTypeOnlyAddress:
 		// An exact match or nobody
-		return a.Permission == AccessTypeNobody || (a.Permission == AccessTypeOnlyAddress && a.Address == superSet.Address)
+		return a.Permission == AccessTypeNobody || (a.Permission == AccessTypeOnlyAddress && a.Address == superSet.Address) ||
+			(a.Permission == AccessTypeAnyOfAddresses && isSubset([]string{superSet.Address}, a.Addresses))
+	case AccessTypeAnyOfAddresses:
+		// An exact match or nobody
+		return a.Permission == AccessTypeNobody || (a.Permission == AccessTypeOnlyAddress && isSubset(superSet.Addresses, []string{a.Address})) ||
```
ethanfrey, line 363: Nice with the downgrade from AnyOf to a Only

x/wasm/types/types.go
```diff
@@ -356,8 +356,32 @@ func (a AccessConfig) IsSubset(superSet AccessConfig) bool {
 	switch superSet.Permission {
 	case AccessTypeOnlyAddress:
 		// An exact match or nobody
-		return a.Permission == AccessTypeNobody || (a.Permission == AccessTypeOnlyAddress && a.Address == superSet.Address)
+		return a.Permission == AccessTypeNobody || (a.Permission == AccessTypeOnlyAddress && a.Address == superSet.Address) ||
+			(a.Permission == AccessTypeAnyOfAddresses && isSubset([]string{superSet.Address}, a.Addresses))
```
ethanfrey, line 360: Interesting with the "upgrade" to an AnyOf with one item.

proto/cosmwasm/wasm/v1/types.proto
alpe: It feels wasted to have both fields and may be confusing when used in the API. Added https://github.com/CosmWasm/wasmd/issues/986 to follow up on the deprecation tag

jhernandezb (approved):
Looks good to me

alpe:
Thanks for the feedback!

PR #976, opened 2022-09-04
Community proposal V0.29.0

faddat:
## This is abnormal, please don't directly merge

This PR is a proposed v0.29.0 release.

Specifically, this release would:

* bump ibc-go to v4.0.0
* Use the latest 45 series sdk
* tendermint v0.34.21

In order to do this, we would need to create release branches.  Please, consider this PR as targeting a branch that doesn't yet exist.  It is a magical branch that begins from v0.28.0, and is called "release/v0.29.0". -- this does interrupt the plans for v0.29.0 a bit, we can call it v0.30.0 and it can be same.  I propose that v0.29.0 be a branch that doesn't change the contract features at all, but gets wasmd caught up with stuff like:

* iavl fast cache
* sdk v0.45.8
* ibc-go v4.0.0

And, there's a reason for this, too.  This is basically what Juno needs.  If it wasn't this, then we would need more like:

* sdk v0.46.1
* tendermint v0.34.21
* ibc-go v5.0.0

The latter grouping is in fact my preference but we haven't made a final release for ibc-go v5.0.0 yet.


alpe:
I have renamed the PR to not create confusion with our current [v0.29](https://github.com/CosmWasm/wasmd/milestone/21) milestone and branching model.
Thanks for highlighting the demands for Juno chain. I will discuss this.

ethanfrey:
We will make a v0.29.0 release off of main later this week (within 3 days).
That will include ibc-go v3.2, we already bumped cosmos sdk and tendermint.

v0.29.0 will also include many other features and bug fixes.

PR #979, opened 2022-09-05
Add unpin code flag to store code proposal

### LINKED ISSUE #972: Make contract pinning an optional field in StoreCode proposals
Currently contracts are pinned automatically for every store code proposal, but it does not always makes sense and having to do a second proposal to unpin them just requires more governance work.

I propose to add an optional field to StoreCode proposals that will act as a flag to Pin or not when uploading them.

jhernandezb:
This should close #972 if you think it makes sense to have it as part of  a governance proposal.


x/wasm/client/cli/gov_tx.go
```diff
@@ -80,6 +86,7 @@ func ProposalStoreCodeCmd() *cobra.Command {
 	cmd.Flags().String(flagInstantiateByEverybody, "", "Everybody can instantiate a contract from the code, optional")
 	cmd.Flags().String(flagInstantiateNobody, "", "Nobody except the governance process can instantiate a contract from the code, optional")
 	cmd.Flags().String(flagInstantiateByAddress, "", "Only this address can instantiate a contract instance from the code, optional")
+	cmd.Flags().Bool(flagPinCode, false, "Pin code on upload, optional")
```
ethanfrey, line 89: I agree with this PR in general.
I would change the default to be pinning, and require explicitly unpinning.

You can safely pin a few hundred contracts, and I assume most ones on permissioned chains would be high-value.

ethanfrey: Best way may be to rename to "unpinContract" or so, so default (no protobuf field, no cli flag) would be the same as the current behavior.

ethanfrey (commented):
Looks good. Only real comment is on the default.

x/wasm/client/cli/gov_tx.go
jhernandezb: @ethanfrey thanks for the feedback, so just to get things clear before updating the PR.

You mean only change the flag ?
```
cmd.Flags().Bool(flagUnpinContract, false, "Unpin contract as part of the proposal, optional")
``` 

But send `PinCode: true` as default unless the flag is enabled ?

alpe: The issue is with in flight gov proposals. The default of the new (proto) field would be false and,changes the behaviour. Or we need to provide a store migration to set the field too true. 
When you rename the field to something like `unpinned` then the unset field would default to false and the code can run as before with the same outcome.

x/wasm/keeper/proposal_integration_test.go
```diff
@@ -24,40 +24,58 @@ import (
...
+			require.NoError(t, err)
 
-	storedCode, err := wasmKeeper.GetByteCode(ctx, 1)
-	require.NoError(t, err)
-	assert.Equal(t, wasmCode, storedCode)
+			// then
+			cInfo := wasmKeeper.GetCodeInfo(ctx, 1)
+			require.NotNil(t, cInfo)
+			assert.Equal(t, myActorAddress, cInfo.Creator)
+			assert.Equal(t, spec.pinCode, wasmKeeper.IsPinnedCode(ctx, 1))
```
alpe, line 72: 👍 good testing

proto/cosmwasm/wasm/v1/proposal.proto
```diff
@@ -24,6 +24,8 @@ message StoreCodeProposal {
   reserved 5, 6;
   // InstantiatePermission to apply on contract creation, optional
   AccessConfig instantiate_permission = 7;
+  // Pin code on upload, optional
+  bool pin_code = 8;
```
alpe, line 28: Please rename to `unpinned` or something similar that indicates code is going to be pinned when not set too `true`. See comment on the cli flag

alpe (changes requested):
Very nice work! 🏅 
Please change the proto field name and cli/ gov handler accordingly to complete this.

x/wasm/keeper/proposal_handler.go
```diff
@@ -68,7 +68,10 @@ func handleStoreCodeProposal(ctx sdk.Context, k types.ContractOpsKeeper, p types
 	if err != nil {
 		return err
 	}
-	return k.PinCode(ctx, codeID)
+	if p.PinCode {
```
alpe, line 71: This would become `if !p.Unpinned` (or similar name)

ethanfrey:
If you can get this fixed up by Wednesday, happy to include this in the v0.29 release.
We want to cut by end of the week.
Otherwise, it will come in v0.30.0

x/wasm/client/cli/gov_tx.go
ethanfrey: What Alex said, change the field and change the flag.
Default behaviour remains the same, unless they explicitly add it

proto/cosmwasm/wasm/v1/proposal.proto
jhernandezb: changed to `unpin_code`

x/wasm/keeper/proposal_handler.go
jhernandezb: switched to early return instead
```
// if code should not be pinned return earlier
	if p.UnpinCode {
		return nil
	}
```
I think it has better readability, double negation logic confuses me a bit personally 😅   but I can change it if there is a preference.

x/wasm/client/rest/gov.go
```diff
@@ -28,8 +28,8 @@ type StoreCodeProposalJSONReq struct {
 	// InstantiatePermission to apply on contract creation, optional
 	InstantiatePermission *types.AccessConfig `json:"instantiate_permission" yaml:"instantiate_permission"`
 
-	// PinCode indicates if the code should be pinned as part of the proposal.
-	PinCode bool `json:"pin_code" yaml:"pin_code"`
+	// UnpinCode indicates if the code should not be pinned as part of the proposal.
+	UnpinCode bool `json:"pin_code" yaml:"pin_code"`
```
ethanfrey: json/yaml comments need to be updated here

x/wasm/client/cli/gov_tx.go
jhernandezb: Changed

ethanfrey (commented):
Minor issue in the rest client.
Otherwise, this looks great and ready to merge.

Please adjust those fields and mention me

x/wasm/client/rest/gov.go
jhernandezb: just noticed this and pushed the change

x/wasm/client/rest/gov.go
```diff
@@ -29,7 +29,7 @@ type StoreCodeProposalJSONReq struct {
 	InstantiatePermission *types.AccessConfig `json:"instantiate_permission" yaml:"instantiate_permission"`
 
 	// UnpinCode indicates if the code should not be pinned as part of the proposal.
-	UnpinCode bool `json:"pin_code" yaml:"pin_code"`
+	UnpinCode bool `json:"unpin_code" yaml:"unpin_code"`
```
ethanfrey, line 32: Haha. You fixed this while I was reviewing. Great job!

ethanfrey (approved):
Looking good

alpe (approved):
🏇 Awesome

Issue #980, opened 2022-09-05
Upgrade to wasmvm 1.1.1


ethanfrey:
- [x] https://github.com/CosmWasm/wasmvm/pull/347 is merged
- [x] wasmvm 1.1.0 is tagged.
- [x] Update wasmvm (go.mod, Dockerfile)
- [x] Features have been renamed to capabilities. Rename all "supported features" variables to "available capabilities".
- [x] Finalize and merge https://github.com/CosmWasm/wasmd/pull/903
- [x] Add `cosmwasm_1_1` to available capabilities
- [x] Add integration test to cover new supply query
- [x] Doc new `cosmwasm_1_1`

This ticket closes https://github.com/CosmWasm/wasmd/issues/932

webmaster128:
Here we go: https://github.com/CosmWasm/wasmvm/releases/tag/v1.1.0 is ready for integration.

webmaster128:
What do you think needs to be done for "Doc new cosmwasm_1_1"? Do you have a list of capabilities somewhere already? For users of wasmd I don't think this is terribly important since they should just use or append to the wasmd defaults and wasmd implements cosmwasm_1_1 now.

How does appending to defaults work by the way? I hope users do not need to list our defaults when adding their capabilities.

alpe:
I have added some notes to the [changelog](https://github.com/CosmWasm/wasmd/pull/1024) but it would be good to have a better place for them in the future. IMHO it makes sense to list and describe them im cosmwasm where they are defined. In this case, I could link to them from the source code and maybe some integration doc

PR #988, opened 2022-09-07
Better wasmvm 1.1 integration test

alpe:
Relates to #903 and #980 

Add an integration test to call the new supply query via reflect contract
Includes:
- reflect contract version upgrade to v1.1.0
- Add `cosmwasm_1_1` capabiltiy


app/app.go
```diff
@@ -507,7 +507,7 @@ func NewWasmApp(
 
 	// The last arguments can contain custom message handlers, and custom query handlers,
 	// if we want to allow any custom callbacks
-	supportedFeatures := "iterator,staking,stargate"
+	supportedFeatures := "iterator,staking,stargate,cosmwasm_1_1"
```
jhernandezb, line 510: Should this be documented for the release? 

Also we started to enforce version check for Stargaze to prevent validators from using a different wasm library and avoid non-determinism  from nodes that don't upgrade correctly.

https://github.com/public-awesome/stargaze/blob/main/cmd/starsd/cmd/start.go

Not really sure if this is something that should also be part of wasmd.

alpe: >  using a different wasm library and avoid non-determinism from nodes that don't upgrade correctly.

Not sure if I got that correct. Do you have validators that overwrite the wasmvm dependency? That would be hard to cover as they may overwrite other parts as well. Feel free to open an issue to discuss this further.

jhernandezb: > > using a different wasm library and avoid non-determinism from nodes that don't upgrade correctly.
> 
> Not sure if I got that correct. Do you have validators that overwrite the wasmvm dependency? That would be hard to cover as they may overwrite other parts as well. Feel free to open an issue to discuss this further.

Not overwrite but they often compile on one machine upload the binary to the remote/secured host and run it so it may be running with an old libwasm. It has caused consensus issues before, at least from our side we prevent nodes from running if the lib is not the right version as you can see the code I pointed. But yes it is a different discussion.

x/wasm/keeper/reflect_test.go
```diff
@@ -298,6 +298,57 @@ func TestReflectStargateQuery(t *testing.T) {
 	assert.Equal(t, simpleBalance.Amount[0].Denom, expectedBalance[0].Denom)
 }
 
+func TestReflectTotalSupplyQuery(t *testing.T) {
```
pinosu, line 301: Nice test! Variable _currentStateSupply_ could be renamed as suggested by Tom.

pinosu (approved):
LGTM 👍

app/app.go
alpe: Thanks for adding more context. I think this is a valid use case. I prefer static compilation but this is an issue indeed

PR #990, opened 2022-09-08
fix bug of contract history

yann-sjtu:
position was stored in key but not value


alpe:
Good 👁️ ! Thanks for the fix! 💯 🏅

PR #993, opened 2022-09-08
Rename features to availableCapabilities

alpe:
See #980 

To use the same term "capabilities" as there. Not to be mixed with cosmos capabilities module

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -507,7 +507,7 @@ func NewWasmApp(
 
 	// The last arguments can contain custom message handlers, and custom query handlers,
 	// if we want to allow any custom callbacks
-	supportedFeatures := "iterator,staking,stargate,cosmwasm_1_1"
+	availableCapabilities := "iterator,staking,stargate,cosmwasm_1_1"
 	app.WasmKeeper = wasm.NewKeeper(
 		appCodec,
 		keys[wasm.StoreKey],
@@ -524,7 +524,7 @@ func NewWasmApp(
 		app.GRPCQueryRouter(),
 		wasmDir,
 		wasmConfig,
-		supportedFeatures,
+		availableCapabilities,
 		wasmOpts...,
 	)
 
--- a/x/wasm/keeper/keeper.go
+++ b/x/wasm/keeper/keeper.go
@@ -103,10 +103,10 @@ func NewKeeper(
 	queryRouter GRPCQueryRouter,
 	homeDir string,
 	wasmConfig types.WasmConfig,
-	supportedFeatures string,
+	availableCapabilities string,
 	opts ...Option,
 ) Keeper {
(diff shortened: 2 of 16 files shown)


x/wasm/keeper/keeper.go
```diff
@@ -196,16 +196,16 @@ func (k Keeper) create(ctx sdk.Context, creator sdk.AccAddress, wasmCode []byte,
...
 	k.storeCodeInfo(ctx, codeID, codeInfo)
 
 	evt := sdk.NewEvent(
 		types.EventTypeStoreCode,
 		sdk.NewAttribute(types.AttributeKeyCodeID, strconv.FormatUint(codeID, 10)),
 	)
-	for _, f := range strings.Split(report.RequiredFeatures, ",") {
-		evt.AppendAttributes(sdk.NewAttribute(types.AttributeKeyFeature, strings.TrimSpace(f)))
+	for _, f := range strings.Split(report.RequiredCapabilities, ",") {
+		evt.AppendAttributes(sdk.NewAttribute(types.AttributeKeyCapability, strings.TrimSpace(f)))
```
webmaster128, line 208: Can you rename this to `AttributeKeyRequiredCapability` / `"required_capability"`? The contract that is stored does not have capabilities. It asks the host to provide those capabilities.

app/app.go
```diff
@@ -507,7 +507,7 @@ func NewWasmApp(
 
 	// The last arguments can contain custom message handlers, and custom query handlers,
 	// if we want to allow any custom callbacks
-	supportedFeatures := "iterator,staking,stargate"
+	availableCapabilities := "iterator,staking,stargate"
```
webmaster128, line 510:
```suggestion
	availableCapabilities := "iterator,staking,stargate,cosmwasm_1_1"
```

x/wasm/keeper/keeper_test.go
```diff
@@ -40,15 +40,15 @@ func init() {
 
 var hackatomWasm []byte
 
-const SupportedFeatures = "iterator,staking,stargate"
+const AvailableCapabilities = "iterator,staking,stargate"
```
webmaster128, line 43:
```suggestion
const AvailableCapabilities = "iterator,staking,stargate,cosmwasm_1_1"
```

webmaster128 (approved):
LGTM.

I think `cosmwasm_1_1` should be listed as available capability by default now. I suggested adding it to the lists here but I'm not sure of those are all the relevant places. Feel free to ignore if you want to do that separately.

alpe:
> I think cosmwasm_1_1 should be listed as available capability by default now. 

This makes perfect sense! I did this in #988 already which was started and should be merged before this PR. But it would be cleaner here.

PR #996, opened 2022-09-09
Make contract addresses predictable

### LINKED ISSUE #942: Make contract addresses predictable ("deterministic")
This is a request from @sunnya97 from quite some time ago, which I am just getting around to writing down properly.

**Request**

The desired state is that a contract creator can pre-compute the next 5 (or 500) contract addresses it will create. This will allow them to publish said addresses and send tokens (or vesting tokens) to such addresses before they exist. The original use case was some sort of exchange creating many new addresses for users, then deploying later. However, I have come up with many more use cases:

1. Expose a query both to users (via gRPC) and contracts (via QueryRequest) to list the next address they will create. For contracts, it will remove the most common case of reply_on, to simply get the address of the contract we just instantiated, making this a simpler one-step process. It will also make it easier to produce cycles (contract A refers to contract B, contract B refers to contract A - like a cw20 contract and the staking contract)
2. Allow sending "vesting tokens" to a (multsig/DAO) contract. Currently, when we instantiate a contract, we create a BaseAccount. Once created, we can no longer send vesting tokens to it. Before it was created, we don't know for certainty what address it will have. This will allow one to receive vesting tokens on a pre-computed address, then deploy a new multisig/dao contract there

**Current State**

The current state is that we create a "sub-module address" for each contract, but the inputs we use are the [codeID and a global counter](https://github.com/CosmWasm/wasmd/blob/f14c46988aa26d5bb393a200cfb77e5ccc380e68/x/wasm/keeper/keeper.go#L947-L959) of number of instances created. If someone else deploys a contract, they could claim my address or make the address unreachable. This doesn't allow the creator to have control of the address it will receive.

Ethereum has a `CREATE` opcode that does a "sender + nonce hash" to create the new address. And [CREATE2](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1014.md) which does a more complex hash, including initialisation data.

**Proposed Changes**

In order to make these maximally predictable, I would base on the `CREATE` syntax. However, we cannot use the typical nonce, as (a) an external account can instantiate multiple contracts in one transaction / same nonce and (b) in CosmWasm contracts don't have nonces, and we need to count something there.

I would add a new counter *per-address* that maintains how many contracts this sender has created. If a contract creates 3 sub-contracts in the same transaction, each would get another counter value. I would then use (`0xcafe || senderAddress (20-32 bytes) || counter (8 bytes)`) as input into the standard module account hash function. 

The use of that prefix ensures the has space doesn't overlap with any previously created addresses from the `x/wasm` module. And use of module account derivation means that it doesn't overlap with any addresses created by other modules. Existing contracts already have an address and this will just affect the generation of new addresses without breaking any backwards compatibility. 

Once this is deployed, we can immediately add support for clients to query the future contract address. With the next CosmWasm upgrade (v1.2?) we can add a new query to pre-compute the address that will be used upon the next instantiate call.

alpe:
Resolves #942 

This PR introduces the new predictable contract address model. 
It is based on hashing `(len(checksum) | checksum | len(sender_address) | sender_address | len(label) | label)`. Please see the issue linked above for more details and discussion

In order to deal with existing accounts in the wasm address space, the following algorithm is applied.

```
If an account exists:
    assert sequence == 0 && pubkey == nil
    if account type is in accept list:
        use it
    if account type is in prune list:
       overwrite with base account type and burn balance
    else:
        return error
else:
    crease base account
```
See [keeper.go](https://github.com/CosmWasm/wasmd/pull/996/files#diff-4e16010ab332946722b789b22962fb5a0814d529cd8de443b925d20ccd13705fR316-R340) for the concrete implementation.
We had long internal discussions which account types should go into the accept and prune lists and came up with some defaults that support many use cases and are still safe to use. But as we can not know all use cases and preferences, there are new [Options](https://github.com/CosmWasm/wasmd/pull/996/files#diff-9b6fce18b5d6e5666a46c228799ae067751c96343a322cdc4dbd4e9dbf18f5a2) to customize these lists and the balance handling.

### CLI example to calculate an new contract address
```sh
 wasmd query wasm build-address [code-hash] [creator-address] [label] [flags]
```


x/wasm/keeper/keeper.go
```diff
@@ -278,10 +298,53 @@ func (k Keeper) instantiate(ctx sdk.Context, codeID uint64, creator, admin sdk.A
 	var codeInfo types.CodeInfo
 	k.cdc.MustUnmarshal(bz, &codeInfo)
 
-	if !authZ.CanInstantiateContract(codeInfo.InstantiateConfig, creator) {
+	if !authPolicy.CanInstantiateContract(codeInfo.InstantiateConfig, creator) {
 		return nil, nil, sdkerrors.Wrap(sdkerrors.ErrUnauthorized, "can not instantiate")
 	}
 
+	contractAddress := BuildContractAddress(codeInfo.CodeHash, creator, label)
+	if k.HasContractInfo(ctx, contractAddress) {
```
ethanfrey, line 306: Important to have this assertion independent of the account type

x/wasm/keeper/keeper.go
```diff
@@ -278,10 +298,53 @@ func (k Keeper) instantiate(ctx sdk.Context, codeID uint64, creator, admin sdk.A
...
+		return nil, nil, types.ErrDuplicate.Wrap("instance with this code id, sender and label exists: try a different label")
+	}
+
+	// check account
+	// every cosmos module can define custom account types when needed. The cosmos-sdk comes with extension points
+	// to support this and a set of base and vesting account types that we integrated in our default lists.
+	// But not all account types of other modules are known or may make sense for contracts, therefore we kept this
+	// decision logic also very flexible and extendable. We provide new options to overwrite the default settings via WithAcceptedAccountTypesOnContractInstantiation and
+	// WithPruneAccountTypesOnContractInstantiation as constructor arguments
+	existingAcct := k.accountKeeper.GetAccount(ctx, contractAddress)
```
ethanfrey, line 316: Nice clean implementation, helpful comment :+1:

x/wasm/keeper/keeper.go
```diff
@@ -955,18 +1018,19 @@ func (k Keeper) consumeRuntimeGas(ctx sdk.Context, gas uint64) {
...
-func BuildContractAddress(codeID, instanceID uint64) sdk.AccAddress {
-	contractID := make([]byte, 16)
-	binary.BigEndian.PutUint64(contractID[:8], codeID)
-	binary.BigEndian.PutUint64(contractID[8:], instanceID)
-	return address.Module(types.ModuleName, contractID)[:types.ContractAddrLen]
+// BuildContractAddress generates a contract address for the wasm module with len = types.ContractAddrLen using the
+// Cosmos SDK address.Module function.
+// Internally a key is built containing (len(checksum) | checksum | len(sender_address) | sender_address | len(label) | label).
+// All method parameter values must be valid and not be empty or nil.
+func BuildContractAddress(checksum []byte, creator sdk.AccAddress, label string) sdk.AccAddress {
```
ethanfrey, line 1025: Nice, well defined algorithm.

There was some discussion on including the `initMsg` data (or a hash of it) as well, to be compatible with CREATE2 from Ethereum. I don't remember the decision regarding this. The other data, and using the label string as user defined nonce is great. Including `initMsg` may make it harder for some of the use cases (like an exchange pre-generating many receive addresses) that were requested in the original issue. 

I am undecided here, and it probably is better without initMsg, just want to make this decision explicit.

x/wasm/keeper/keeper.go
```diff
@@ -278,10 +298,53 @@ func (k Keeper) instantiate(ctx sdk.Context, codeID uint64, creator, admin sdk.A
...
+		if _, accept := k.acceptedAccountTypes[reflect.TypeOf(existingAcct)]; accept {
+			// keep account and balance as it is
+			k.Logger(ctx).Info("instantiate contract with existing account", "address", contractAddress.String())
+		} else if _, clear := k.pruneAccountTypes[reflect.TypeOf(existingAcct)]; clear {
+			k.Logger(ctx).Info("pruning existing account for contract instantiation", "address", contractAddress.String())
+			// consider an account in the wasmd namespace spam and overwrite it.
+			contractAccount := k.accountKeeper.NewAccountWithAddress(ctx, contractAddress)
+			k.accountKeeper.SetAccount(ctx, contractAccount)
+			// also handle balance to not open cases where these accounts are abused and become liquid
+			if err := k.coinPruner.PruneBalances(ctx, contractAddress); err != nil {
```
ethanfrey, line 330: Nice. No tricks by "auto-unvesting" those tokens when we replace the vesting account with base account

ethanfrey (approved):
Great stuff!

alpe:
Thanks for the feedback. Merging this now to set a RC tag today for further tests. Please provide more feedback. This is not set in stone.

PR #997, opened 2022-09-09
Add ctx and accountKeeper to Custom encoder

LCyson:


x/wasm/keeper/handler_plugin_encoders.go
```diff
@@ -31,25 +31,28 @@ type (
 
 type MessageEncoders struct {
 	Bank         func(sender sdk.AccAddress, msg *wasmvmtypes.BankMsg) ([]sdk.Msg, error)
-	Custom       func(sender sdk.AccAddress, msg json.RawMessage) ([]sdk.Msg, error)
+	Custom       func(ctx sdk.Context, accountKeeper types.AccountKeeper, sender sdk.AccAddress, msg json.RawMessage) ([]sdk.Msg, error)
```
alpe, line 34: `Custom` is just function with minimal parameters. Instead of adding the account keeper here, you would set a custom encoder method:
```go
func MyCustomEncoder(accountKeeper types.AccountKeeper)func(ctx sdk.Context, accountKeeper types.AccountKeeper, sender sdk.AccAddress, msg json.RawMessage) ([]sdk.Msg, error){
	return func(ctx sdk.Context, accountKeeper types.AccountKeeper, sender sdk.AccAddress, msg json.RawMessage) ([]sdk.Msg, error) {
		acct: = accountKeeper.GetAccount(ctx, someAccount)
		// ...do impl
	}
}
```

alpe:
Instead of extending the parameters, you can use a custom function or a method from a struct that contain the account keeper. 
I will close this now but if you need help to set this up, please add an issue with your concrete example.

LCyson:
thank you @alpe, added a custom messenger instead. Pasted the PR here: https://github.com/sei-protocol/sei-chain/pull/245

Issue #896, opened 2022-07-03
Store contract info before vm.Instantiate


jhernandezb:
Currently contract information is stored after the instantiate method of the VM has been invoked.

Currently the issue with this is that if  contract A during instantiation is also instantiating contract B and B wants to know about its caller by querying contract info (to know  the code id and for potential whitelist check ) it will fail because the information is only stored after the `vm.Instantiate()` method has been executed.

Here is a diagram of the execution flow 
![flow](https://user-images.githubusercontent.com/3452489/189933175-69cd4d44-9cdc-405b-b721-0ec1f152b9dc.png)

 I don't see any dependency that prevents from  moving  the logic above the vm instantiation call

https://github.com/CosmWasm/wasmd/blob/14688c09855ee928a12bcb7cd102a53b78e3cbfb/x/wasm/keeper/keeper.go#L317-L321

alpe:
The contract response is handled in L329 after the metadata for contract A is stored. Therefore Contract B should be able to query Contract A's metadata. Contract A can not query it's own contract info though. 

> to know the code id and for potential whitelist check 

I have not heard about code id for authz, yet. Can you elaborate a bit more on this, please? 
Btw we have examples for forwarding an address or pre-authz that give strong guarantees and work nicely with code migration.
You can find examples for this in https://github.com/confio/tfi or https://github.com/confio/poe-contracts .

jhernandezb:
Confirmed is working

PR #1001, opened 2022-09-14
Add factory contract scenario

alpe:
See  #896
Implement flow diagram as unit test. 

@jhernandezb can you confirm the workflow is correct?


jhernandezb:
I can confirm is working

alpe:
Let's add this test for regression

pinosu (approved):
nice test! 💯

Issue #987, opened 2022-09-07
Remove genesis messages


alpe:
[Genesis messages](https://github.com/CosmWasm/wasmd/blob/main/x/wasm/client/cli/genesis_msg.go) are a complex functionality in the code and not easy to main. 
Not sure if it is used by anybody.
- [ ] Remove from CLI
- [ ] Remove from genesis protobuf

ethanfrey:
I would agree with this and only use wasmd genesis to load existing contracts and state (from a state dump) rather than add new contracts. But maybe there are some use cases that need this 🤷 

If we do keep it, I think this needs more testing (also manual testing in production by others)

alpe:
I used them for system tests in tgrade only

PR #1002, opened 2022-09-14
Fix genesis message tooling

alpe:
The addressed built for genesis message tooling calculate the wrong hash as the code is zipped

### DIFF
--- a/x/wasm/client/cli/genesis_msg.go
+++ b/x/wasm/client/cli/genesis_msg.go
@@ -8,6 +8,8 @@ import (
 	"errors"
 	"fmt"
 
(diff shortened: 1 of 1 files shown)


pinosu (changes requested):
Changes look good! 
Could you add some tests for the function **GetAllCodes**?

alpe:
> Could you add some tests for the function GetAllCodes?

Good call but I would love to remove the whole section instead, see #987 . This is a minimal fix as system tests in tgrade covered this: https://github.com/confio/tgrade/pull/451

pinosu (approved):
Makes sense. LGTM 👍

Issue #1000, opened 2022-09-13
Improvements to new address generation


alpe:
Feedback on the new address generation highlighted 2 cases that we could improve:
A) (existing) factory contracts that create new instances may run into address collusions with static labels
B) Overloading the label to be used as a salt may damage the label use case

See https://github.com/CosmWasm/wasmd/issues/942 for context

Some solutions to discuss would be:
A) adding the `Init Message` to the hash generation
B) adding an optional salt or nonce field to the MsgInstantiateContract. When sent from a contract or not set by the CLI we default to 
   B1) a constant or empty string
   B2) sender, label or init message again

  // tag: @the-frey , @ethanfrey , @webmaster128

ethanfrey:
I fully support (A) and had [raised that issue before](https://github.com/CosmWasm/wasmd/pull/996#discussion_r966971694). I never saw a reason explaining why it was not included, so adding it in before tagging v0.29.0 seems very obvious improvement.

For (B) I am open to adding a field, but it should be optional and would only be set by the client, as it would be api breaking on the contract interface. I [have a contract](https://github.com/confio/cw-ibc-demo/blob/main/contracts/simple-ica-host/src/contract.rs#L101-L112) which uses a different label with the same initMsg to create multiple sub-contracts. I have been recommending this pattern to get multiple addresses controlled by one contract, and cw1-whitelist doesn't really offer meaningful differences in initMsg in such cases. 

In order to maintain the control of the prehash space, I strongly suggest (B2) with label.

I think we need (A) + (B2) with label in order to avoid breaking existing factory contracts.

webmaster128:
Issue A is a good point. I'm not convinced Issue B is a problem. The label is an instance description and any outtomated system can put some nice human readable counter there.

Adding the init message does not help to prevent Issue A. It might make it less likely but you can have a factory that uses constant init messages. Trying to add a binary representation of the init message can lead to all sorts of problems as there is an infinite amount of valid binary representations of the init message. You need to have certainty to know the exact init message not only in object form but byte-by-byte when generating the address. That's easy to get wrong. Now, even if it works, you cannot pre-generated addresses and remain flexible with the init message later.

Adding a field to a protobuf message is a breaking change (as discussed here: https://github.com/cosmos/cosmos-sdk/blob/main/docs/architecture/adr-044-protobuf-updates-guidelines.md).

[...]

PR #1008, opened 2022-09-16
Add normalized init msg to address generation

alpe:
🚧 Early version - tests need to be fixed

See changes in [`keeper.go`](https://github.com/CosmWasm/wasmd/pull/1008/files#diff-4e16010ab332946722b789b22962fb5a0814d529cd8de443b925d20ccd13705f)

Resolves part of #1000


alpe:
As discussed, we don't do the normalization. Passing the init message though can be useful

PR #1012, opened 2022-09-20
Upgrade to wasmvm 1.1.1

### LINKED ISSUE #1010: Upgrade wasmvm to 1.1.1
The bug discovered in Juno testnet and described in https://github.com/CosmWasm/cosmwasm/issues/1426 should be fixed for other users who do a wasmd 0.28 -> 0.29 upgrade. Fix is simple. wasmvm 1.1.1 is coming this week.

webmaster128:
Closes #1010

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -161,6 +161,17 @@ jobs:
       - run:
           name: Build Docker artifact
           command: docker build --pull -t "cosmwasm/wasmd:${CIRCLE_SHA1}" .
+      - run:
+          name: Ensure libwasmvm version is correct
+          command: |
+            IN_DOCKER=$(docker run --rm "cosmwasm/wasmd:${CIRCLE_SHA1}" /usr/bin/wasmd query wasm libwasmvm-version)
+            echo "Runtime libwasmvm-version in docker: $IN_DOCKER"
+            IN_GOMOD=$(go list -m github.com/CosmWasm/wasmvm | cut -d" " -f2 | cut -d"v" -f2)
+            echo "wasmvm version in go.mod: $IN_GOMOD"
+            if [[ "$IN_DOCKER" != "$IN_GOMOD" ]]; then
+              echo "Mismatch of wasmvm versions detected"
+              exit 1
+            fi
       - when:
           condition:
             equal: [ main, << pipeline.git.branch >> ]
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -196,14 +196,9 @@ workflows:
   version: 2
   test-suite:
     jobs:
-      # this is now a slow process... let's only run on main and PR branches we care about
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -194,14 +196,15 @@ workflows:
   version: 2
   test-suite:
     jobs:
-      # this is now a slow process... let's only run on master
+      # this is now a slow process... let's only run on main and PR branches we care about
```
webmaster128, line 199: The `docker-image` job is not blocking the CI completion. So I vote for removing the filter here (only keep the filter on the publishing step). Is this okay for you, @alpe?
Bildschirmfoto 2022-09-20 um 10 15 47

webmaster128: okay, it's not even close
Bildschirmfoto 2022-09-20 um 10 26 43

webmaster128:
Bildschirmfoto 2022-09-20 um 13 15 13

Oh yeah, it works when I forget to update the Docker file :)

.circleci/config.yml
```diff
@@ -162,14 +162,27 @@ jobs:
...
-              docker tag "cosmwasm/wasmd:${CIRCLE_SHA1}" cosmwasm/wasmd:latest
-              docker login --password-stdin -u "$DOCKER_USER" \<<<"$DOCKER_PASS"
-              docker push cosmwasm/wasmd:latest
-              docker logout
+            IN_DOCKER=$(docker run --rm "cosmwasm/wasmd:${CIRCLE_SHA1}" /usr/bin/wasmd query wasm libwasmvm-version)
+            echo "Runtime libwasmvm-version in docker: $IN_DOCKER"
+            IN_GOMOD=$(go list -m github.com/CosmWasm/wasmvm | cut -d" " -f2 | cut -d"v" -f2)
+            echo "wasmvm version in go.mod: $IN_GOMOD"
+            if [[ "$IN_DOCKER" != "$IN_GOMOD" ]]; then
+              echo "Mismatch of wasmvm versions detected"
```
alpe, line 172: 💯 very nice check

.circleci/config.yml
```diff
@@ -162,14 +162,27 @@ jobs:
...
+              exit 1
             fi
+      - when:
+          condition:
+            equal: [ main, << pipeline.git.branch >> ]
```
alpe, line 177: This pushes to docker hub on main brach builds only, right?

alpe (approved):
Thanks for adding this dockerfile check, too!  💯

.circleci/config.yml
webmaster128: Right, that's the idea. Not 100 % tested but it is coming from the docs. At least in this PR the `when` step is skipped because the condition is not met. Let's see what is happening on main.

PR #1014, opened 2022-09-20
Implement improvements to new address generation

### LINKED ISSUE #1000: Improvements to new address generation

alpe:
Resolves #1000 
With this PR the sequence based (classic) address generation is added again. This PR got rather big as a lot of files are touched. But tests and code were restored from the original version. 

A first round of feedback from @webmaster128 was applied on an early version, before this branch was rebased. More is very welcome!

Modifications to the CLI for the predictable address type `instantiate2` was introduced


x/wasm/keeper/keeper.go
```diff
@@ -1005,21 +1027,48 @@ func (k Keeper) consumeRuntimeGas(ctx sdk.Context, gas uint64) {
...
+	creator = address.MustLengthPrefix(creator)
+	salt = address.MustLengthPrefix(salt)
+	key := make([]byte, len(checksum)+len(creator)+len(salt))
+	copy(key[0:], checksum)
+	copy(key[len(checksum):], creator)
+	copy(key[len(checksum)+len(creator):], salt)
+	return address.Module(types.ModuleName, key)[:types.ContractAddrLen]
+}
+
+func BuildContractAddress3(checksum []byte, creator sdk.AccAddress, salt, initMsg []byte) sdk.AccAddress {
```
alpe, line 1045: variation of `BuildContractAddress2` that includes the initMsg

proto/cosmwasm/wasm/v1/tx.proto
```diff
@@ -64,6 +69,31 @@ message MsgInstantiateContract {
...
+  bytes msg = 5 [ (gogoproto.casttype) = "RawContractMessage" ];
+  // Funds coins that are transferred to the contract on instantiation
+  repeated cosmos.base.v1beta1.Coin funds = 6 [
+    (gogoproto.nullable) = false,
+    (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coins"
+  ];
+  // Salt is an arbitrary value provided by the sender
+  bytes salt = 7;
+  // IncludeInitMsg into the hash for the predictable address. Default is false
+  bool include_init_msg = 8;
```
webmaster128, line 94: "init" was removed from all places as far as I can tell. It's no good abbreviation for "instantiate". Also in the context of MsgInstantiateContract2, the message is always included. We need something that makes clear this is for the address generation.

I suggest `fix_msg` which means we fix the message as part of the address generation.

proto/cosmwasm/wasm/v1/tx.proto
```diff
@@ -64,6 +69,31 @@ message MsgInstantiateContract {
...
+  string label = 4;
+  // Msg json encoded message to be passed to the contract on instantiation
+  bytes msg = 5 [ (gogoproto.casttype) = "RawContractMessage" ];
+  // Funds coins that are transferred to the contract on instantiation
+  repeated cosmos.base.v1beta1.Coin funds = 6 [
+    (gogoproto.nullable) = false,
+    (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coins"
+  ];
+  // Salt is an arbitrary value provided by the sender
+  bytes salt = 7;
```
webmaster128, line 92: We should have some size constraints here

proto/cosmwasm/wasm/v1/tx.proto
```diff
@@ -72,6 +102,14 @@ message MsgInstantiateContractResponse {
   bytes data = 2;
 }
 
+// MsgInstantiateContract2Response return instantiation result data
+message MsgInstantiateContract2Response {
+  // Address is the bech32 address of the new contract instance.
+  string address = 1;
+  // Data contains base64-encoded bytes to returned from the contract
```
webmaster128, line 109: "base64-encoded" is outdated, right? Sounds like copied from some JSON interface (same for `MsgInstantiateContractResponse` above).

x/wasm/client/cli/genesis_msg.go
```diff
@@ -333,28 +317,22 @@ func GetAllContracts(state *types.GenesisState) ([]ContractMeta, error) {
 		}
 	}
 	// add inflight
+	seq := contractSeqValue(state)
 	for _, m := range state.GenMsgs {
 		if msg := m.GetInstantiateContract(); msg != nil {
```
webmaster128, line 322: Do we need to add a case here for the new `InstantiateContract2` in `GenMsgs`?

x/wasm/keeper/addresses.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+// BuildContractAddressPredictable1 generates a contract address for the wasm module with len = types.ContractAddrLen using the
+// Cosmos SDK address.Module function.
+// Internally a key is built containing (len(checksum) | checksum | len(sender_address) | sender_address | len(salt) | salt).
+// All method parameter values must be valid and not be empty or nil.
+func BuildContractAddressPredictable1(checksum []byte, creator sdk.AccAddress, salt []byte) sdk.AccAddress {
+	checksum = address.MustLengthPrefix(checksum)
+	creator = address.MustLengthPrefix(creator)
+	salt = address.MustLengthPrefix(salt)
```
webmaster128, line 48: This practically limits the salt to 255 bytes due to implementation details)

x/wasm/keeper/addresses.go
```diff
@@ -0,0 +1,71 @@
...
+	initMsg = address.MustLengthPrefix(initMsg)
+	key := make([]byte, len(checksum)+len(creator)+len(salt)+len(initMsg))
+	copy(key[0:], checksum)
+	copy(key[len(checksum):], creator)
+	copy(key[len(checksum)+len(creator):], salt)
+	copy(key[len(checksum)+len(creator)+len(salt):], initMsg)
+	return address.Module(types.ModuleName, key)[:types.ContractAddrLen]
+}
```
webmaster128, line 71: Do we need the two implementations? An empty `msg` can be used for the case where the message should not be included. Since the empty message is not valid JSON, this can never be an actual message.

x/wasm/keeper/contract_keeper.go
```diff
@@ -43,8 +55,40 @@ func (p PermissionedKeeper) Create(ctx sdk.Context, creator sdk.AccAddress, wasm
...
+) (sdk.AccAddress, []byte, error) {
+	return p.nested.instantiate(
+		ctx,
+		codeID,
+		creator,
+		admin,
+		initMsg,
+		label,
+		deposit,
+		PredicableAddressGenerator(creator, salt, initMsg, includeInitMsg),
```
webmaster128, line 89: Can't you just calculate the address here and send the full address down the stack? Then we can avoid the function creating a function and `AddressGenerator` type.

x/wasm/types/validation.go
```diff
@@ -5,7 +5,10 @@ import (
 )
 
 var (
-	// MaxLabelSize is the longest label that can be used when Instantiating a contract
+	// MaxSaltSize is the longest label that can be used when instantiating a contract
+	MaxSaltSize = 128 // extension point for chains to customize via compile flag.
```
webmaster128, line 9: The way the length encoding is specified does not allow value > 255. So I don't think this should be configurable.

x/wasm/client/cli/genesis_msg.go
alpe: I commented on the `GenMsgs` proto that this is not supported. It is a very complex part of the system that we should remove instead https://github.com/CosmWasm/wasmd/issues/987

x/wasm/keeper/contract_keeper.go
alpe: 👍 The module needs to own the address generation but this is one indirection that can be avoided indeed

x/wasm/keeper/addresses.go
alpe: 👍 I will use this as max size for now.

x/wasm/keeper/addresses.go
alpe: My motivation was to not include the `len(0)` for the msg into the hash but on the other hand, why not. Better to do this and remove the whole method 👍

ethanfrey:
> Modifications to the CLI for the predictable address type (instantiate2):

> --salt [hex,optional]
> --fix-msg [bool,optional]

I am confused of `--salt` optional.
It is required for instantiate2.
If I understand correctly, if not provided, it uses old algorithm. If provided, it uses new algorithm?

Just we need to document that better than "optional" (I will review cli help text)

x/wasm/keeper/addresses.go
```diff
@@ -0,0 +1,58 @@
...
+	"encoding/binary"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/types/address"
+
+	"github.com/CosmWasm/wasmd/x/wasm/types"
+)
+
+// AddressGenerator abstract address generator to be used for a single contract address
+type AddressGenerator func(ctx sdk.Context, codeID uint64, checksum []byte) sdk.AccAddress
```
ethanfrey, line 13: Nice level of abstraction to keep this clean in the callers.

x/wasm/keeper/contract_keeper.go
```diff
@@ -43,8 +55,40 @@ func (p PermissionedKeeper) Create(ctx sdk.Context, creator sdk.AccAddress, wasm
 	return p.nested.create(ctx, creator, wasmCode, instantiateAccess, p.authZPolicy)
 }
 
-func (p PermissionedKeeper) Instantiate(ctx sdk.Context, codeID uint64, creator, admin sdk.AccAddress, initMsg []byte, label string, deposit sdk.Coins) (sdk.AccAddress, []byte, error) {
-	return p.nested.instantiate(ctx, codeID, creator, admin, initMsg, label, deposit, p.authZPolicy)
+// Instantiate creates an instance of a WASM contract using the classic sequence based address generator
```
ethanfrey, line 58: Okay, cool, how you wrap these two entry points here to that one inner function for maximal code reuse

x/wasm/keeper/contract_keeper_test.go
```diff
@@ -0,0 +1,148 @@
...
+	exampleContract := func(t *testing.T, ctx sdk.Context, fixMsg bool) {
+		_, _, err := keepers.ContractKeeper.Instantiate2(
+			ctx,
+			example.CodeID,
+			example.CreatorAddr,
+			nil,
+			initMsg,
+			"my label",
+			sdk.NewCoins(sdk.NewInt64Coin("denom", 1)),
+			[]byte(`my salt`),
```
ethanfrey, line 35: nit, maybe make this a variable as you use it many places?

`mySalt := []byte("my salt")`

(totally not needed for merge, just a thought)
same with `label := "my label"`

x/wasm/keeper/contract_keeper_test.go
```diff
@@ -0,0 +1,148 @@
...
+		)
+		require.NoError(t, err)
+	}
+	exampleWithFixMsg := func(t *testing.T, ctx sdk.Context) {
+		exampleContract(t, ctx, true)
+	}
+	exampleWithoutFixMsg := func(t *testing.T, ctx sdk.Context) {
+		exampleContract(t, ctx, false)
+	}
+	specs := map[string]struct {
```
ethanfrey, line 46: Nice set of cases

x/wasm/keeper/keeper.go
```diff
@@ -277,19 +285,15 @@ func (k Keeper) instantiate(ctx sdk.Context, codeID uint64, creator, admin sdk.A
...
 	}
-	var codeInfo types.CodeInfo
-	k.cdc.MustUnmarshal(bz, &codeInfo)
-
 	if !authPolicy.CanInstantiateContract(codeInfo.InstantiateConfig, creator) {
 		return nil, nil, sdkerrors.Wrap(sdkerrors.ErrUnauthorized, "can not instantiate")
 	}
 
-	contractAddress := BuildContractAddress(codeInfo.CodeHash, creator, label)
+	contractAddress := addressGenerator(ctx, codeID, codeInfo.CodeHash)
```
ethanfrey, line 296: Nice way to make it flexible and avoid re-reading the codeInfo unnecessarily in the addressGenerator. :+1:

webmaster128:
> I am confused of `--salt` optional.

This is because in the current implementation, no `instantiate2` CLI command was created. But I think we should have `instantiate` and `instantiate2` separately.

x/wasm/keeper/msg_server.go
```diff
@@ -74,6 +75,37 @@ func (m msgServer) InstantiateContract(goCtx context.Context, msg *types.MsgInst
...
+		return nil, sdkerrors.Wrap(err, "sender")
+	}
+	var adminAddr sdk.AccAddress
+	if msg.Admin != "" {
+		if adminAddr, err = sdk.AccAddressFromBech32(msg.Admin); err != nil {
+			return nil, sdkerrors.Wrap(err, "admin")
+		}
+	}
+
+	ctx.EventManager().EmitEvent(sdk.NewEvent(
```
ethanfrey, line 93: This event is combined with the message type, right?

We need to document that people basing on events will have to search for two event values now.

x/wasm/types/validation.go
```diff
@@ -28,7 +31,18 @@ func ValidateLabel(label string) error {
...
-		return sdkerrors.Wrap(ErrLimit, "cannot be longer than 128 characters")
+		return ErrLimit.Wrapf("cannot be longer than %d characters", MaxLabelSize)
+	}
+	return nil
+}
+
+// ValidateSalt ensure salt constraints
+func ValidateSalt(salt []byte) error {
+	switch n := len(salt); {
+	case n == 0:
```
ethanfrey, line 42: Seems reasonable validation

ethanfrey (approved):
Looks good and fine to merge as is.

Comments are either stylistic (on test code) or documentation to add (or just praising the code).

x/wasm/keeper/addresses.go
```diff
@@ -0,0 +1,58 @@
...
+// Cosmos SDK address.Module function. Similar to BuildContractAddressPredictable1 but including the initMsg.
+// Internally a key is built containing:
+// (len(checksum) | checksum | len(sender_address) | sender_address | len(salt) | salt| len(initMsg) | initMsg).
+//
+// All method parameter values must be valid and not nil.
+func BuildContractAddressPredictable(checksum []byte, creator sdk.AccAddress, salt, initMsg []byte) sdk.AccAddress {
+	checksum = address.MustLengthPrefix(checksum)
+	creator = address.MustLengthPrefix(creator)
+	salt = address.MustLengthPrefix(salt)
+	initMsg = address.MustLengthPrefix(initMsg)
```
webmaster128, line 51: Here we need a different solution since 1 byte lengths is not enough

webmaster128:
I updated the tests at https://gist.github.com/webmaster128/e4d401d414bd0e7e6f70482f11877fbe (https://github.com/cosmos/cosmjs/pull/1253) to the version on this PR. `msg: null` means `fix_msg: false`. A solution to the msg length problem is not yet implemented.

contrib/local/02-contracts.sh
```diff
@@ -27,6 +30,16 @@ wasmd tx wasm instantiate "$CODE_ID" "$INIT" --admin="$(wasmd keys show validato
 
 CONTRACT=$(wasmd query wasm list-contract-by-code "$CODE_ID" -o json | jq -r '.contracts[-1]')
 echo "* Contract address: $CONTRACT"
+
+echo "## Create new contract instance with predictable address"
+wasmd tx wasm instantiate "$CODE_ID" "$INIT" --admin="$(wasmd keys show validator -a)" \
+  --from validator --amount="100ustake" --label "local0.1.0" \
+  --salt=$(echo -n "testing" | xxd -ps) --fix-msg \
```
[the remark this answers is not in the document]
webmaster128: I'm very much in favor of hex instead os base64 for such things. Strings are not an option because they cannot hold all values the user wants to use. E.g. if you use hashes or int counters as the salt, you can end of with non-UTF8 data.

x/wasm/keeper/keeper_test.go
```diff
@@ -2087,68 +2037,68 @@ func TestBuildContractAddress(t *testing.T) {
...
-			label:      "x",
+			salt:       []byte("x"),
 			creator:    "purple1nxvenxve42424242hwamhwamenxvenxvhxf2py",
 			expAddress: "purple16pc8gt824lmp3dh2sxvttj0ykcs02n5p3ldswhv3j7y853gghlfq7mqeh9",
 		},
-		"max label size": {
+		"max salt size": {
 			checksum:   fromHex("13A1FC994CC6D1C81B746EE0C0FF6F90043875E0BF1D9BE6B7D779FC978DC2A5"),
-			label:      strings.Repeat("x", types.MaxLabelSize),
+			salt:       bytes.Repeat([]byte("x"), types.MaxLabelSize),
```
webmaster128, line 2101: If this tests passes then we should have a salt size check or assertion in BuildContractAddressPredictable because this is too long (due to bug here)

x/wasm/keeper/addresses.go
alpe: you are right. 255 may not be enough!

webmaster128 (commented):
Very nice!

The two open issues for me are
- I strongly believe we should have two different CLI commands. `instantiate` and `instantiate2` are two different operations, which is reflected in two messages and will lead to two high level client functions `instantiate` and `instantiate2` in CosmJS. Using two has a number of advantages. First, existing `instantiate` users do not need to bother about the new arguments which just confuse them and might lead to misusage. Second, by making all the new important fields required for the instantiate2 case, we get much stronger input validation. Third, the code becomes easier to audit since less code branches exists per command.
- The `len(msg)` can easily exceed 255. We can either make this one length field a big endian uin32 or uint64. Or (my preference) we use the same length implementation for all fields and move away to the address specific prefixer function. In our use case we do not need to save space (in contrast to the storage use case for which the address prefixer was written for), so we can just use uint64 for all of them.

x/wasm/keeper/addresses.go
alpe: I have pushed a change. The slice can not be > max int  https://stackoverflow.com/questions/27647737/maximum-length-of-a-slice-in-go

x/wasm/keeper/keeper_test.go
alpe: There were no sanity checks in the address builder. The data can not be invalid when called by a message but I add some constraints when called as a library.

x/wasm/keeper/msg_server.go
alpe: ja, see below (this is the list of events for the `hackatom` example)

```json
[
  {
    "events": [
      {
        "type": "coin_received",
        "attributes": [
          {
            "key": "receiver",
            "value": "wasm1ayj3kc8grpe7axlvdhc0etrhev9knjs0v4ghwc0qn8hda3mdathq75ahs3"
          },
          {
            "key": "amount",
            "value": "100ustake"
          }
        ]
      },
      {
        "type": "coin_spent",
        "attributes": [
          {
            "key": "spender",
            "value": "wasm1alz4eapgf0rq3ezf3xge9pz7uhkpdyu2mmrrup"
          },
          {
            "key": "amount",
            "value": "100ustake"
          }
        ]
      },
      {
        "type": "instantiate",
        "attributes": [
          {
            "key": "_contract_address",
            "value": "wasm1ayj3kc8grpe7axlvdhc0etrhev9knjs0v4ghwc0qn8hda3mdathq75ahs3"
          },
          {
            "key": "code_id",
            "value": "1"
          }
        ]
      },
      {
        "type": "message",
        "attributes": [
          {
            "key": "action",
            "value": "/cosmwasm.wasm.v1.MsgInstantiateContract2"
          },
          {
            "key": "module",
            "value": "wasm"
          },
          {
            "key": "sender",
            "value": "wasm1alz4eapgf0rq3ezf3xge9pz7uhkpdyu2mmrrup"
          }
        ]
      },
      {
        "type": "transfer",
        "attributes": [
          {
            "key": "recipient",
            "value": "wasm1ayj3kc8grpe7axlvdhc0etrhev9knjs0v4ghwc0qn8hda3mdathq75ahs3"
          },
          {
            "key": "sender",
            "value": "wasm1alz4eapgf0rq3ezf3xge9pz7uhkpdyu2mmrrup"
          },
          {
            "key": "amount",
            "value": "100ustake"
          }
        ]
      },
      {
        "type": "wasm",
        "attributes": [
          {
            "key": "_contract_address",
            "value": "wasm1ayj3kc8grpe7axlvdhc0etrhev9knjs0v4ghwc0qn8hda3mdathq75ahs3"
          },
          {
            "key": "Let the",
            "value": "hacking begin"
          }
        ]
      }
    ]
  }
]
```

x/wasm/keeper/addresses.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+// UInt64LengthPrefix prepend big endian encoded byte length
+func UInt64LengthPrefix(bz []byte) ([]byte, error) {
+	bzLen := len(bz)
+	if bzLen == 0 {
+		return bz, nil
+	}
+	return append(sdk.Uint64ToBigEndian(uint64(bzLen)), bz...), nil
+}
```
webmaster128, line 80: Here we don't need two functions because this has no error case. The only error case for address.LengthPrefix in Cosmos SDK is length > 255.

x/wasm/keeper/addresses.go
```diff
@@ -0,0 +1,89 @@
...
+	contractID := make([]byte, 16)
+	binary.BigEndian.PutUint64(contractID[:8], codeID)
+	binary.BigEndian.PutUint64(contractID[8:], instanceID)
+	return address.Module(types.ModuleName, contractID)[:types.ContractAddrLen]
+}
+
+// BuildContractAddressPredictable generates a contract address for the wasm module with len = types.ContractAddrLen using the
+// Cosmos SDK address.Module function.
+// Internally a key is built containing:
+// (len(checksum) | checksum | len(sender_address) | sender_address | len(salt) | salt| len(initMsg) | initMsg).
```
webmaster128, line 45: You want to use different `len` functions for the `msg` and the other fields? Then this should be clearly reflected here as well.

x/wasm/keeper/addresses.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	if err := sdk.VerifyAddressFormat(creator); err != nil {
+		panic(fmt.Sprintf("creator: %s", err))
+	}
+	if err := types.ValidateSalt(salt); err != nil {
+		panic(fmt.Sprintf("salt: %s", err))
+	}
+	if err := initMsg.ValidateBasic(); len(initMsg) != 0 && err != nil {
+		panic(fmt.Sprintf("initMsg: %s", err))
+	}
```
webmaster128, line 60: Nice checks

x/wasm/keeper/addresses.go
alpe: It was late already 🤦

x/wasm/keeper/addresses.go
alpe: The example did include types so it was still correct IMHO but I added them now to be more precise

contrib/local/02-contracts.sh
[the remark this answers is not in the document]
alpe: with the salt as an argument, we can be more flexible. I make hex default in the decoder. Other options are base64 and ascii. This is an existing component that I will use

x/wasm/keeper/msg_server.go
ethanfrey: I guess they could also search for type: instantiate?

x/wasm/module.go
```diff
@@ -134,11 +134,6 @@ func NewAppModule(
 func (am AppModule) RegisterServices(cfg module.Configurator) {
 	types.RegisterMsgServer(cfg.MsgServer(), keeper.NewMsgServerImpl(keeper.NewDefaultPermissionKeeper(am.keeper)))
 	types.RegisterQueryServer(cfg.QueryServer(), NewQuerier(am.keeper))
-
-	m := keeper.NewMigrator(am.keeper)
```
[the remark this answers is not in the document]
alpe: The whole migrate.go file was removed https://github.com/CosmWasm/wasmd/pull/1014/files#diff-4357c2137e24f583b8f852cc210320cb71af18e2fdfb8c21b55d8667cfe54690

x/wasm/client/cli/tx.go
```diff
@@ -174,24 +178,30 @@ func parseAccessConfigFlags(flags *flag.FlagSet) (*types.AccessConfig, error) {
 // InstantiateContractCmd will instantiate a contract from previously uploaded code.
 func InstantiateContractCmd() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:     "instantiate [code_id_int64] [json_encoded_init_args] --label [text] --admin [address,optional] --amount [coins,optional]",
-		Short:   "Instantiate a wasm contract",
+		Use: "instantiate [code_id_int64] [json_encoded_init_args] --label [text] --admin [address,optional] --amount [coins,optional] " +
+			"--salt [hex,optional] --fix-msg [bool,optional]",
```
webmaster128, line 182:
```suggestion
		Use: "instantiate [code_id_int64] [json_encoded_init_args] --label [text] --admin [address,optional] --amount [coins,optional]",
```

x/wasm/client/cli/tx.go
```diff
@@ -203,43 +213,105 @@ func InstantiateContractCmd() *cobra.Command {
...
+// InstantiateContract2Cmd will instantiate a contract from previously uploaded code with predicable address generated
+func InstantiateContract2Cmd() *cobra.Command {
+	decoder := newArgDecoder(hex.DecodeString)
+	cmd := &cobra.Command{
+		Use: "instantiate2 [code_id_int64] [json_encoded_init_args] [salt] --label [text] --admin [address,optional] --amount [coins,optional] " +
+			"--fix-msg [bool,optional]",
+		Short: "Instantiate a wasm contract with predictable address",
+		Long: fmt.Sprintf(`Creates a new instance of an uploaded wasm code with the given 'constructor' message.
+Each contract instance has a unique address assigned. They are assigned automatically but in order to have predictable addresses 
+for special use cases, the given 'salt' argument and '--fix-mesg' parameters can be used to generate a custom address.
```
webmaster128, line 225:
```suggestion
for special use cases, the given 'salt' argument and '--fix-msg' parameters can be used to generate a custom address.
```

x/wasm/client/cli/tx.go
```diff
@@ -203,43 +213,105 @@ func InstantiateContractCmd() *cobra.Command {
...
+
+Predictable address example (also see '%s query wasm build-address -h'):
+$ %s wasmd tx wasm instantiate 1 '{"foo":"bar"}' --admin="$(%s keys show mykey -a)" \
```
webmaster128, line 228:
```suggestion
$ %s wasmd tx wasm instantiate2 1 '{"foo":"bar"}' --admin="$(%s keys show mykey -a)" \
```

x/wasm/client/cli/tx.go
```diff
@@ -203,43 +213,105 @@ func InstantiateContractCmd() *cobra.Command {
...
+  --from mykey --amount="100ustake" --label "local0.1.0" \
+  --salt=$(echo -n "testing" | xxd -ps) --fix-msg
```
webmaster128, line 230: Salt needs to be positional now, right?

webmaster128 (commented):
Very nice! Just a few smaller things but overall good to go as fas as I can see.

x/wasm/client/cli/tx.go
alpe: needs to be an argument. I will fix the example

PR #1032, opened 2022-10-04
Revert module version to 1 as there is no migration anymore

alpe:
> In RC0, the wasmd module was bumped to version '2' with a migration. In RC1 the migration was dropped as we restored the "classic" contract address generation process. Though I kept the module version at '2' for testnets running RC0 (Juno). 

This PR reverts the module version to `1`


pinosu (approved):
LGTM 👍

PR #1036, opened 2022-10-06
Make SenderPrivKey field public

### LINKED ISSUE #1019: Make `senderPrivateKey` public on `ibctesting.TestChain`
The specific thing I run into often is wanting to test that a call to `chain.SendMsgs` results in an error. The `chain.SendMsgs` function calls `wasmd.SignAndDeliver` with `expPass` and `expSimPass` both set to `true`, this means that if an error occurs during the transaction's execution the test will immediately fail, making it hard to pass a test that wants a failure.

That wouldn't be too much trouble to just write a modified version, but sadly the `senderPrivateKey` field on chain is private so it can't be used in a custom version of `SendMsgs`.

I really have no idea what I'm doing, but the best thing I've been [able to come up](https://github.com/public-awesome/ics721/blob/7f791d15ad42566c8a3143911caec63ff7c4db55/e2e/adversarial_test.go#L168-L177) with is to create a new account, store that private key, and then write a custom `SendMsg` function that allows failures.

One option here would be to just add a `shouldPass` boolean to the `SendMsgs` method, but just making the private key available in the public api feels like it would make things generally more flexible.

If this sounds reasonable I'm happy to make a pr. :)

pinosu:
Resolves #1019

### DIFF
--- a/x/wasm/ibctesting/chain.go
+++ b/x/wasm/ibctesting/chain.go
@@ -60,7 +60,7 @@ type TestChain struct {
 	Vals    *tmtypes.ValidatorSet
 	Signers []tmtypes.PrivValidator
 
-	senderPrivKey cryptotypes.PrivKey
+	SenderPrivKey cryptotypes.PrivKey
 	SenderAccount authtypes.AccountI
 
 	PendingSendPackets []channeltypes.Packet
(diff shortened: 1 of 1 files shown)


alpe (approved):
Nice to get this ready for v0.29.0

ezekiiel:
thank you @pinosu !

PR #1047, opened 2022-10-10
Validate incoming messages

alpe:
Validate incoming data on the service level, too


pinosu (approved):
LGTM 👍

ethanfrey:
Nice one. Can never be too safe,.

PR #1050, opened 2022-10-13
Permission

### LINKED ISSUE #1020: Support instantiation permission > chain permission for gov store
> Chain permission parameters are set at Nobody for both store code and instantiation. When we try to submit a proposal to store a contract code with different instantiation config than Nobody it gives error as the permission must be subset of the Nobody permission. What I'm confused about is, since it anyways goes via a governance vote and the instantiation config is visible to the voters, why can't we directly upload with right permissions directly. Currently, it requires 2 governance proposals (one for store, one for config update) to be floated which is not ideal imo since one can also serve the same purpose. It would have made sense if we never allowed any contract to have instantiation config outside the subset of the chain config, but since we do, how does it matter that if we have 2 proposals or 1 to achieve exactly the same config. 

https://discord.com/channels/737637324434833438/737643344712171600/1017724135897579621

vuong177:
Close #1020

### DIFF
--- a/x/wasm/keeper/authz_policy.go
+++ b/x/wasm/keeper/authz_policy.go
@@ -6,8 +6,19 @@ import (
 	"github.com/CosmWasm/wasmd/x/wasm/types"
 )
 
+// ChainAccessConfigs chain settings
+type ChainAccessConfigs struct {
+	Upload      types.AccessConfig
+	Instantiate types.AccessConfig
+}
+
+// NewChainAccessConfigs constructor
+func NewChainAccessConfigs(upload types.AccessConfig, instantiate types.AccessConfig) ChainAccessConfigs {
+	return ChainAccessConfigs{Upload: upload, Instantiate: instantiate}
+}
+
 type AuthorizationPolicy interface {
-	CanCreateCode(c1 types.AccessConfig, creator sdk.AccAddress, c2 types.AccessConfig, c3 types.AccessConfig) bool
+	CanCreateCode(chainConfigs ChainAccessConfigs, actor sdk.AccAddress, contractConfig types.AccessConfig) bool
 	CanInstantiateContract(c types.AccessConfig, actor sdk.AccAddress) bool
 	CanModifyContract(admin, actor sdk.AccAddress) bool
 	CanModifyCodeAccessConfig(creator, actor sdk.AccAddress, isSubset bool) bool
@@ -15,12 +26,9 @@ type AuthorizationPolicy interface {
 
 type DefaultAuthorizationPolicy struct{}
 
-func (p DefaultAuthorizationPolicy) CanCreateCode(config types.AccessConfig, actor sdk.AccAddress, instantiateAccess types.AccessConfig, defaultAccessConfig types.AccessConfig) bool {
-	isSubset := true
-	if !instantiateAccess.IsSubset(defaultAccessConfig) {
-		isSubset = false
-	}
-	return config.Allowed(actor) && isSubset
+func (p DefaultAuthorizationPolicy) CanCreateCode(chainConfigs ChainAccessConfigs, actor sdk.AccAddress, contractConfig types.AccessConfig) bool {
+	return chainConfigs.Instantiate.Allowed(actor) &&
+		contractConfig.IsSubset(chainConfigs.Upload)
 }
 
 func (p DefaultAuthorizationPolicy) CanInstantiateContract(config types.AccessConfig, actor sdk.AccAddress) bool {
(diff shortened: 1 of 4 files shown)


x/wasm/keeper/keeper.go
```diff
@@ -187,16 +187,20 @@ func (k Keeper) create(ctx sdk.Context, creator sdk.AccAddress, wasmCode []byte,
...
-	// figure out proper instantiate access
 	defaultAccessConfig := k.getInstantiateAccessConfig(ctx).With(creator)
 	if instantiateAccess == nil {
 		instantiateAccess = &defaultAccessConfig
-	} else if !instantiateAccess.IsSubset(defaultAccessConfig) {
-		// we enforce this must be subset of default upload access
-		return 0, checksum, sdkerrors.Wrap(sdkerrors.ErrUnauthorized, "instantiate access must be subset of default upload access")
+	}
+
+	if !authZ.IsGovAuthorization() {
```
alpe, line 195: This is the switch that I wanted to avoid by moving the conditions below fully into the  policies. My motivation was the [separation of concerns](https://en.wikipedia.org/wiki/Separation_of_concerns). The AuthorizationPolicies are an extension point where custom implementations can be used for individual decision logic. This would not be possible if we keep the code here.
Ideally we have a 3 liner here like:
```
if !authZ.CanCreateCode(chainDefaults, creator, instantiateAccess) {
	return 0, checksum, sdkerrors.Wrap(sdkerrors.ErrUnauthorized, "can not create code")
}
```
I hope this makes sense for you, too

alpe (commented):
Thank you for your PR! This is a straight forward solution.
It is breaking some design concepts, though that were not documented well. Please see my comment.

x/wasm/keeper/keeper.go
vuong177: @alpe thanks for your suggestion. Is it fine if I move the subset check into CanCreateCode like this ?: 
```
func (p DefaultAuthorizationPolicy) CanCreateCode(config types.AccessConfig, actor sdk.AccAddress, defaultConfig types.AccessConfig) bool {
	isSubset := true
	if !config.IsSubset(defaultConfig) {
		isSubset = false
	}
	return config.Allowed(actor) && isSubset
}
```

vuong177: I've force-pushed my code into this branch to avoid having too much PR!

vuong177: tysm for your suggestion!

x/wasm/keeper/authz_policy.go
```diff
@@ -7,16 +7,20 @@ import (
...
 	CanModifyCodeAccessConfig(creator, actor sdk.AccAddress, isSubset bool) bool
 }
 
 type DefaultAuthorizationPolicy struct{}
 
-func (p DefaultAuthorizationPolicy) CanCreateCode(config types.AccessConfig, actor sdk.AccAddress) bool {
-	return config.Allowed(actor)
+func (p DefaultAuthorizationPolicy) CanCreateCode(config types.AccessConfig, actor sdk.AccAddress, instantiateAccess types.AccessConfig, defaultAccessConfig types.AccessConfig) bool {
+	isSubset := true
+	if !instantiateAccess.IsSubset(defaultAccessConfig) {
```
alpe, line 20: this can be a 1 liner as `IsSubset` returns a bool

x/wasm/keeper/authz_policy.go
```diff
@@ -7,16 +7,20 @@ import (
...
+		isSubset = false
+	}
+	return config.Allowed(actor) && isSubset
```
alpe, line 23: 👍 all decision logic in one place and no "if gov then" logic. nice

alpe (changes requested):
Thanks for the changes! The decision logic is in the right place now but the method params can be very confusing. I have opened  https://github.com/CosmWasm/wasmd/pull/1054 for some improvements. Please take a look and let me know if this makes sense for you, too.

There is a tests missing that covers the gov cases now. Maybe `TestCreateStoresInstantiatePermission` is a good candidate to extract the policy and test with both policies. The code changes are small but important to cover for regression. 🙏

x/wasm/keeper/authz_policy.go
vuong177: @alpe yeah I've read your suggestion and it makes more sense.
I'm very appreciative of all the support you've given me.

alpe (commented):
Thanks for the updates! Again very good progress. Unfortunately, I paid more attention to the interface design than the actual impl so that I missed the error. 🙈 
The tests should have covered that. 

My rule of thump is making them pass first and then add new cases. When tests were modified by others it can be better to start over by resetting them from main. It is easy to miss small changes. I have rebased and updated #1054 to break fixing the tests into smaller steps.

PR #1052, opened 2022-10-14
update cosmos-sdk to v45.9, add dragonberry.

puneet2019:
- updated cosmos-sdk
- replace confio/ics23 to cosmos/ics23
- updated GenTx function in helper as GenTx now does it within itself.


ethanfrey (approved):
Thanks for this quick patch

PR #1054, opened 2022-10-17
[Do not merge] 1050 feedback

alpe:
Example for #1050 api design

@vuong177  please 👁️  and copy if this makes sense for you, too


alpe:
obsolete

PR #1060, opened 2022-10-24
Update instantiate cli example

giansalex:
Current example shows:
```sh
Example:
$ junod wasmd tx wasm instantiate 1 '{"foo":"bar"}' --admin="$(junod keys show mykey -a)" \
  --from mykey --amount="100ustake" --label "local0.1.0"
```


pinosu (approved):
LGTM 👍

PR #1064, opened 2022-10-27
Add simulation for gov

### LINKED ISSUE #1063: Add simulation for gov proposal
Follow with #797 
We need to add simulation for gov proposal

GNaD13:
#1063 
ad sims for gov proposal:
- Instantiate contract
- Execute contract
- Update admin
- Clear admin


x/wasm/simulation/proposals.go
```diff
@@ -0,0 +1,221 @@
...
+			SimulateClearAdminProposal(
+				wasmKeeper,
+				DefaultSimulateClearAdminProposalContractSelector,
+			),
+		),
+	}
+}
+
+// simulate store code proposal (unused now)
+// Current problem: out of gas (defaul gaswanted config of gov SimulateMsgSubmitProposal is 10_000_000)
```
alpe, line 69: Do you have a proposal for a better gas value? We can open an issue/PR on the SDK to get this addressed

alpe (commented):
Thanks a lot for your contribution! 👍 

Unfortunately the simulations do fail [now](https://app.circleci.com/pipelines/github/CosmWasm/wasmd/2637/workflows/e94f215c-947c-479b-8661-59de8c93b87a/jobs/9588?invite=true#step-102-46) . The issue does not look related but please take a look.

Note: current `main` contains sdk v0.45.10 now which seem to not cause the error

GNaD13:
> Thanks a lot for your contribution! +1
> 
> Unfortunately the simulations do fail [now](https://app.circleci.com/pipelines/github/CosmWasm/wasmd/2637/workflows/e94f215c-947c-479b-8661-59de8c93b87a/jobs/9588?invite=true#step-102-46) . The issue does not look related but please take a look.
> 
> Note: current `main` contains sdk v0.45.10 now which seem to not cause the error

Yeah, I have checked the issue. The errors occur in the sims of module authz [(this issues)](url). It will pass if we move to v0.45.10

x/wasm/simulation/proposals.go
GNaD13: I think we have 2 solution to solve this problem:
- Open a PR to increase defaul gaswanted so that this proposal can easily pass
- Write a simple .wasm code which size is smaller than current. So that it can decrease the gas used.
But the 2 solution quite difficult. Because this may change all the current sims test. One more thing, in the future, we need to implement sudo entry point for Sudo proposal (it will increase the size of .wasm contract)

alpe: Store code and sudo gov operations are well covered by integration tests already. Simulations are just for regression and fuzzing. They don't add super much value here IMHO. Let's have a follow up task to revisit the store code when we are on SDK v0.47. There is some work on the gov module.

PR #1068, opened 2022-11-01
Add SilenceUsage flag to hide menu when an error occurs.

### LINKED ISSUE #973: Errors styling & hide help menu
### Background
When smart contract reverts transaction, it is hard to find out what happened because of:
- Error string is overloaded
- Help menu scrolls terminal, however all is okay with wasmd configuration

### Example
It looks really crazy:
```
Error: rpc error: code = InvalidArgument desc = Person with this identifier is not found: query wasm contract failed: invalid request
Usage:
  wasmd query wasm contract-state smart [bech32_address] [query] [flags]

Flags:
      --ascii           ascii encoded query argument
      --b64             base64 encoded query argument
      --height int      Use a specific height to query state at (this can error if the node is pruning state)
  -h, --help            help for smart
      --hex             hex encoded  query argument
      --node string     <host>:<port> to Tendermint RPC interface for this chain (default "tcp://localhost:26657")
  -o, --output string   Output format (text|json) (default "text")

Global Flags:
      --chain-id string     The network chain ID
      --home string         directory for config and data (default "/home/wasm_user/.wasmd")
      --log_format string   The logging format (json|plain) (default "plain")
      --log_level string    The logging level (trace|debug|info|warn|error|fatal|panic) (default "info")
      --trace               print out full stack trace on errors
```

### How do I see the solution
- Ability to hide help menu is needed, suppress option and corresponding env variable may be implemented
- Hiding help menu on this type of error (smart contract reverts transaction) may be implemented
- Error response may be simplified to `Error: smart contract paniced. REASON`
- If back compatibility is needed, simplified error response may be shown only when corresponding option is provided

pinosu:
Resolves #973

### DIFF
--- a/x/wasm/client/cli/genesis_msg.go
+++ b/x/wasm/client/cli/genesis_msg.go
@@ -68,6 +68,7 @@ func GenesisStoreCodeCmd(defaultNodeHome string, genesisMutator GenesisMutator) 
 				return nil
 			})
 		},
+		SilenceUsage: true,
 	}
 	cmd.Flags().String(flagRunAs, "", "The address that is stored as code creator")
 	cmd.Flags().String(flagInstantiateByEverybody, "", "Everybody can instantiate a contract from the code, optional")
@@ -136,6 +137,7 @@ func GenesisInstantiateContractCmd(defaultNodeHome string, genesisMutator Genesi
 				return nil
 			})
 		},
+		SilenceUsage: true,
 	}
 	cmd.Flags().String(flagAmount, "", "Coins to send to the contract during instantiation")
 	cmd.Flags().String(flagLabel, "", "A human-readable name for this contract in lists")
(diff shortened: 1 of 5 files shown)


alpe (commented):
Thanks for the PR!

The changes look good but a lot of commands were touched. I was wondering if it is sufficient to set `SilenceUsage: true` on the `GetTxCmd()` only. WDYT?

pinosu:
> Thanks for the PR!
> 
> The changes look good but a lot of commands were touched. I was wondering if it is sufficient to set `SilenceUsage: true` on the `GetTxCmd()` only. WDYT?

I agree that a more general soluton would be better. Unfortunately, I have tried your suggestion as first option but it didn't work. I had to add it to every command.

alpe:
👍 Good work

PR #1072, opened 2022-11-02
feat: Provide source, builder and codehash information in store code proposal message

orkunkl:
This PR's aim is to standardize contract verification on store code proposals.
On store code [proposals](https://www.mintscan.io/stargaze/proposals/67), proposal creators usually pass code source, builder and code hash information as a part of the description. Let's move these fields back to proposal content and run a check during store to verify provided hash is correct.


proto/cosmwasm/wasm/v1/proposal.proto
```diff
@@ -26,6 +26,13 @@ message StoreCodeProposal {
   AccessConfig instantiate_permission = 7;
   // UnpinCode code on upload, optional
   bool unpin_code = 8;
+  // Source is the URL where the code is hosted
+  string source = 9;
```
orkunkl, line 30: Not sure if 9,10,11 are reserved

x/wasm/types/proposal.pb.go
```diff
@@ -44,18 +40,23 @@ type StoreCodeProposal struct {
...
+	// CodeHash is the SHA256 sum of the code outputted by builder, used for smart contract verification
+	CodeHash []byte `protobuf:"bytes,11,opt,name=code_hash,json=codeHash,proto3" json:"code_hash,omitempty"`
 }
 
 func (m *StoreCodeProposal) Reset()      { *m = StoreCodeProposal{} }
 func (*StoreCodeProposal) ProtoMessage() {}
 func (*StoreCodeProposal) Descriptor() ([]byte, []int) {
 	return fileDescriptor_be6422d717c730cb, []int{0}
 }
-
```
orkunkl, line 54: I've used `make proto-gen` to generate proto code. All proto files had these formatting changes, reverted them.

x/wasm/keeper/proposal_handler.go
```diff
@@ -64,11 +65,15 @@ func handleStoreCodeProposal(ctx sdk.Context, k types.ContractOpsKeeper, p types
 	if err != nil {
 		return sdkerrors.Wrap(err, "run as address")
 	}
-	codeID, _, err := k.Create(ctx, runAsAddr, p.WASMByteCode, p.InstantiatePermission)
+	codeID, checksum, err := k.Create(ctx, runAsAddr, p.WASMByteCode, p.InstantiatePermission)
 	if err != nil {
 		return err
 	}
 
+	if !bytes.Equal(checksum, p.CodeHash) {
```
alpe, line 73: There may be in flight proposals without this data. It would be good to not enforce the check when p.CodeHash is empty

x/wasm/types/proposal.go
```diff
@@ -113,6 +113,16 @@ func (p StoreCodeProposal) ValidateBasic() error {
 			return sdkerrors.Wrap(err, "instantiate permission")
 		}
 	}
+
+	if len(p.Source) == 0 {
```
alpe, line 117: As mentioned above, these fields should be optional on the server side to support a transition period for in flight proposals. 
But there is no problem to enforce checks on the CLI before submitting a proposal.
There can be stronger constraints on the expected data (when set).

alpe (commented):
Thanks for bringing this topic up and this very nice PR 💐 

I think validators have a hard time to verify a `StoreCodeProposal` currently. Providing additional data for this process can help with to build better tooling on the client side.
Nevertheless the data needs to be specified well and be verified, so that the expectations for everybody are the same.

We also need to support a transition period (on the server side) where fields are empty to not break execution of existing in flight proposals.

x/wasm/types/proposal.pb.go
alpe: Please use `make proto-all` which also runs the formatter (`make format`)

x/wasm/client/cli/tx_test.go
```diff
@@ -57,3 +63,53 @@ func TestParseAccessConfigFlags(t *testing.T) {
...
+		},
+		"builder missing": {
+			args:   []string{"--source=bar", "--code-hash=" + checksumStr},
+			expErr: true,
+		},
+		"code hash missing": {
+			args:   []string{"--source=bar", "--builder=foo"},
+			expErr: true,
+		},
+		"happy path": {
```
orkunkl, line 97: I am getting an error here that I could not solve.
The hash generated on 76th line and in `parseStoreCodeArgs` do not match. I decided to not get stuck and ask your elven 👀  @alpe here.

x/wasm/client/cli/tx.go
```diff
@@ -108,9 +116,35 @@ func parseStoreCodeArgs(file string, sender sdk.AccAddress, flags *flag.FlagSet)
...
+		return types.MsgStoreCode{}, fmt.Errorf("builder: %s", err)
+	}
+	if builder == "" {
+		return types.MsgStoreCode{}, fmt.Errorf("builder: must not be empty")
+	}
+	codeHash, err := flags.GetBytesHex(flagCodeHash)
+	if err != nil {
+		return types.MsgStoreCode{}, fmt.Errorf("code-hash: %s", err)
+	}
+	checksum := sha256.Sum256(wasm)
```
alpe, line 137: You are building the hash on the zipped content. I has to be the raw contract instead

Please note that wasmvm is using sha256 but this may change in the future. I have opened https://github.com/CosmWasm/wasmvm/issues/359 to no have this coupling here. Can you link to the issue on this line? IMHO it makes sense to fail fast here

x/wasm/client/cli/tx_test.go
```diff
@@ -57,3 +63,53 @@ func TestParseAccessConfigFlags(t *testing.T) {
 		})
 	}
 }
+
+func TestParseStoreCodeFlags(t *testing.T) {
+	anyValidWasmFile, err := os.CreateTemp(t.TempDir(), "wasm")
+	require.NoError(t, err)
+
+	anyValidWasmFile.Write(wasmIdent)
```
alpe, line 71: The content is not fully written before the file is closed.

x/wasm/client/cli/tx_test.go
```diff
@@ -57,3 +63,53 @@ func TestParseAccessConfigFlags(t *testing.T) {
...
+
+	wasmBin, err := io.ReadAll(anyValidWasmFile)
+	require.NoError(t, err)
+
+	checkSum := sha256.Sum256(wasmBin)
```
alpe, line 76: Please don't use the same code to generate the hash as it is used to verify it. Instead use the shell to hash it or some other source

x/wasm/client/cli/tx_test.go
alpe: I have commented on `parseStoreCodeArgs` which contains a 🐛

x/wasm/client/cli/tx_test.go
```diff
@@ -57,3 +63,53 @@ func TestParseAccessConfigFlags(t *testing.T) {
...
+			expErr: true,
+		},
+		"code hash missing": {
+			args:   []string{"--source=bar", "--builder=foo"},
+			expErr: true,
+		},
+		"happy path": {
+			args:   []string{"--source=bar", "--builder=foo", "--code-hash=" + checksumStr},
+			expErr: false,
+		},
```
alpe, line 100: Please add a case where the source file is zipped so that you are not running into a problem when comparing the hash

alpe (commented):
Thanks for the updates but I think updating the TX and gen-messages should be discussed first, given the negative experience on chain in the past. Let's have an issue for them and fully focus on the gov store proposal here to get it into v0.30 if possible

proto/cosmwasm/wasm/v1/tx.proto
```diff
@@ -41,7 +41,16 @@ message MsgStoreCode {
   // InstantiatePermission access control to apply on contract creation,
   // optional
   AccessConfig instantiate_permission = 5;
+  // Source is a valid absolute HTTPS URI to the contract's source code,
```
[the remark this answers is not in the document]
orkunkl: Makes sense :+1:

alpe:
fyi: @pinosu is working on the [StoreAndInstantiateProposal](https://github.com/CosmWasm/wasmd/pull/1074) that would also benefit from the new fields
Do you need help with the implementation?

orkunkl:
@alpe I got carried away with other tasks, I will finish this today/tomorrow thanks!

x/wasm/client/cli/tx.go
```diff
@@ -25,6 +25,9 @@ import (
 const (
 	flagAmount                    = "amount"
 	flagLabel                     = "label"
+	flagSource                    = "code-source-url"
+	flagBuilder                   = "builder"
```
orkunkl, line 29: maybe `code-builder` for consistency?

x/wasm/keeper/proposal_handler.go
alpe: I think you got me wrong. `if !bytes.Equal(checksum, p.CodeHash) ` makes a lot of sense and ensures consistency of the proposal. Please add it back but with a `if p.CodeHash != nil && !bytes.Equal(checksum, p.CodeHash)` so that it works for old in flight proposals (without the new fields) , too.

x/wasm/client/cli/gov_tx.go
```diff
@@ -73,6 +86,9 @@ func ProposalStoreCodeCmd() *cobra.Command {
 	cmd.Flags().String(flagInstantiateByAddress, "", "Only this address can instantiate a contract instance from the code, optional")
 	cmd.Flags().Bool(flagUnpinCode, false, "Unpin code on upload, optional")
 	cmd.Flags().StringSlice(flagInstantiateByAnyOfAddress, []string{}, "Any of the addresses can instantiate a contract from the code, optional")
+	cmd.Flags().String(flagSource, "", "Code Source URL is a valid absolute HTTPS URI to the contract's source code,")
+	cmd.Flags().String(flagBuilder, "", "Builder is a valid docker image name with tag, such as \"cosmwasm/workspace-optimizer:0.12.9\"")
+	cmd.Flags().BytesHex(flagCodeHash, []byte{}, "CodeHash is the sha256 hash of the wasm code")
```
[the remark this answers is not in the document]
orkunkl:
```suggestion
	cmd.Flags().BytesHex(flagCodeHash, nil, "CodeHash is the sha256 hash of the wasm code")
```

x/wasm/client/cli/gov_tx.go
```diff
@@ -81,6 +97,48 @@ func ProposalStoreCodeCmd() *cobra.Command {
 	return cmd
 }
 
+func parseCodeInfoFlags(wasm []byte, flags *flag.FlagSet) (string, string, []byte, error) {
+	source, err := flags.GetString(flagSource)
+	if err != nil {
+		return "", "", []byte{}, fmt.Errorf("source: %s", err)
```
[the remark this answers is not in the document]
orkunkl:
```suggestion
		return "", "", nil, fmt.Errorf("source: %s", err)
```

go.mod
```diff
@@ -10,6 +10,7 @@ require (
 	github.com/cosmos/iavl v0.19.4
 	github.com/cosmos/ibc-go/v3 v3.3.1
 	github.com/cosmos/interchain-accounts v0.1.0
+	github.com/docker/distribution v2.8.1+incompatible
```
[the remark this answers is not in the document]
orkunkl: For docker reference verification I used [reference.ParseDockerRef](https://github.com/distribution/distribution/blob/main/reference/normalize.go#L63)
Here is the line https://github.com/CosmWasm/wasmd/blob/3d74a53c7a137e06d2490820ee7fb4b51fd12bec/x/wasm/types/validation.go#L69-L69

orkunkl:
Maybe best to wrap source, builder and code hash information under `VerificationInfo` and treat them as one pack?

x/wasm/types/validation.go
```diff
@@ -49,3 +53,27 @@ func ValidateSalt(salt []byte) error {
 	}
 	return nil
 }
+
+// ValidateVerificationInfo ensure source, builder and checksum constraints
+func ValidateVerificationInfo(source, builder string, codeHash []byte) error {
+	// if any set require others to be set
+	if len(source) != 0 || len(builder) != 0 || codeHash != nil {
+		if source == "" {
+			return fmt.Errorf("source is required")
```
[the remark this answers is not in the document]
alpe: The sdk package comes with some nice error support. Next time, please use them instead

alpe (approved):
Very good work! Thanks for applying all my requests. 🥇 
I am looking forward to see some client side tooling that makes use of the new fields. Great innovation 👍

faddat:
Incredibly important work. Thank you!

PR #1077, opened 2022-11-04
Contract authz - redesign

### LINKED ISSUE #803: Authz module integration - more granularity for WasmExecuteMsg authorizations

alpe:
Part of #803 - server side

Based on the great work started by @giansalex in #966 

This PR includes a new and more extendable model than #978 . A contract execution or migration grant has:

#### # Supported "limits" - that are stateful and restrict the number of executions

* max calls
* max token transferred
* max calls + max token transferred

#### # Supported "filters" - that check if the submitted message is allowed
* json object key name
* any payload wild card
* byte equal payload (not messing with json in this iteration)

### Not implemented, to discuss
* CLI - the server sides supports one to many grants for a contract but this would be hell for the CLI
* Additional gas costs - beside storage there are no extra gas costs on setup, see: https://github.com/cosmos/cosmos-sdk/issues/8311#issuecomment-1300934288


ethanfrey:
Great stuff. I think these limits and filters are a very good start.
There have been requests for more power, but those shouldn't be needed in v0.30.

Important for me is that the protobuf schema is built to be extensible to possibly add options features later in a backwards compatible way. Not sure if we need some version field or such for this? But then we can keep from feature creep but allow for more power later

alpe:
Thanks for the feedback

@ethanfrey With the interfaces new functionality can be added easily. Can you elaborate on some of the advanced use cases, please?

@kakucodes
> ... if it's possible to make a grant without either a call or coin limit which i think is important functionality to have.

A call limit of 2^64 is almost infinite. I have removed the infinite declarations to reduce complexity. Can you elaborate on your use case, please?

kakucodes:
> Thanks for the feedback
> 
> @ethanfrey With the interfaces new functionality can be added easily. Can you elaborate on some of the advanced use cases, please?
> 
> @kakucodes
> 
> > ... if it's possible to make a grant without either a call or coin limit which i think is important functionality to have.
> 
> A call limit of 2^64 is almost infinite. I have removed the infinite declarations to reduce complexity. Can you elaborate on your use case, please?

This is a totally fair rationale, thank you for the explanation.
Typically for Yieldmos we request a grant with unlimited access to funds and work mainly off the expiration date. I see no reason why this wouldn't be any different now.

ethanfrey:
First off, I think this currently solves the issues as requested in https://github.com/CosmWasm/wasmd/issues/803#issuecomment-1303709343 perfectly. You can provide some account access to call "foo" and "bar" on address X as well as "baz" on address Y, right?

ethanfrey:
> @ethanfrey With the interfaces new functionality can be added easily. Can you elaborate on some of the advanced use cases, please?

My first two thoughts of an advanced use-cases would be:

* funds limits, but using cw20 rather than native (this can actually be handled via allowance already in cw20-base)
* define partial payloads. like `{"vote": {"proposal": 47}}` but not specifying yes or no (to sell one vote). or withdrawing to a fixed recipient but arbitrary amount. This falls between "specifying the method name" and "specifying exact bytes". 

I am not sure these have many real world use cases, and the second one definitely brings up a huge design issue to do safely and clearly and I would not consider this for v0.30 or even v0.31

Anyway, just some food for thought on where this might grow next year and glad there is an extension point for this

proto/cosmwasm/wasm/v1/authz.proto
```diff
@@ -0,0 +1,109 @@
...
+// migration. Since: wasmd 0.30
+message ContractMigrationAuthorization {
+  option (cosmos_proto.implements_interface) = "Authorization";
+
+  // Grants for contract migrations
+  repeated ContractGrant grants = 1 [ (gogoproto.nullable) = false ];
+}
+
+// ContractGrant a granted permission for a single contract
+// Since: wasmd 0.30
```
pinosu, line 31: very good!

tests/e2e/grants_test.go
```diff
@@ -0,0 +1,116 @@
...
+	otherAddr := sdk.AccAddress(otherPrivKey.PubKey().Address().Bytes())
+
+	chain.Fund(granteeAddr, sdk.NewInt(1_000_000))
+	chain.Fund(otherAddr, sdk.NewInt(1_000_000))
+	assert.Equal(t, sdk.NewInt(1_000_000), chain.Balance(granteeAddr, sdk.DefaultBondDenom).Amount)
+
+	myAmount := sdk.NewCoin(sdk.DefaultBondDenom, sdk.NewInt(2_000_000))
+
+	specs := map[string]struct {
+		limit          types.ContractAuthzLimitX
```
pinosu, line 47: nice tests

pinosu (approved):
Excellent work! 🥇

PR #1079, opened 2022-11-07
Authz redesign: CLI implementation

pinosu:
Follow up on #1077


x/wasm/client/cli/tx.go
```diff
@@ -415,18 +423,61 @@ func GrantAuthorizationCmd() *cobra.Command {
...
+					return fmt.Errorf("max funds: %s", err)
+				}
+				limit = types.NewCombinedLimit(maxCalls, maxFunds)
+			case maxFundsStr != "":
+				maxFunds, err := sdk.ParseCoinsNormalized(maxFundsStr)
+				if err != nil {
+					return fmt.Errorf("max funds: %s", err)
+				}
+				limit = types.NewMaxFundsLimit(maxFunds)
+			case maxCalls != 0:
```
alpe, line 441: This option does not allow any token transfers. It would be nice if we can be explicit in the cli like `--max-calls=1 --no-token-transfer`. WDYT?

x/wasm/client/cli/tx.go
```diff
@@ -415,18 +423,61 @@ func GrantAuthorizationCmd() *cobra.Command {
...
+			case maxFundsStr != "" && maxCalls != 0:
+				maxFunds, err := sdk.ParseCoinsNormalized(maxFundsStr)
+				if err != nil {
+					return fmt.Errorf("max funds: %s", err)
+				}
+				limit = types.NewCombinedLimit(maxCalls, maxFunds)
+			case maxFundsStr != "":
+				maxFunds, err := sdk.ParseCoinsNormalized(maxFundsStr)
+				if err != nil {
+					return fmt.Errorf("max funds: %s", err)
```
alpe, line 438: 👍

x/wasm/client/cli/tx.go
```diff
@@ -415,18 +423,61 @@ func GrantAuthorizationCmd() *cobra.Command {
...
+				limit = types.UndefinedLimit{}
+			}
+
+			var filter types.ContractAuthzFilterX
+			switch len(msgs) {
+			case 0:
+				filter = types.NewAllowAllMessagesFilter()
+			default:
+				filter = types.NewAcceptedMessageKeysFilter(msgs...)
+			}
```
alpe, line 453: The third option is to set the concrete json message(s) (byte equal).

x/wasm/client/cli/tx.go
```diff
@@ -415,18 +423,61 @@ func GrantAuthorizationCmd() *cobra.Command {
...
+
+			grant, err := types.NewContractGrant(contract, limit, filter)
+			if err != nil {
+				return err
+			}
+
+			var authorization authz.Authorization
+			switch args[1] {
```
alpe, line 461: 👍 this works

x/wasm/client/cli/tx.go
```diff
@@ -415,18 +423,61 @@ func GrantAuthorizationCmd() *cobra.Command {
...
+			if err != nil {
+				return err
+			}
+			return tx.GenerateOrBroadcastTxCLI(clientCtx, cmd.Flags(), grantMsg)
 		},
 	}
 	flags.AddTxFlagsToCmd(cmd)
 	cmd.Flags().StringSlice(flagAllowedMsgs, []string{}, "Allowed msgs")
-	cmd.Flags().Bool(flagRunOnce, false, "Allow to execute only once")
+	cmd.Flags().Uint64(flagMaxCalls, 0, "Maximal number of calls to the contract")
```
alpe, line 479: `--allow-msgs` is not the best name anymore. Not sure if `--allow-contract-msg-keys` is better but we need to differentiate between the raw json message (byte equal) and the json object keys. CLI doc is important here.
For example:
With a contract payload message: `{"foo":"bar"}`
the key option would be `foo` 
and the raw msg option `{"foo":"bar"}`

tests/e2e/setup.go
```diff
@@ -14,16 +13,16 @@ import (
 // Any method implemented for this struct will be ran when suite.Run() is called.
 type CCVTestSuite struct {
 	suite.Suite
-	coordinator   *ibctesting.Coordinator
-	providerChain *ibctesting.TestChain
-	consumerChain *ibctesting.TestChain
+	// coordinator   *ibctesting.Coordinator
```
alpe, line 16: Sorry, I have removed this file on a local branch already. These changes will cause merge conflicts for you when I push.

alpe (commented):
Great start. The CLI will be hard to come with a good UX. I have added some notes to help. My gut feeling is that more concrete parameter may be better to guide the user than overloading max-calls and amount for the different cases. But other opinions are welcome, too.

x/wasm/client/cli/tx.go
```diff
@@ -415,18 +431,80 @@ func GrantAuthorizationCmd() *cobra.Command {
...
+			case maxCalls != 0 && noTokenTransfer:
+				// TODO set combined limit
+			case maxCalls != 0 && !noTokenTransfer:
+				limit = types.NewMaxCallsLimit(maxCalls)
+			default:
+				limit = types.UndefinedLimit{}
+			}
+
+			var filter types.ContractAuthzFilterX
+			switch {
```
alpe, line 468: Filters are mutual exclusive within one grant. Better fail fast if more than one is set.

alpe (commented):
Good progress. I added some notes to unblock you. I will do a deeper review tomorrow.

alpe:
@Mergifyio rebase

x/wasm/client/cli/tx.go
```diff
@@ -420,18 +440,84 @@ func GrantAuthorizationCmd() *cobra.Command {
...
+			}
+
+			noTokenTransfer, err := cmd.Flags().GetBool(flagNoTokenTransfer)
+			if err != nil {
+				return err
+			}
+
+			var limit types.ContractAuthzLimitX
+			switch {
+			case maxFundsStr != "" && maxCalls != 0 && !noTokenTransfer:
```
alpe, line 456: good check

x/wasm/client/cli/tx.go
```diff
@@ -420,18 +440,84 @@ func GrantAuthorizationCmd() *cobra.Command {
...
+				maxFunds, err := sdk.ParseCoinsNormalized(maxFundsStr)
+				if err != nil {
+					return fmt.Errorf("max funds: %s", err)
+				}
+				limit = types.NewCombinedLimit(maxCalls, maxFunds...)
+			case maxFundsStr != "" && !noTokenTransfer:
```
alpe, line 462: nit: To be on the safe side, I would be very explicit: `case maxFundsStr != "" && maxCalls == 0 && !noTokenTransfer:`

x/wasm/client/cli/tx.go
```diff
@@ -420,18 +440,84 @@ func GrantAuthorizationCmd() *cobra.Command {
...
+				maxFunds, err := sdk.ParseCoinsNormalized(maxFundsStr)
+				if err != nil {
+					return fmt.Errorf("max funds: %s", err)
+				}
+				limit = types.NewMaxFundsLimit(maxFunds...)
+			case maxCalls != 0 && noTokenTransfer:
```
alpe, line 468: to be on the safe side: `maxCalls != 0 && noTokenTransfer && maxFundsStr == ""`

x/wasm/client/cli/tx.go
```diff
@@ -420,18 +440,84 @@ func GrantAuthorizationCmd() *cobra.Command {
...
+				limit = types.NewMaxCallsLimit(maxCalls)
+			default:
+				limit = types.UndefinedLimit{}
```
alpe, line 471: you can fail fast here already
```suggestion
				limit = errors.New("invalid limit setup")
```

x/wasm/client/cli/tx.go
```diff
@@ -420,18 +440,84 @@ func GrantAuthorizationCmd() *cobra.Command {
...
+			}
+
+			var filter types.ContractAuthzFilterX
+			switch {
+			case allowAllMsgs && len(msgKeys) > 0 || allowAllMsgs && len(rawMsgs) > 0 || len(msgKeys) > 0 && len(rawMsgs) > 0:
```
alpe, line 476: 👍 you cover all 3 cases here. I am a bit more used to read `len(x) != 0` instead but all correct

x/wasm/client/cli/tx.go
```diff
@@ -420,18 +440,84 @@ func GrantAuthorizationCmd() *cobra.Command {
...
+				return fmt.Errorf("cannot set more than one filter within one grant")
```
alpe, line 477: nit
```suggestion
				return errors.New("cannot set more than one filter within one grant")
```

x/wasm/client/cli/tx.go
```diff
@@ -420,18 +440,84 @@ func GrantAuthorizationCmd() *cobra.Command {
...
+			case len(msgKeys) > 0:
+				filter = types.NewAcceptedMessageKeysFilter(msgKeys...)
+			case len(rawMsgs) > 0:
+				msgs := make([]types.RawContractMessage, len(rawMsgs))
+				for i, msg := range rawMsgs {
+					msgs[i] = types.RawContractMessage(msg)
+				}
+				filter = types.NewAcceptedMessagesFilter(msgs...)
+			default:
+				filter = &types.UndefinedFilter{}
```
alpe, line 489: nit: as above, just return an error here

x/wasm/client/cli/tx.go
```diff
@@ -384,9 +390,13 @@ func parseExecuteArgs(contractAddr string, execMsg string, sender sdk.AccAddress
...
 func GrantAuthorizationCmd() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "grant [grantee] [contract_addr_bech32] --allow-msgs [msg1,msg2,...]",
+		Use:   "grant [grantee] [message_type=\"execution\"|\"migration\"] [contract_addr_bech32] --allow-raw-msgs [msg1,msg2,...] --allow-msg-keys [key1,key2,...] --allow-all-messages",
 		Short: "Grant authorization to an address",
-		Args:  cobra.ExactArgs(2),
+		Long: fmt.Sprintf(`Grant authorization to an address.
+Examples:
+$ %s tx grant <grantee_addr> execution <contract_addr> --allow-all-messages --maxCalls 1 --no-token-transfer --expiration 1667979596
+`, version.AppName),
```
alpe, line 398: Would it make sense to have some more concrete examples for the limits and filters? I think you can cover all in 3 examples

alpe (approved):
Good work! I added some nits and personal preferences. But nothing blocking!

PR #1076, opened 2022-11-04
Change genesis preserving contract history

GNaD13:
Close #878

PR #1089, opened 2022-11-11
Return contract history updates on query

alpe:
Follow up on https://github.com/CosmWasm/wasmd/pull/1076#pullrequestreview-1177018775


pinosu (approved):
LGTM 👍

PR #1095, opened 2022-11-17
Enable larger wasm bytecode upload for gov proposals

### LINKED ISSUE #1038: Allow gov votes to upload larger wasm bytecode
We currently limit the [max size for wasm to 800 KB]( https://github.com/CosmWasm/wasmd/blob/a76999a1d7c3ed3570a5f319940667a5e5b61eaf/x/wasm/types/validation.go#L15)

This is very reasonable for most contracts, but I have passed this limit when trying to embed an interpreter in a cosmwasm contract. This is a pretty unusual case, but could open interesting possibilities. wasm sizes when compiling some repos I found are around 2MB however.

A chain could raise global limit to 3MB, but that is not always desirable. To enable such use case without opening other issues, I was thinking we could use a different limit for gov contracts. Say a gov vote could upload wasm up to 4MB, while normal tx only the 800 KB limit.

Open for other approaches, but I think it would be nice to allow some way to bypass this limit when governance agrees.

@alpe I would love this on 0.30 backlog if possible, as I think it is a simple code change once we agree on an approach.

pinosu:
Resolves #1038

### DIFF
--- a/x/wasm/types/genesis.go
+++ b/x/wasm/types/genesis.go
@@ -47,7 +47,7 @@ func (c Code) ValidateBasic() error {
 	if err := c.CodeInfo.ValidateBasic(); err != nil {
 		return sdkerrors.Wrap(err, "code info")
 	}
-	if err := validateWasmCode(c.CodeBytes, MaxWasmSize); err != nil {
+	if err := validateWasmCode(c.CodeBytes, MaxProposalWasmSize); err != nil {
 		return sdkerrors.Wrap(err, "code bytes")
 	}
 	return nil
--- a/x/wasm/types/genesis.go
+++ b/x/wasm/types/genesis.go
@@ -47,7 +47,7 @@ func (c Code) ValidateBasic() error {
 	if err := c.CodeInfo.ValidateBasic(); err != nil {
 		return sdkerrors.Wrap(err, "code info")
 	}
-	if err := validateWasmCode(c.CodeBytes); err != nil {
+	if err := validateWasmCode(c.CodeBytes, MaxWasmSize); err != nil {
 		return sdkerrors.Wrap(err, "code bytes")
 	}
 	return nil
--- a/x/wasm/types/validation.go
+++ b/x/wasm/types/validation.go
@@ -13,14 +13,17 @@ var (
 
 	// MaxWasmSize is the largest a compiled contract code can be when storing code on chain
 	MaxWasmSize = 800 * 1024 // extension point for chains to customize via compile flag.
+
+	// MaxProposalWasmSize is the largest a gov proposal compiled contract code can be when storing code on chain
+	MaxProposalWasmSize = 4 * 1024 * 1024 // extension point for chains to customize via compile flag.
 )
 
-func validateWasmCode(s []byte) error {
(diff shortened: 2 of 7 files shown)


x/wasm/types/validation.go
```diff
@@ -13,14 +13,17 @@ var (
 
 	// MaxWasmSize is the largest a compiled contract code can be when storing code on chain
 	MaxWasmSize = 800 * 1024 // extension point for chains to customize via compile flag.
+
+	// MaxProposalWasmSize is the largest a gov proposal compiled contract code can be when storing code on chain
+	MaxProposalWasmSize = 4 * 1024 * 1024 // extension point for chains to customize via compile flag.
 )
 
-func validateWasmCode(s []byte) error {
+func validateWasmCode(s []byte, maxSize int) error {
```
alpe, line 21: good adoption of function!

x/wasm/types/validation.go
```diff
@@ -13,14 +13,17 @@ var (
 
 	// MaxWasmSize is the largest a compiled contract code can be when storing code on chain
 	MaxWasmSize = 800 * 1024 // extension point for chains to customize via compile flag.
+
+	// MaxProposalWasmSize is the largest a gov proposal compiled contract code can be when storing code on chain
+	MaxProposalWasmSize = 4 * 1024 * 1024 // extension point for chains to customize via compile flag.
```
alpe, line 18: Good to have this as a variable, too. I am not fully convinced that 4mb is a good value, given the spam case. 2mb feels a bit more conservative but  it is also a random value

x/wasm/types/genesis.go
```diff
@@ -47,7 +47,7 @@ func (c Code) ValidateBasic() error {
 	if err := c.CodeInfo.ValidateBasic(); err != nil {
 		return sdkerrors.Wrap(err, "code info")
 	}
-	if err := validateWasmCode(c.CodeBytes); err != nil {
+	if err := validateWasmCode(c.CodeBytes, MaxWasmSize); err != nil {
```
alpe, line 50: This needs to be raised or an import fails for gov contracts
```suggestion
	if err := validateWasmCode(c.CodeBytes, MaxProposalWasmSize); err != nil {
```

alpe (commented):
Very nice! 
One minor issue and some feedback on the max value for gov content would be great.

x/wasm/types/validation.go
ethanfrey: gov proposals charge like 3x gas for the proposal size (than just raw tx size), but yes it is written to state

It is easy to get 1.5MB contracts, maybe 2.5MB or 3MB if you want to be a bit conservative but still allow more complexity (like a js interpreter)

Also, we could just charge extra gas for every byte above the base (800kb) level up to this hard cap, but that might be too confusing/complex

ethanfrey: If chains can override, we just need a reasonable default. I am fine with 3MB. Feel 2MB is a bit tight (I got a demo contract I could only trim to 1.8MB and could easily grow a bit if I added any features)

alpe (approved):
👍 thanks for the updates 🏄

ethanfrey:
Great. Thank you for this!

PR #1088, opened 2022-11-10
Upgrade to IBC v4.2.0 with interchain-accounts v0.2.4

alpe:
Resolves #960
Upgrade to [ibc-go v4.2.0](https://github.com/cosmos/ibc-go/releases/tag/v4.2.0)

Includes:
* IBC protocol handshake
* [interchain-accounts v0.2.4](https://github.com/cosmos/interchain-accounts-demo/releases/tag/v0.2.4)
* Updates to the ibctesting package coming from the ibc-go repo. Credits and big thank you to the original authors

You can read more about ibc fees middleware included in v4 in this [article](https://medium.com/the-interchain-foundation/ibc-relaying-as-a-service-the-in-protocol-incentivization-story-9922c7b953f0)

### Notes for users
In order to submit a relayer fee for an IBC packet, a `MsgPayPacketFee` needs to be in the TX before the packet creating msg. 

* native ics20 transfer:

```go
transferCoin := sdk.NewCoin(sdk.DefaultBondDenom, sdk.NewInt(1))
ibcPayloadMsg := ibctransfertypes.NewMsgTransfer(path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID, transferCoin, actorChainA.String(), receiver.String(), clienttypes.Height{}, uint64(time.Now().Add(time.Minute).UnixNano()))
ibcPackageFee := ibcfee.NewFee(oneToken, oneToken, sdk.Coins{})
feeMsg := ibcfee.NewMsgPayPacketFee(ibcPackageFee, ibctransfertypes.PortID, path.EndpointA.ChannelID, actorChainA.String(), nil)

_, err = chainA.SendMsgs(feeMsg, ibcPayloadMsg)
```
* cw20-ics transfer

```go
execMsg := wasmtypes.MsgExecuteContract{
	Sender:   actorChainA.String(),
	Contract: cw20ContractAddr.String(),
	Msg:      exec,
}
ibcPackageFee := ibcfee.NewFee(oneToken, oneToken, sdk.Coins{})
feeMsg := ibcfee.NewMsgPayPacketFee(ibcPackageFee, ibcContractPortID, path.EndpointA.ChannelID, actorChainA.String(), nil)

_, err = chainA.SendMsgs(feeMsg, &execMsg)
```

PR #1100, opened 2022-11-18
Add extension point for ibc app version decoding

alpe:
After #1088 

The version string stored with the channel can be overloaded with ibc middleware data. This PR adds an extension point to add a custom decoder to return the original IBC app level version.
* ICS29AppVersionDecoder - supports the fee middleware
* AppVersionDecoderChain - util to chain decoders as you can do with middleware and cascade output

### DIFF
--- a/x/wasm/ibc_middleware_support.go
+++ b/x/wasm/ibc_middleware_support.go
@@ -1,0 +1,68 @@
+package wasm
+
+import (
+	"encoding/json"
+	"strings"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	ibcfees "github.com/cosmos/ibc-go/v4/modules/apps/29-fee/types"
+	channeltypes "github.com/cosmos/ibc-go/v4/modules/core/04-channel/types"
+
+	"github.com/CosmWasm/wasmd/x/wasm/types"
+)
+
+var _ IBCAppVersionDecoder = AppVersionDecoderFn(nil)
+
+// AppVersionDecoderFn custom type that implements IBCAppVersionDecoder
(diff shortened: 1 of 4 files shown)


ethanfrey (commented):
I am confused here, and we should check with the ibc-go team.

When I discussed middleware with them this summer,  understood it to be transparent. So if I have middleware(A) talking to middleware(B), neither A or B is aware they are wrapped by the middleware. The middleware sees the whole packet and version and after stripping off it's envelope, passes the content to the wrapped protocol.

For me, this is the only way that it can easily be adopted without forking all protocols that may be wrapped. Can @AdityaSripal comment on this issue?

alpe:
The issue is that we not always have the full channel data passed via the IBC module but load the channel info from the store. There the raw version is stored.

x/wasm/ibc_middleware_support.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+// ICS29AppVersionDecoder decodes the ibc app version from an ics-29 fee middleware version when supported
+// fallback is raw version when it can not be determined
+func ICS29AppVersionDecoder(channelSource interface {
+	IsFeeEnabled(ctx sdk.Context, portID, channelID string) bool
+},
+) IBCAppVersionDecoder {
+	return AppVersionDecoderFn(func(ctx sdk.Context, rawVersion, portID, channelID string) (string, error) {
+		if channelSource.IsFeeEnabled(ctx, portID, channelID) {
```
ethanfrey, line 50: If needed to get it to work, it is a hack and shows middleware stack is not complete.

There may be many new middleware and the wrapped applications shouldn't need to care, so no need to add eg ICS49 middleware that comes later

alpe:
Obsolete by https://github.com/CosmWasm/wasmd/pull/1102

PR #1102, opened 2022-11-21
Get ibc app version from middleware

alpe:
After #1088
This looks better now

### DIFF
--- a/x/wasm/ibc.go
+++ b/x/wasm/ibc.go
@@ -2,13 +2,11 @@ package wasm
 
 import (
 	"math"
-	"strings"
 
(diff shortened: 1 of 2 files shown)


x/wasm/ibc.go
```diff
@@ -19,13 +17,20 @@ import (
 
 var _ porttypes.IBCModule = IBCHandler{}
 
+// internal interface that is implemented by ibc middleware
```
ethanfrey, line 20: I assume this works regardless of the stack? (eg no middleware or with middleware)

This looks much cleaner than the previous approach.

PR #1103, opened 2022-11-21
Enable key name usage for admin

### LINKED ISSUE #1039: CLI: Allow using key name for --admin
The CLI field `--from string              Name or address of private key with which to sign` allows me to use the local key alias. Now when I instantiate a contract with myself as the admin, I would like to do `--from my_super_key --admin my_super_key` without having to look up the address manually to avoid the risk of using the wrong one. However, when I try to do that I get

> Error: admin: decoding bech32 failed: invalid separator index -1

Would that be possible to implement? Does it make sense?

pinosu:
Resolves #1039


alpe (approved):
LGTM 🌷

webmaster128:
❤️

PR #1110, opened 2022-11-28
Test wasmd.exe in CI

webmaster128:


alpe:
This PR is blocked. See https://github.com/CosmWasm/wasmvm/issues/389

PR #1111, opened 2022-11-28
Run unit tests on windows

webmaster128:
This is a follow-up of #1110 (merge that first, then rebase this PR). The main point here is that we are running wasmd unit-tests on Windows. This works more or less. But a lot of tests are failing.

A lot of failures look something like this:

>     testing.go:1097: TempDir RemoveAll cleanup: remove C:\Users\CIRCLE~1.PAC\AppData\Local\Temp\TestPinPong2683230876\001\data\snapshots\metadata.db\000001.log: The process cannot access the file because it is being used by another process.

I don't know how this happens. Maybe the app needs to be shut down before `nodeHome := t.TempDir()` is deleted or something like this.

Any help would be appreciated.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -121,7 +121,19 @@ jobs:
       - run:
           name: Build
           command: |
+            set -o errexit
             LEDGER_ENABLED=false make build
+      - run:
+          name: Check libwasmvm-version
+          command: |
+            set -o errexit
+            ./build/wasmd version
+            ./build/wasmd query wasm libwasmvm-version
+      - run:
+          name: Test
+          command: |
+            set -o errexit
+            make test
 
   benchmark:
     executor: golang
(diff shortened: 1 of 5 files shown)


alpe:
This looks familiar. The DB on the IBC tests may need to be closed. I will 👁️ at this when I get a chance

webmaster128:
> This was intended on a new branch but I misstyped the push. Sorry for this

No worries. Happy to share this one to fix wasmd unit tests on Windows. We'll need to clean up here anyways after #1110 is in.

webmaster128:
Thank you @alpe, that's very helpful. Don't worry about the remaining issues, especially the crashes. There is something fishy going on in wasmvm. I'm on it.

alpe:
This PR is blocked on https://github.com/CosmWasm/wasmvm/issues/389

PR #1112, opened 2022-11-30
fix: remove validator set update

swelf19:
removed updatesValidatorSet from initGenesis to make wasmd module interchain security consumer compatible


swelf19:
wrong target

PR #1117, opened 2022-12-02
fix: sort coins in `ConvertWasmCoinsToSdkCoins`

apollo-sturdy:
When sending e.g. `BankMsg::Send` from a contract, it appears that currently the coins in `amount` parameter need to be sorted. This is inconvenient as contracts need to add additional logic to sort coins before sending them. This PR simply sorts them instead in wasmd when converting them into SDK Coins.


alpe (approved):
Very useful. Thanks! 🥇

PR #1125, opened 2022-12-13
Address cli behaviour on tx and query errors

### LINKED ISSUE #926: Is the difference in codespace between execute and query intended behavior?
I would like to check something about the behavior of **codespace** when an error occurs.

The following contract with errors was created, and the behavior of **execute** and **query** was investigated.
https://github.com/da1suk8/cosmwasm_cosmwasm/blob/error_behavior/contracts/queue/src/contract.rs

1. When an error occurs, in execute, the codespace is `wasm`, and in query, the codespace is `std`. Is this **intentional behavior**?
2. When an error occurs in query, usage is displayed, is this **correct behavior**?
please see below for log
```
$wasmd q wasm contract-state smart wasm1nc5tatafv6eyq7llkr2gv50ff9e22mnf70qgjlv737ktmt4eswrqr5j2ht '{"count":{}}'
Error: rpc error: code = InvalidArgument desc = Error calling the VM: Error executing Wasm: Wasmer runtime error: RuntimeError: Aborted: panicked at 'explicit panic', src/contract.rs:110:9: query wasm contract failed: invalid request
Usage:
wasmd query wasm contract-state smart [bech32_address] [query] [flags]
 
Flags:
--ascii ascii encoded query argument
--b64 base64 encoded query argument
--height int Use a specific height to query state at (this can error if the node is pruning state)
-h, --help help for smart
--hex hex encoded query argument
--node string <host>:<port> to Tendermint RPC interface for this chain (default "tcp://localhost:26657")
-o, --output string Output format (text|json) (default "text")
 
Global Flags:
--chain-id string The network chain ID
--home string directory for config and data (default "/Users/jp30571/.wasmd")
--log_format string The logging format (json|plain) (default "plain")
--log_level string The logging level (trace|debug|info|warn|error|fatal|panic) (default "info")
--trace print out full stack trace on errors
```

pinosu:
Closes #926 

- Removed old script


alpe (approved):
Nice 👍

PR #1130, opened 2022-12-17
lint tests

faddat:
The cosmos-sdk, ibc-go and gaia all fail to lint tests, leading to 
inconsistencies.  This PR addresses that.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -22,7 +22,6 @@ linters:
     - prealloc
     - exportloopref
     - staticcheck
-    - structcheck
     - stylecheck
     - typecheck
     - unconvert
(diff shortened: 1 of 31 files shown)


alpe:
Thank you for the PR! This was probably a lot of work. I would like to get some more feedback if we really want to enforce linter constraints on tests. This was sometimes a heated debate on previous projects.

faddat:
Opinion: it's important, and results in better tests.

Best example for testing is errcheck. 

While we test, we often don't check errors, and in the long run this could lead to faulty tests.

alpe:
There are already a lot of merge conflicts with #1149 . I appreciate your effort and energy but in order to reduce the workload while maintaining 2 active development branches, I close this now.
Please note, there was no resistance to have tests coded covered, so we can try again after  #1149 is merged to main.

PR #1131, opened 2022-12-18
add proposal annotations

pyramation:
add `implements_interface` annotations which are useful for new telescope encoders/decoders/amino. 

example: https://gist.github.com/pyramation/217af77396606ddd6d3c7873fb85e9c7


alpe (approved):
Thanks for the PR! This is useful!

PR #1135, opened 2022-12-21
Bump golangci-lint

faddat:
This PR bumps golangci-lint to the latest version, making it possible to use go 1.19 more easily.


alpe:
Thank you 🌻

PR #1159, opened 2023-01-17
Remove intertx for vanilla ICA

alpe:
Integrate Interchain Accounts from ibc-go module
- contains sdk version bump
- removes intertx module


pinosu (approved):
LGTM 👍

PR #1164, opened 2023-01-19
Better to sdk coin convertion

### LINKED ISSUE #1118: Wasmd does not sort coins when converting from CosmWasm Coins to SDK  Coins
Cosmos SDK expects its `Coins` object to be sorted alphabetically by denom, or else it will throw an `invalid coins` error.
In function [ConvertWasmCoinsToSdkCoins](https://github.com/CosmWasm/wasmd/blob/main/x/wasm/keeper/handler_plugin_encoders.go#L325) when converting from CosmWasm coins to SDK coins, they are not sorted, which means that they are also not sorted when encoding `CosmosMsg`s that come from `Response`s returned from contract executions. This means that all contracts must ensure to sort coins before issuing responses with `CosmosMsg`s that contain native coins. This is obviously not ideal and leads to bugs when contract authors forget to sort the coins.

My proposed solution is to simply sort the coins when converting them into SDK coins. I have implemented a fix here: https://github.com/CosmWasm/wasmd/pull/1117

alpe:
Resolves #1118 

* drops empty coins
* merges duplicate denoms

### DIFF
--- a/x/wasm/keeper/handler_plugin_encoders_test.go
+++ b/x/wasm/keeper/handler_plugin_encoders_test.go
@@ -635,3 +635,78 @@ func TestConvertWasmCoinToSdkCoin(t *testing.T) {
 		})
 	}
 }
+
+func TestConvertWasmCoinsToSdkCoins(t *testing.T) {
+	specs := map[string]struct {
+		src    []wasmvmtypes.Coin
+		exp    sdk.Coins
+		expErr bool
+	}{
+		"empty": {
+			src: []wasmvmtypes.Coin{},
+			exp: nil,
+		},
+		"single coin": {
+			src: []wasmvmtypes.Coin{{Denom: "foo", Amount: "1"}},
+			exp: sdk.NewCoins(sdk.NewCoin("foo", sdk.NewInt(1))),
+		},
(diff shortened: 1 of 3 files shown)


x/wasm/keeper/handler_plugin_encoders_test.go
```diff
@@ -635,3 +635,78 @@ func TestConvertWasmCoinToSdkCoin(t *testing.T) {
...
+				sdk.NewCoin("foo", sdk.NewInt(1)),
+			),
+		},
+		"duplicate denoms merged": {
+			src: []wasmvmtypes.Coin{
+				{Denom: "foo", Amount: "1"},
+				{Denom: "foo", Amount: "1"},
+			},
+			exp: []sdk.Coin{sdk.NewCoin("foo", sdk.NewInt(2))},
+		},
```
webmaster128, line 690: Can we add a test for merging zero amounts? I.e. 2foo + 0foo = 2foo

webmaster128 (commented):
Very nice. Just one suggested test to ensure Coins.Add does not do something strange.

webmaster128 (approved):

PR #1162, opened 2023-01-19
Build without cgo 

alpe:
Resolves #1129

* Cherry pick https://github.com/public-awesome/wasmd/commit/cffe53fa0eeee0a9735ac2dd6d419d9d1d25eccf 
* Minor updates, formatting

Credits and ❤️ go to the original author @jhernandezb 👏 
[Apache license](https://github.com/public-awesome/wasmd/blob/main/LICENSE)

PR #1166, opened 2023-01-21
add non cgo build

jhernandezb:
This is a replacement for https://github.com/CosmWasm/wasmd/pull/1162 which now only requires a change in the keeper. 

The original POC required the cli to be modified because `wasmvm.LibwasmvmVersion()` wasn't handling the cgo,no_cgo but now it does by returning an error.


alpe (approved):
Replaces #1162 
Nice, thank you!

PR #1167, opened 2023-01-22
Fix msg format in EVENTS.md

llllllluc:
Looks like the format of event log on contract address is outdated in `EVENTS.md`.
Latest format is `_contract_address` instead of `_contract_addr`.


alpe (approved):
Good 👁️ ! Thanks for fixing this 👍

Issue #1198, opened 2023-02-02
🚨: Public Goods work halted


alpe:
Please see https://twitter.com/CosmWasm/status/1620783214125256704 for the announcement and discussion.

alpe:
Thank you all for your support!
https://twitter.com/confio_tech/status/1630965447419133954

PR #1201, opened 2023-02-13
list-contract-by-code bugfix

nik-suri:
Adds Pagination Flags to the `list-contract-by-code` wasm query command.

Without the pagination flags, the query panics with
```
panic: flag accessed but not defined: page-key
```


alpe:
Thank you! See #1198

alpe (approved):
Very nice! Thank you 🌷

PR #1208, opened 2023-02-23
fix: stargate querier does not reset the structure state

dadamu:
Currently, the stargate querier marshal the result to the given structure stored in accepted list, but it does not reset the state. So the result will be unexpected. For example, the current stargate querier process is like:

```go
cdc, _ := app.MakeCodecs()

protoResponse := &banktypes.QueryAllBalancesResponse{}

bz, err := cdc.Marshal(&banktypes.QueryAllBalancesResponse{
Balances: sdk.NewCoins(sdk.NewInt64Coin("test", 1)),
})

// First stargate query to QueryAllBalances 
err = cdc.Unmarshal(bz, protoResponse)
if err != nil {
panic(err)
}

// Second stargate query to QueryAllBalances 
err = cdc.Unmarshal(bz, protoResponse)
if err != nil {
panic(err)
}

fmt.Println(protoResponse.String())
// balances:<denom:"test" amount:"1" > balances:<denom:"test" amount:"1" >
// but the expected result is: balances:<denom:"test" amount:"1" >>
```

As a result, we should reset the protoResponse after marshaling.

### DIFF
--- a/x/wasm/keeper/query_plugins.go
+++ b/x/wasm/keeper/query_plugins.go
@@ -600,6 +600,7 @@ func ConvertProtoToJSONMarshal(cdc codec.Codec, protoResponse codec.ProtoMarshal
 		return nil, sdkerrors.Wrap(err, "to json")
 	}
 
+	protoResponse.Reset()
 	return bz, nil
 }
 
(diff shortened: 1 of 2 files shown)


alpe:
Thanks for the PR and this important fix! I hope that we can resume OS work very soon. Please see #1198 for context

alpe:
Would you mind to convert your description into a unit test for regression so that we also have confidence in the future?

dadamu:
@alpe Added.

faddat:
Thanks!

alpe:
Great work! Thanks a lot for adding the test, too. Important work!🥇

PR #1213, opened 2023-02-28
comet

faddat:
This PR introduces comet bft, to replace tendermint.


alpe:
Thanks for the PR! 
With the changes in sdk release/v0.45.x to CometBFT, we need to fix dependencies in wasmvm before. A simple replace statement in go.mod does not work anymore.
See #1217 for the linked issue.

PR #1225, opened 2023-03-02
Return IBC packet sequence number

### LINKED ISSUE #1154: Return IBC packet sequence number in the handler_plugin
For certain use cases, such as handling packets to ICQ and ICA hosts, the smart contract needs to connect outgoing packets to incoming acknowledgements to be able to handle the internal serialized bytes of the acknowledgement. To do this, the  `IBCRawPacketHandler` `DispatchMsg` method should return the sequence number in some format. This could be a protobuf, for example 
```
message IbcSendPacketResponse {
  // sequence number of the transfer packet sent
  uint64 sequence = 1;
}
```
for an example of needed changes; see [here](https://github.com/quasar-finance/wasmd/commit/92009f61a2e3a7e5e54cac8147239c0e94279bac)

pinosu:
Resolves #1154


proto/cosmwasm/wasm/v1/ibc.proto
```diff
@@ -29,3 +29,9 @@ message MsgIBCSend {
 message MsgIBCCloseChannel {
   string channel = 2 [ (gogoproto.moretags) = "yaml:\"source_channel\"" ];
 }
+
+// MsgIBCSendResponse
+message MsgIBCSendResponse {
+  // sequence number of the transfer packet sent
+  uint64 sequence = 1;
```
alpe, line 36: 👍

x/wasm/keeper/handler_plugin.go
```diff
@@ -188,7 +188,19 @@ func (h IBCRawPacketHandler) DispatchMsg(ctx sdk.Context, _ sdk.AccAddress, cont
...
 		msg.IBC.SendPacket.Timeout.Timestamp,
 	)
-	return nil, nil, h.channelKeeper.SendPacket(ctx, channelCap, packet)
+
+	if err := h.channelKeeper.SendPacket(ctx, channelCap, packet); err != nil {
+		return nil, nil, sdkerrors.Wrap(err, "failed to send packet")
+	}
+
+	resp := &types.MsgIBCSendResponse{Sequence: sequence}
+	val, err := resp.Marshal()
```
alpe, line 197: This does work correct! 
I had expected to see the `codec.Codec` passed to the `IBCRawPacketHandler` object as an abstraction layer but this is fine.

x/wasm/keeper/handler_plugin.go
```diff
@@ -188,7 +188,19 @@ func (h IBCRawPacketHandler) DispatchMsg(ctx sdk.Context, _ sdk.AccAddress, cont
...
+	if err != nil {
+		return nil, nil, sdkerrors.Wrap(err, "failed to marshal sequence response")
+	}
+
+	data = append(data, val)
```
alpe, line 202: No need to append. Just set the value here
```suggestion
	data = val
```
or better
```go 
return nil, val, nil
```

x/wasm/keeper/handler_plugin_test.go
```diff
@@ -308,14 +308,20 @@ func TestIBCRawPacketHandler(t *testing.T) {
...
 			if spec.expErr != nil {
 				return
 			}
-			assert.Nil(t, data)
+
 			assert.Nil(t, evts)
+			require.NotNil(t, data)
+
+			msg := &types.MsgIBCSendResponse{Sequence: 1}
+			val, err := msg.Marshal()
```
alpe, line 322: This is the same code as used in the handler. When you `unmarshal` the returned data instead and compare the seq, you are acting as the client. This gives confidence that both sides work.

alpe (approved):
Great work. Just some minor nits. 
Please 👁️ at them before merging.

x/wasm/keeper/handler_plugin_test.go
```diff
@@ -308,14 +308,23 @@ func TestIBCRawPacketHandler(t *testing.T) {
...
 				return
 			}
-			assert.Nil(t, data)
+
 			assert.Nil(t, evts)
+			require.NotNil(t, data)
+
+			expMsg := types.MsgIBCSendResponse{Sequence: 1}
+
+			actualMsg := types.MsgIBCSendResponse{}
```
ethanfrey, line 323: Nice test case

x/wasm/keeper/handler_plugin_test.go
```diff
@@ -308,14 +308,23 @@ func TestIBCRawPacketHandler(t *testing.T) {
...
-			data, evts, gotErr := h.DispatchMsg(ctx, RandomAccountAddress(t), ibcPort, wasmvmtypes.CosmosMsg{IBC: &wasmvmtypes.IBCMsg{SendPacket: &spec.srcMsg}})
+			evts, data, gotErr := h.DispatchMsg(ctx, RandomAccountAddress(t), ibcPort, wasmvmtypes.CosmosMsg{IBC: &wasmvmtypes.IBCMsg{SendPacket: &spec.srcMsg}})
 			// then
 			require.True(t, spec.expErr.Is(gotErr), "exp %v but got %#+v", spec.expErr, gotErr)
 			if spec.expErr != nil {
 				return
 			}
-			assert.Nil(t, data)
+
 			assert.Nil(t, evts)
```
ethanfrey, line 318: Really? The events are nil? How do we get the ibc packet events to the relayers here?
(I guess they just pass through context and not via the return value?)

ethanfrey (approved):
Looks good.
I'd love to see a full-stack test with contracts, but not sure if that is possible.

(We do have some ibc contract tests I believe, maybe some check there in a follow up PR)

x/wasm/keeper/handler_plugin_test.go
alpe: yes, via context.

proto/cosmwasm/wasm/v1/ibc.proto
```diff
@@ -29,3 +29,9 @@ message MsgIBCSend {
 message MsgIBCCloseChannel {
   string channel = 2 [ (gogoproto.moretags) = "yaml:\"source_channel\"" ];
 }
+
+// MsgIBCSendResponse
+message MsgIBCSendResponse {
+  // Sequence number of the IBC packet sent
+  uint64 sequence = 1;
+}
```
webmaster128, line 37: `proto/cosmwasm/wasm/v1/tx.proto` is not super consistent but usually we have the response type right after the message type. I think that's helpful for navigating. I'd move it above `message MsgIBCCloseChannel {`.

x/wasm/keeper/handler_plugin.go
```diff
@@ -188,7 +188,18 @@ func (h IBCRawPacketHandler) DispatchMsg(ctx sdk.Context, _ sdk.AccAddress, cont
...
-	return nil, nil, h.channelKeeper.SendPacket(ctx, channelCap, packet)
+
+	if err := h.channelKeeper.SendPacket(ctx, channelCap, packet); err != nil {
+		return nil, nil, sdkerrors.Wrap(err, "failed to send packet")
+	}
+
+	resp := &types.MsgIBCSendResponse{Sequence: sequence}
+	val, err := resp.Marshal()
+	if err != nil {
+		return nil, nil, sdkerrors.Wrap(err, "failed to marshal sequence response")
```
webmaster128, line 199:
```suggestion
		return nil, nil, sdkerrors.Wrap(err, "failed to marshal IBC send response")
```

webmaster128 (commented):
Very nice, two small nits

PR #1226, opened 2023-03-03
Separate server and windows client

alpe:
Some personal preferences for better dev UX
* rename build target and artifact to be more precise what is built
* add error message to default build target

Untested


pinosu (approved):
I tested it locally and it works fine. Nice work! 👍

PR #1236, opened 2023-03-06
Test rust panic for regression

alpe:
Regression test for panic in Rust


x/wasm/keeper/reflect_test.go
```diff
@@ -542,6 +542,29 @@ func TestWasmRawQueryWithNil(t *testing.T) {
 	require.Equal(t, []byte{}, reflectRawRes.Data)
 }
 
+func TestRustPanicIsHandled(t *testing.T) {
```
pinosu, line 545: Nice tests!

PR #1247, opened 2023-03-08
Support msg update params gov proposal

pinosu:
Replace v1beta parameter update model with gov v1 `MsgUpdateParams`


x/wasm/keeper/migrations.go
```diff
@@ -25,3 +27,27 @@ func (m Migrator) Migrate1to2(ctx sdk.Context) error {
 	})
 	return nil
 }
+
+// Migrate2to3 migrates the x/wasm module state from the consensus
+// version 2 to version 3. Specifically, it takes the parameters that are currently stored
+// and managed by the x/params module and stores them directly into the x/wasm
+// module state.
+func (m Migrator) Migrate2to3(ctx sdk.Context) error {
```
alpe, line 35: Good 👁️ 
Should we move migrations to a new package as the SDK [did](https://github.com/cosmos/cosmos-sdk/tree/release/v0.47.x/x/staking/migrations)?

alpe (commented):
Good start! 👍

proto/cosmwasm/wasm/v1/tx.proto
```diff
@@ -32,6 +33,11 @@ service Msg {
   // UpdateInstantiateConfig updates instantiate config for a smart contract
   rpc UpdateInstantiateConfig(MsgUpdateInstantiateConfig)
       returns (MsgUpdateInstantiateConfigResponse);
+  // UpdateParams defines a governance operation for updating the x/wasm
+  // module parameters. The authority is defined in the keeper.
+  //
+  // Since: cosmos-sdk 0.47
```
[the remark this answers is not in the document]
ethanfrey: How about starying wasmd 0.40 with sdk 0.47 and comet and all.
Nice number and leaves plenty of space for backfilling any 0.3x.y released on cosmos-sdk 0.45 (depending on how it takes people to merge)

x/wasm/module.go
```diff
@@ -209,6 +209,8 @@ func (AppModule) GenerateGenesisState(simState *module.SimulationState) {
 }
 
 // ProposalContents doesn't return any content functions for governance proposals.
+//
+//nolint:staticcheck
```
alpe, line 204: No need to deactivate the linter. The message was renamed
```go
// ProposalMsgs returns msgs used for governance proposals for simulations.
func (AppModule) ProposalMsgs(simState module.SimulationState) []simtypes.WeightedProposalMsg {

```

x/wasm/simulation/proposals.go
```diff
@@ -146,6 +149,8 @@ func SimulateStoreCodeProposal(wasmKeeper WasmKeeper) simtypes.ContentSimulatorF
 }
 
 // Simulate instantiate contract proposal
+//
+//nolint:staticcheck
```
alpe, line 153: This and others: no need to deactivate the linter. 
This function should return `sdk.Msg` instead. This is how [bank](https://github.com/cosmos/cosmos-sdk/blob/release/v0.47.x/x/bank/simulation/proposals.go) looks like for example

alpe: Looks like we need to support old and new proposal types for some time. There will be a new PR to add new proposal types to  sim

alpe (approved):

PR #1248, opened 2023-03-09
Make `CaptureIbcEvents` in ibctesting public

0xekez:
Before this change, it wasn't possible to implement the `chain.SendMsgs` method without [copying](https://github.com/public-awesome/ics721/blob/main/e2e/suite_helpers.go#L81-L98) the method and its dependents over to your testing code.


alpe (approved):
Thank you for your contribution! 💐

PR #1261, opened 2023-03-15
linting 47 pr

faddat:
This PR lints the sdk 47 work, including tests.  

It also includes #1259 since I think that'll merge into develop_sdk47 first.

PR #1265, opened 2023-03-15
Minor chore

alpe:
After #1261
* Regenerate from proto
* Remove dead code
* Format and polish code

/cc @faddat


x/wasm/alias.go
```diff
@@ -1,4 +1,3 @@
-//nolint
```
alpe, line 1: The linter and formatter where fighting on this line

faddat: it would hate that line, great catch!

pinosu (approved):
LGTM 👍

PR #1269, opened 2023-03-16
[WIP] Deprecate proposal types

pinosu:


x/wasm/simulation/proposals.go
```diff
@@ -78,81 +74,63 @@ func ProposalContents(bk BankKeeper, wasmKeeper WasmKeeper) []simtypes.WeightedP
...
 }
 
 // simulate store code proposal (unused now)
 // Current problem: out of gas (defaul gaswanted config of gov SimulateMsgSubmitProposal is 10_000_000)
 // but this proposal may need more than it
-//
+func SimulateStoreCodeProposal(wasmKeeper WasmKeeper) simtypes.MsgSimulatorFn {
+	return func(r *rand.Rand, ctx sdk.Context, accs []simtypes.Account) sdk.Msg {
+		// use the default gov module account address as authority
+		var authority sdk.AccAddress = address.Module("gov")
```
alpe, line 110: Nit: nice to see this! To be on the safe side for non gov authority scenarios, you could modify the interface and do:
```go
authority := wasmKeeper.GetAuthority() # returns string
```
Here and other places in this file

x/wasm/simulation/proposals.go
```diff
@@ -239,20 +200,19 @@ func DefaultSimulateUpdateAdminProposalContractSelector(
...
 		}
 
-		return types.NewUpdateAdminProposal(
-			simtypes.RandStringOfLength(r, 10),
-			simtypes.RandStringOfLength(r, 10),
-			simtypes.RandomAccounts(r, 1)[0].Address.String(),
-			ctAddress.String(),
-		)
+		return &types.MsgUpdateAdmin{
+			Sender:   "",
```
alpe, line 212: Here and others: shouldn't this be set to the `authority` address? If we lease this empty, we can not compare it to the authority address set on the keeper for gov context. I am also a bit nervous about side effects and that this scenario could be created by a contract that ignores [authorization](https://github.com/CosmWasm/wasmd/blob/v0.31.0/x/wasm/keeper/handler_plugin.go#L89)

When we discussed empty address, I had CLI usage in mind only where the authority address would be fetched from chain.

alpe (commented):
Very good start. This is not a trivial task 👍

faddat:
> Very good start. This is not a trivial task 👍

Indeed, @alpe is right, this is absolutely not a trivial task.  Thanks for taking this on.

x/wasm/types/tx.go
```diff
@@ -478,5 +478,165 @@ func (msg MsgUpdateParams) GetSignBytes() []byte {
 }
 
 func (msg MsgUpdateParams) ValidateBasic() error {
+	if _, err := sdk.AccAddressFromBech32(msg.Authority); err != nil {
```
alpe, line 481: 👍

x/wasm/types/tx.go
```diff
@@ -478,5 +478,165 @@ func (msg MsgUpdateParams) GetSignBytes() []byte {
...
+
+func (msg MsgStoreAndInstantiateContract) GetSignBytes() []byte {
+	return sdk.MustSortJSON(ModuleCdc.MustMarshalJSON(&msg))
+}
+
+func (msg MsgStoreAndInstantiateContract) ValidateBasic() error {
+	if _, err := sdk.AccAddressFromBech32(msg.Authority); err != nil {
+		return errorsmod.Wrap(err, "authority")
+	}
+
```
alpe, line 604: Use `ValidateVerificationInfo(p.Source, p.Builder, p.CodeHash)` as in `StoreAndInstantiateContractProposal` for optional fields

x/wasm/types/tx.go
```diff
@@ -478,5 +478,165 @@ func (msg MsgUpdateParams) GetSignBytes() []byte {
...
+			return errorsmod.Wrap(err, "instantiate permission")
+		}
+		// AccessTypeOnlyAddress is still considered valid as legacy instantiation permission
+		// but not for new contracts
+		if msg.InstantiatePermission.Permission == AccessTypeOnlyAddress {
+			return ErrInvalid.Wrap("unsupported type, use AccessTypeAnyOfAddresses instead")
+		}
+		// AccessTypeOnlyAddress is still considered valid as legacy instantiation permission
+		// but not for new contracts
+		if msg.InstantiatePermission.Permission == AccessTypeOnlyAddress {
```
alpe, line 637: same code as above

alpe (commented):
Thanks for the updates! I added some minor ones. 
Looking forward to see the integration with the keeper for the gov process

x/wasm/keeper/msg_server.go
```diff
@@ -11,26 +11,14 @@ import (
...
-type paramStore interface {
-	GetAuthority() string
-	SetParams(ctx sdk.Context, params types.Params) error
-}
-
 // grpc message server implementation
 type msgServer struct {
-	keeper     types.ContractOpsKeeper
-	paramStore paramStore
+	keeper Keeper
```
alpe, line 16: let's use a pointer here, in case we need to mock some functionality later in tests

x/wasm/keeper/msg_server.go
```diff
@@ -44,7 +32,14 @@ func (m msgServer) StoreCode(goCtx context.Context, msg *types.MsgStoreCode) (*t
 		return nil, errorsmod.Wrap(err, "sender")
 	}
 
-	codeID, checksum, err := m.keeper.Create(ctx, senderAddr, msg.WASMByteCode, msg.InstantiatePermission)
+	var policy AuthorizationPolicy
+	if msg.Sender == m.keeper.GetAuthority() {
+		policy = GovAuthorizationPolicy{}
+	} else {
+		policy = DefaultAuthorizationPolicy{}
+	}
```
alpe, line 40: This is the switch that we discussed but seeing this in front of me now in all the methods, I think it would be nice to move this into some helper function like `func selectAuthorizationPolicy(actor string)AuthorizationPolicy` or with a better name

x/wasm/simulation/operations.go
```diff
@@ -38,6 +38,7 @@ const (
 
 // WasmKeeper is a subset of the wasm keeper used by simulations
 type WasmKeeper interface {
+	GetAuthority() string
```
alpe, line 41: 👍

x/wasm/types/codec.go
```diff
@@ -68,6 +72,10 @@ func RegisterInterfaces(registry types.InterfaceRegistry) {
 		&MsgIBCSend{},
 		&MsgUpdateInstantiateConfig{},
 		&MsgUpdateParams{},
+		&MsgSudoContract{},
+		&MsgPinCodes{},
+		&MsgUnpinCodes{},
+		&MsgStoreAndInstantiateContract{},
```
alpe, line 78: 👍

x/wasm/keeper/msg_server_integration_test.go
```diff
@@ -136,3 +141,643 @@ func TestUpdateParams(t *testing.T) {
...
+
+			//then
+			if spec.expErr {
+				require.Error(t, err)
+			} else {
+				require.NoError(t, err)
+			}
+		})
+	}
+}
```
alpe, line 783: Very nice to see all these tests! This will give high confidence that things work as expected. Thanks for taking care!

Can you add a test for `UpdateInstantiateConfig` so that we have this covered, too?

PR #1298, opened 2023-03-27
Update OnRecvPacket method to panic when an error is returned by the VM

### LINKED ISSUE #1220: Let contract fully abort IBC receive in certain case
### Placeholder

To be refined better by @ethanfrey 

### Background
In IBC-go:
`OnRecvPacket` returns an Ack object that can either indicate success or failure of the operation. State is only persisted for the success case. 
Returning `nil` would persist state but not store the Acknowledgment.
The forth case is `panic` which would not persist state nor store an Ack.

```go
	cacheCtx, writeFn = ctx.CacheContext()
	ack := cbs.OnRecvPacket(cacheCtx, msg.Packet, relayer)
	if ack == nil || ack.Success() {
		// write application state changes for asynchronous and successful acknowledgements
		writeFn()
	} else {
		// NOTE: The context returned by CacheContext() refers to a new EventManager, so it needs to explicitly set events to the original context.
		// Events should still be emitted from failed acks and asynchronous acks
		ctx.EventManager().EmitEvents(cacheCtx.EventManager().Events())
	}

	// Set packet acknowledgement only if the acknowledgement is not nil.
	// NOTE: IBC applications modules may call the WriteAcknowledgement asynchronously if the
	// acknowledgement is nil.
	if ack != nil {
		if err := k.ChannelKeeper.WriteAcknowledgement(ctx, cap, msg.Packet, ack); err != nil {
			return nil, err
		}
	}
```
in `modules/core/keeper/msg_server.go`

pinosu:
Resolves #1220

### DIFF
--- a/x/wasm/relay_test.go
+++ b/x/wasm/relay_test.go
@@ -32,10 +32,13 @@ func TestFromIBCTransferToContract(t *testing.T) {
 
 	transferAmount := sdk.NewInt(1)
 	specs := map[string]struct {
-		contract             wasmtesting.IBCContractCallbacks
-		setupContract        func(t *testing.T, contract wasmtesting.IBCContractCallbacks, chain *wasmibctesting.TestChain)
-		expChainABalanceDiff math.Int
-		expChainBBalanceDiff math.Int
+		contract                    wasmtesting.IBCContractCallbacks
+		setupContract               func(t *testing.T, contract wasmtesting.IBCContractCallbacks, chain *wasmibctesting.TestChain)
+		expChainAPendingSendPackets int
+		expChainBPendingSendPackets int
+		expChainABalanceDiff        math.Int
+		expChainBBalanceDiff        math.Int
+		expErr                      bool
 	}{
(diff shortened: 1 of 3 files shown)


x/wasm/relay_test.go
```diff
@@ -44,26 +47,33 @@ func TestFromIBCTransferToContract(t *testing.T) {
...
 		},
 		"error": {
 			contract: &errorReceiverContract{},
 			setupContract: func(t *testing.T, contract wasmtesting.IBCContractCallbacks, chain *wasmibctesting.TestChain) {
 				c := contract.(*errorReceiverContract)
 				c.t = t
 			},
-			expChainABalanceDiff: sdk.ZeroInt(),
-			expChainBBalanceDiff: sdk.ZeroInt(),
+			expChainAPendingSendPackets: 1,
```
alpe, line 72: 👍 good test

alpe (approved):
LGTM :surfer:

Issue #1289, opened 2023-03-23
Return non redacted error content on IBC packet recv


alpe:
Current response in json is: `{"error":"ABCI code: 5: error handling packet: see events for details"}`
Instead of the default error constructor, we could pass the string directly

Note: Only for this case: https://github.com/CosmWasm/wasmd/blob/v0.31.0/x/wasm/keeper/relay.go#L135-L136

See also #1220

alpe:
Does it make sense to prevent empty error strings? Technically it is not an issue and would save some gas for the relayer.

0xekez:
this would be a really nice feature 🙏

Issue #1288, opened 2023-03-23
Emit event for ibc error returned


alpe:
The json returned for contract errors on IBC package receive is redacted: 
```json
{"error":"ABCI code: 5: error handling packet: see events for details"}`
```

> It is the responsibility of the application developer to emit error details in events.

https://github.com/cosmos/ibc-go/pull/1565/files#diff-c84c5415c1edd86b6332b4c8e95c325d4551550993fafc997bf9453a81f1b192R31

## Update: 
With the work in #1220  and #1289  there should not be any need for emitting the original error message as event. The contract's error string is included in the error Ack.
But this is not the case for sub-messages that fail. See https://github.com/cosmos/ibc-go/blob/v7.0.0/modules/apps/transfer/ibc_module.go#L202 for an example event

In [ibc-go](https://github.com/cosmos/ibc-go/blob/v7.0.0/modules/core/keeper/msg_server.go#L459) all events are emitted on error Ack as well. This is a different behaviour than we have with other contract calls.
There are 2 cases to consider:
1) the contract returns an error Ack and events
2) the contract returns a success Ack but any of the sub-messages returned fails

As a simple solution, we can drop all events on error Ack in wasm/ibc.go
Or we always accept contract events but drop sub-message events on error in keeper/relay.go.

PR #1299, opened 2023-03-27
Spike error ack

alpe:
Spike that covers:

* #1289 
* #1288
* #1220


x/wasm/keeper/relay.go
```diff
@@ -130,13 +133,34 @@ func (k Keeper) OnRecvPacket(
...
 	}
 	// note submessage reply results can overwrite the `Acknowledgement` data
-	return k.handleContractResponse(ctx, contractAddr, contractInfo.IBCPortID, res.Ok.Messages, res.Ok.Attributes, res.Ok.Acknowledgement, res.Ok.Events)
+	data, err := k.handleContractResponse(ctx, contractAddr, contractInfo.IBCPortID, res.Ok.Messages, res.Ok.Attributes, res.Ok.Acknowledgement, res.Ok.Events)
+	if err != nil {
+		// submessage errors result in error ACK
+		return nil, err
+	}
+	// success ACK
+	return ContractConfirmStateAck(data), nil
```
webmaster128, line 151: I think this code or comment is misleading because it can be either a success or and error or a non-binary acknowledgement type. We should not try to give it any meaning.

x/wasm/keeper/relay.go
```diff
@@ -130,13 +133,34 @@ func (k Keeper) OnRecvPacket(
...
+}
+
+var _ ibcexported.Acknowledgement = ContractConfirmStateAck{}
+
+type ContractConfirmStateAck []byte
+
+func (w ContractConfirmStateAck) Success() bool {
+	return true // always commit state
+}
```
webmaster128, line 160: Is there any documentation what `Success()` in the `ibcexported.Acknowledgement` means?

x/wasm/ibc.go
```diff
@@ -267,23 +268,38 @@ func (i IBCHandler) OnRecvPacket(
 		return channeltypes.NewErrorAcknowledgement(errorsmod.Wrapf(err, "contract port id"))
```
webmaster128, line 268: Is this something where we can provide a better error message than the redacted error too?

x/wasm/ibc.go
```diff
@@ -267,23 +268,38 @@ func (i IBCHandler) OnRecvPacket(
...
+
+	em := sdk.NewEventManager()
+	ack, err := i.keeper.OnRecvPacket(ctx.WithEventManager(em), contractAddr, msg)
 	if err != nil {
+		// the state gets reverted, we keep only wasm events that do not contain custom data
+		for _, e := range em.Events() {
+			if types.IsAcceptedEventOnRecvPacketErrorAck(e.Type) {
+				ctx.EventManager().EmitEvent(e)
+			}
+		}
```
webmaster128, line 280: Do we really want to preserve those events? I might be messing something but to me it's clearer if we discard them and add all relevent information in the error event below.

x/wasm/ibc.go
```diff
@@ -267,23 +268,38 @@ func (i IBCHandler) OnRecvPacket(
...
+		// the `NewErrorAcknowledgement` redacts the error message, it is
+		// recommended by the ibc-module devs to log the raw error as event:
+		ctx.EventManager().EmitEvent(
+			sdk.NewEvent(
```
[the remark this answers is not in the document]
webmaster128: What about 2 events:
1. EventTypePacketRecv: we receive a packet
2. EventTypePacketRecvContractErr: we got a contract error which is turned into an error ack

While `EventTypePacketRecvContractErr` is an error, the absence of it does not mean success.

x/wasm/ibc.go
```diff
@@ -267,23 +268,38 @@ func (i IBCHandler) OnRecvPacket(
...
-func (w ContractConfirmStateAck) Acknowledgement() []byte {
-	return w
+	// emit all contract and submessage events on success
+	ctx.EventManager().EmitEvents(em.Events())
+	ctx.EventManager().EmitEvent(
+		sdk.NewEvent(
+			types.EventTypePacketRecv,
+			sdk.NewAttribute(types.AttributeKeyContractAddr, contractAddr.String()),
+			sdk.NewAttribute(sdk.AttributeKeyModule, types.ModuleName),
+			sdk.NewAttribute(types.AttributeKeyAckSuccess, fmt.Sprintf("%t", ack.Success())),
```
webmaster128, line 300: We don't know if this us a success or not. It's just _bytes from contract_ that can mean anything.

webmaster128 (commented):
Good stuff. Some first thoughs here.

x/wasm/ibc.go
alpe: In a more complex contract-calls-contracts environment, the event trace may have revealed some infos to track the root cause of an error. But there are some costs of maintaining this in the future. 
👌 let's drop them and keep things simple.

x/wasm/keeper/relay.go
alpe: The best source for `Success()` that I found is the code that calls this in [ibc-go](https://github.com/cosmos/ibc-go/blob/v7.0.0/modules/core/keeper/msg_server.go#L453).

x/wasm/ibc.go
alpe: This attribute was inspired by the ics-20 [impl](https://github.com/cosmos/ibc-go/blob/v7.0.0/modules/apps/transfer/ibc_module.go#L198). At this stage we do know that the ibc packet was successfully handled by the contract (and any sub-messages succeeded). The state will be committed and the Ack data stored.

x/wasm/ibc.go
alpe: Good 👁️ ! Unless we touch our contract-to-port mapping we should never run into this case. Either we `panic` here or we need to return the error event, too.
I would prefer panic to make clear this must not happen

x/wasm/keeper/relay.go
alpe: In this line Success can be defined as case 1 + 2 of https://github.com/CosmWasm/wasmd/issues/697#issuecomment-992573516 . The contract was able to fully handle the packet. State will be committed.

x/wasm/ibc.go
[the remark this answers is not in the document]
alpe: This was also inspired by ics-20 and ics-27 . They use a single event with additional error attribute when set.

alpe:
Thanks for the feedback! 💐 
I have added the `EmitAcknowledgementEvent` function to reduce some duplication. It is inspired by code in ics-27

x/wasm/keeper/relay.go
webmaster128: I raised the issue here: https://github.com/cosmos/ibc-go/issues/3434

alpe:
Implemented in #1353

Issue #1310, opened 2023-04-01
Check and fix usages of error type ErrInvalidMsg


webmaster128:
In an error ack I got error code 10, which is documented as

```go
	// ErrInvalidMsg error when we cannot process the error returned from the contract
	ErrInvalidMsg = errorsmod.Register(DefaultCodespace, 10, "invalid CosmosMsg from the contract")
```

However, it seems like the error `ErrInvalidMsg` is used in places where no contract created [CosmosMsg](https://github.com/CosmWasm/cosmwasm/blob/v1.2.3/packages/std/src/results/cosmos_msg.rs#L27) is in case, such as

```go
func (msg MsgUpdateAdmin) ValidateBasic() error {
	if _, err := sdk.AccAddressFromBech32(msg.Sender); err != nil {
		return errorsmod.Wrap(err, "sender")
	}
	if _, err := sdk.AccAddressFromBech32(msg.Contract); err != nil {
		return errorsmod.Wrap(err, "contract")
	}
	if _, err := sdk.AccAddressFromBech32(msg.NewAdmin); err != nil {
		return errorsmod.Wrap(err, "new admin")
	}
	if strings.EqualFold(msg.Sender, msg.NewAdmin) {
		return errorsmod.Wrap(ErrInvalidMsg, "new admin is the same as the old") // <---- should not be ErrInvalidMsg
	}
	return nil
}
```

For other usages of ErrInvalidMsg it would also be good to double check if "invalid CosmosMsg from the contract" is the correct definition.

alpe:
Fixed in #1317 for v0.32

webmaster128:
Thanks!

PR #1311, opened 2023-04-03
fix : removed the unnecessary usage of  ErrInvalidMsg 

vishal-kanna:
Part of #1310


pinosu:
Thanks for your pr! 
As part of #1310 , I don't think we need a new error type for each possible error. Probably the more generic _ErrInvalid_ could be used.
In this case it could be something like:
`
if strings.EqualFold(msg.Sender, msg.NewAdmin) {
	return errorsmod.Wrap(ErrInvalid, "new admin is the same as the old")
}
`
It would be nice to check if there are other places where _ErrInvalidMsg_ usge is not proper.

pinosu (approved):
Thank you for your contribution! 💯

PR #1316, opened 2023-04-03
feat: QOL Capabilities functions

Reecepbcups:
Per: https://github.com/CosmWasm/wasmvm/pull/413

I have added some capability functions here instead of in wasmvm. 

This way a chain like Juno can add `wasmd.GetDefaultWasmCapabilities()` for future upgrades without us forgetting to add the capabilities (that wasmd implement)

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -161,6 +161,16 @@ func GetEnabledProposals() []wasm.ProposalType {
 	return proposals
 }
 
+// GetDefaultCosmosCapabilities returns the default capabilities for a cosmos-sdk based chain.
+func GetDefaultCosmosCapabilities() string {
+	return "iterator,staking,stargate"
+}
+
+// GetDefaultWasmCapabilities returns the default capabilities for a cosmwasm based chain.
+func GetDefaultWasmCapabilities() string {
+	return "cosmwasm_1_1,cosmwasm_1_2"
+}
+
 // These constants are derived from the above variables.
 // These are the ones we will want to use in the code, based on
 // any overrides above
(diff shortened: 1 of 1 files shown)


pinosu:
Thanks for your pr! 
Alex is on holidays and will be back in 2 weeks and he can give a better feedback on this work.
Regarding your changes, I am not sure if "default cosmos capabilities" and "default wasm capabilities" names are a bit misleading.
Moreover don't you lose visibility on which capabilities you are actually importing, and don't you need to check anyway on every version upgrade that those functions haven't been modified and some capability was added/removed?

Reecepbcups:
@pinosu 

I agree default can be misleading. Will change that worder

> Moreover don't you lose visibility on which capabilities you are actually importing

I just wanted to solve for new wasmvm/wasmd funcationality as it comes out. This way upgraded chains don't forget to add cosmwasm_1_# to their capabilities as we get more features.

That way as you upgrade to a future wasmvm 1.3 it would return capabilities cosmwasm_1_1, 1_2, and 1_3 automatically

This is all I was trying to solve for here as we forgot to add this during our testnet launch

pinosu:
Thank you @Reecepbcups ! 
I think the issue you want to solve makes sense, we just need to find the proper way to do it.
Let's continue this discussion when @alpe is back, he might have some good ideas about it.

alpe:
Thanks for the PR! I do understand that new capabilities are easy to miss. Therefore I had highlighted this in the [changelog](https://github.com/CosmWasm/wasmd/blob/v0.31.0/CHANGELOG.md#notable-changes). 
Nevertheless I am very open to make this better.

With your solution, I have some concerns about the method names as well as the grouping. Let's spike some solutions that work nicely with git diff. I have opened https://github.com/CosmWasm/wasmd/issues/1341

alpe:
Closing in favour of #1361 . Thanks for bringing attention to the problem! 🏅

PR #1321, opened 2023-04-07
Fix: Prevent 0 or duplicate code ids in PinCodeProposal , UnPinCodeProposal

ruthishvitwit:
I have tried to add additional conditions to validate basic of pin and unpin code proposal to prevent 0 and duplicate code ids. I have tried to solve the issue in a way that i can understand


pinosu:
Thank you for your PR!
This is issue is part of v0.32.0 milestone so it needs to have as base branch `[releases/v0.3x](https://github.com/CosmWasm/wasmd/tree/releases/v0.3x)` . 
Please create a new PR with this base branch. 
I will close this PR because it is not relevant for v0.40.0 and sdk-47 integration addressed in main branch.

PR #1326, opened 2023-04-10
Add wasm command to support v1 gov proposals

pinosu:


alpe (changes requested):
Nice work! 
When trying to test a submission, I noticed that the new command does not support `--keyring-backend`.
> Error: unknown flag: --keyring-backend

x/wasm/types/validation.go
```diff
@@ -57,7 +57,7 @@ func ValidateSalt(salt []byte) error {
 // ValidateVerificationInfo ensure source, builder and checksum constraints
 func ValidateVerificationInfo(source, builder string, codeHash []byte) error {
 	// if any set require others to be set
-	if len(source) != 0 || len(builder) != 0 || codeHash != nil {
+	if len(source) != 0 || len(builder) != 0 || len(codeHash) != 0 {
```
alpe, line 60: 👍

x/wasm/client/cli/gov_tx.go
```diff
@@ -0,0 +1,836 @@
...
+			if err != nil {
+				return err
+			}
+			authority, err := cmd.Flags().GetString(flagAuthority)
+			if err != nil {
+				return fmt.Errorf("authority: %s", err)
+			}
+
+			if len(authority) == 0 {
+				authority = sdk.AccAddress(address.Module("gov")).String()
```
alpe, line 67: I was expecting that the address would be queried but I guess it is alright to set it for the client already.
While the address can be changed in the keeper(s) on init, it can not be modified on the CLI. I have opened https://github.com/CosmWasm/wasmd/pull/1356 to have a public var instead. Please 👁️

alpe (approved):
Please see my comment on the static address and https://github.com/CosmWasm/wasmd/pull/1356 .
 Other than this, the PR looks very good now.  👍

Issue #911, opened 2022-07-21
Add tokenfactory module


JakeHartnell:
Inspired by this Twitter thread: https://twitter.com/larry0x/status/1550231427161227265

The [tokenfactory module](https://github.com/osmosis-labs/osmosis/tree/main/x/tokenfactory) has been growing on me, and as @larry0x suggests:
> the best way to promote the adoption of tokenfactory is to make it a part of wasmd, so any chain to include wasm automatically gets it

I think this would be great!

faddat:
I think this would be great, too. 

I think that doing token factory, osmosis-style here, as mentioned by Larry, is simply a great idea and something that we should be doing.

alpe:
The code is owned and maintained by the Osmosis team. We should discuss a path with them to make it available that respects their past and future work.
It can also be a dependency in `go.mod` and set it up in the example `app.go`.

onivalidator:
@alpe To play devil's advocate here - it seems to me that we sometimes get bogged down in the semantics of including the original contributors and what have you (who may also have differing views on future development).  If we can fork tokenfactory in and the ones submitting the PR can competently maintain it, then why wouldn't we move forward in the name of development velocity and progress? Notional is a core contributor over there, so there's no finite reason I can see not to move this forward.

I hate to see development slowed by convoluted team discussions that may or may not resolve on any kind of finite timeline - even though I understand where you are coming from. Perhaps my own ignorance here is showing - but I'm a velocity maxi.

jhernandezb:
The main issue I see is that this repository is a reference implementation, teams probably should import`tokenfactory` directly into their chains, now if custom bindings are needed that's probably a different conversation to have. Osmosis uses their own sdk fork, by importing it we will have conflicts with sdk versions and wasmd being an upstream dependency will cause troubles for many projects depending on it. 

Maybe `tokenfactory` should be extracted to its own repo which should resolve sdk dependency issues.

onivalidator:
This seems like a great solution on all fronts, if there's conflicts with the SDK version. Who can we chat up to shop this idea? Seems like tokenfactory in it's own repo is the cleanest and fastest solution for all.

ethanfrey:
> Maybe tokenfactory should be extracted to its own repo which should resolve sdk dependency issues, but still remains optional for chains that wants it and not directly into this repo.

yes, I would love to make this it's own repo. one for the go bindings (just import and add two lines to app.go) and another for the rust bindings (that can be used for any chain).

[...]

PR #1327, opened 2023-04-10
47 with token factory

faddat:
closes: #911 

Discussed in today's Juno call.  We're not sure that there's anyone who wants to use wasmd without the token factory.

Please close this and #911 if not interested in exploring this path.

Thanks


faddat:
changing this to a draft because it isn't passing sims

faddat:
the proto domain is now `osmosis`

alpe:
Thanks for the PR but as discussed before, I am not going to inline token factory code.

PR #1337, opened 2023-04-14
Don't run checkLibwasmVersion automatically on start (0.30)

webmaster128:
See #1338


webmaster128:
Backport not needed, see #1338 for the main PR

PR #1338, opened 2023-04-14
[don't merge] Don't run checkLibwasmVersion automatically on start (0.31.0)

webmaster128:


jhernandezb:
This makes sense to me, letting apps having more control on how to perform the check is better IMO.

x/wasm/module.go
```diff
@@ -280,7 +278,18 @@ func getExpectedLibwasmVersion() string {
...
+// shared libraries that are copied or moved from their default location, e.g. when building the node
+// on one machine and deploying it to other machines.
+//
+// Usually the libwasmvm version (the Rust project) and wasmvm version (the Go project) match. However,
+// there are situations in which this is not the case. This can be during development or if one of the
+// two is patched. In such cases it is advised to not execute the check.
+//
+// An alternative method to obtain the libwasmvm version loaded at runtime is executing
+// `wasmd query wasm libwasmvm-version`.
+func CheckLibwasmVersion(cmd *cobra.Command, args []string) error {
```
alpe, line 292: Thanks for the comments. These are very valid scenarios that need to be considered!
The version check was a very helpful in the past and I would be sad to let it go.
Some options:
a) move the integration to the `cmd/wasmd/root.go` example CLI 
b) introduce a new start cmd flag `--wasm.skip_wasmvm_version_check` (or better name)
c) Add toggle param to `AddModuleInitFlags(startCmd *cobra.Command, skipVersionCheck bool)` which is called by `root.go`

I am in favour of option b) as it does not require code changes for different scenarios

alpe: ok, then let's go with option b). For a long term solution, a project can always replace the call to the [default init flags](https://github.com/CosmWasm/wasmd/blob/v0.40.0-rc.0/cmd/wasmd/root.go#L173) with a custom solution. No need to optimize for that

webmaster128: okay, cool. Could you take over this PR?

alpe:
This is hard to test via unit test. `wasmd start -h` should show the new flag.

pinosu (approved):
LGTM 👍

x/wasm/module.go
```diff
@@ -280,7 +282,26 @@ func getExpectedLibwasmVersion() string {
...
+	skip, err := cmd.Flags().GetBool(flagWasmSkipWasmVMVersionCheck)
```
webmaster128, line 297: Wouldn't it be nicer to move this switch to the caller code? Then `CheckLibwasmVersion` always performs a check (when called)

x/wasm/module.go
```diff
@@ -280,7 +282,26 @@ func getExpectedLibwasmVersion() string {
...
+// shared libraries that are copied or moved from their default location, e.g. when building the node
+// on one machine and deploying it to other machines.
+//
+// Usually the libwasmvm version (the Rust project) and wasmvm version (the Go project) match. However,
+// there are situations in which this is not the case. This can be during development or if one of the
+// two is patched. In such cases it is advised to not execute the check.
+//
+// An alternative method to obtain the libwasmvm version loaded at runtime is executing
+// `wasmd query wasm libwasmvm-version`.
+func CheckLibwasmVersion(cmd *cobra.Command, args []string) error {
```
webmaster128, line 296: Should we change this back to private or leave it available to folks building their own `AddModuleInitFlags`?

webmaster128 (commented):
> This is hard to test via unit test.

You can test it manually by using wasmvm main. This now has the 1.2.3 libwasmvm build and but some intermediate wasmvm version.

webmaster128:
Thank you for finishing this!

PR #1340, opened 2023-04-14
Don't run checkLibwasmVersion automatically on start (0.29)

webmaster128:
See https://github.com/CosmWasm/wasmd/pull/1338


webmaster128:
Backport not needed, see #1338

PR #1353, opened 2023-04-21
Redesign IBC on packet recv error/ result.Err handling

### LINKED ISSUE #1220: Let contract fully abort IBC receive in certain case

### LINKED ISSUE #1288: Emit event for ibc error returned

### LINKED ISSUE #1289: Return non redacted error content on IBC packet recv

alpe:
Resolved when backported:
* See #1289 
* See #1288 
* See #1220

Based on design spike #1299

See this very nice document for a detailed description: https://github.com/CosmWasm/cosmwasm/blob/ibc_packet_receive-erros/docs/ERROR_HANDLING.md#handing-ibc_packet_receive-errors for

### DIFF
--- a/x/wasm/ibc_test.go
+++ b/x/wasm/ibc_test.go
@@ -4,10 +4,124 @@ import (
 	"testing"
 
 	wasmvmtypes "github.com/CosmWasm/wasmvm/types"
+	abci "github.com/cometbft/cometbft/abci/types"
+	"github.com/cometbft/cometbft/libs/rand"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/types/address"
 	clienttypes "github.com/cosmos/ibc-go/v7/modules/core/02-client/types"
 	channeltypes "github.com/cosmos/ibc-go/v7/modules/core/04-channel/types"
+	ibcexported "github.com/cosmos/ibc-go/v7/modules/core/exported"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/CosmWasm/wasmd/x/wasm/keeper"
+	"github.com/CosmWasm/wasmd/x/wasm/types"
 )
+
+func TestOnRecvPacket(t *testing.T) {
+	anyRelayerAddr := sdk.AccAddress(rand.Bytes(address.Len))
+	anyContractIBCPkg := IBCPacketFixture(func(p *channeltypes.Packet) {
+		p.DestinationPort = "wasm.cosmos1w09vr7rpe2agu0kg2zlpkdckce865l3zps8mxjurxthfh3m7035qe5hh7f"
+	})
+	myCustomEvent := sdk.NewEvent("testing")
+	specs := map[string]struct {
+		ibcPkg      channeltypes.Packet
+		contractRsp ibcexported.Acknowledgement
+		contractErr error
+		expEvents   sdk.Events
+		expPanic    bool
+		expAck      ibcexported.Acknowledgement
+	}{
+		"contract returns success response": {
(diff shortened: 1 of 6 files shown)


x/wasm/ibc_test.go
```diff
@@ -4,11 +4,125 @@ import (
...
 	channeltypes "github.com/cosmos/ibc-go/v7/modules/core/04-channel/types"
+	ibcexported "github.com/cosmos/ibc-go/v7/modules/core/exported"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/CosmWasm/wasmd/x/wasm/keeper"
+	"github.com/CosmWasm/wasmd/x/wasm/types"
 )
 
+func TestOnRecvPacket(t *testing.T) {
```
pinosu, line 21: Very good tests! 👍

pinosu (approved):
Good work! 👍

alpe:
This is the last PR missing for a new RC on the 47 branch. 
I will merge and tag now to unblock people. I would still appreciate feedback from you @webmaster128 . Let's use the backport branch for comments. Any changes will still go into main.

PR #1356, opened 2023-04-21
Minor tweak and test files

alpe:
* Moves "default" authority address to global var so that it can be overwritten by chain architects
* Add manual test script to submit a wasm proposal


contrib/local/04-gov.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
pinosu, line 1: Very good example 👍

pinosu (approved):
LGTM 👍

PR #1368, opened 2023-04-27
[Do not merge] ADR-8 vertical spike

alpe:
Spike to discuss a concrete example.
This includes a some of hacks to make it work without modifying wasmvm.

A good start would be to look at `TestIBCAdr8WithTransfer` for the test scenario. Ics29 fees or edge cases are not covered
The middleware integration is done in `app.go`

For the callback to the contracts, 2 new methods are added to the wasmvm interface and mocked for proof of concept
* OnIBCPacketAcked
* OnIBCPacketTimedOut

For the callback registration a custom message was added that the reflect contract emits in the test:
```go
type RegisterPacketProcessedCallback struct {
	// Sequence is the ibc packet sequence number
	Sequence            uint64 `json:"sequence"`
	ChannelID           string `json:"channel_id"`
	PortID              string `json:"port_id"`
	MaxCallbackGasLimit uint64 `json:"max_callback_gas_limit"`
}
```

The message is handled in the `XMessageHandler` and contains some ugly hack for the reflect contract.

### Authorization
* on callback registration, it is checked that the actor is a contract
* on ack/timeout handling, the packet sender is only accepted for a registered hook

### New Packages
* xwasmvm are extensions to WasmVm for the spike only. If they make sense, then they should be moved upstream
* xadr8 are common parts that are not CosmWasm specific. If they make sense, they could be living in ibc-go for example

### DIFF
--- a/x/xadr8/middleware.go
+++ b/x/xadr8/middleware.go
@@ -12,8 +12,19 @@ import (
 
 var _ porttypes.Middleware = &IBCMiddleware{}
 
+// PacketDecoder IBC packet decoder definition
 type PacketDecoder interface {
+	// DecodeSender returns sender address of the given packet, true
+	// false on any issue of when not found
 	DecodeSender(packet channeltypes.Packet) (string, bool)
+}
+
+var _ PacketDecoder = PacketDecoderFn(nil)
+
+type PacketDecoderFn func(packet channeltypes.Packet) (string, bool)
+
+func (p PacketDecoderFn) DecodeSender(packet channeltypes.Packet) (string, bool) {
+	return p(packet)
 }
 
 // CallbackExecutor is an abstract callback executor. To be implemented by contract engines
--- a/x/xadr8/middleware.go
+++ b/x/xadr8/middleware.go
@@ -1,0 +1,142 @@
+package xadr8
+
+import (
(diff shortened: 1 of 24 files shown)


x/xadr8/middleware.go
```diff
@@ -0,0 +1,142 @@
...
+		sender,
+		ExecutorFn(func(ctx sdk.Context, contractAddr sdk.AccAddress, meta CustomDataI) error {
+			return m.exec.OnIBCPacketTimedOut(ctx, packet, relayer, contractAddr, meta)
+		}))
+}
+
+var _ porttypes.Middleware = &IBCMiddleware{}
+
+// IBCMiddlewareAdapter is an adapter to build a Middleware for IBC modules
+type IBCMiddlewareAdapter struct {
```
alpe, line 119: This is unused

alpe:
In case somebody reads the [article](https://medium.com/the-interchain-foundation/actor-callbacks-unlock-interchain-composable-apps-5c8c5e9324bd) by the IBC team which goes in some different direction. This spike was built for them to demo the integration and callback registration. There were also a call and long discussions on internal slack.
This is an important feature for the community. We need to discuss internally how to proceed, now.

webmaster128:
Closing in favour of ADR-8 implementation deployed as part of 0.52

PR #1369, opened 2023-04-29
Bumped SDK to 0.47.2 and CometBFT to 0.37.1

gjermundgaraba:
One good reason to get this in sooner rather than later is the simtest fix for group that intermittently failed: https://github.com/cosmos/cosmos-sdk/issues/15741


alpe:
Very nice! Thank you!

PR #1371, opened 2023-05-02
Bump sdk to v0.47.2

alpe:
See [v0.47.2 changelog](https://github.com/cosmos/cosmos-sdk/blob/v0.47.2/CHANGELOG.md)


alpe:
Same as #1369

PR #1373, opened 2023-05-02
App v2 update

pinosu:


x/wasm/module.go
```diff
@@ -199,6 +215,91 @@ func (AppModule) EndBlock(_ sdk.Context, _ abci.RequestEndBlock) []abci.Validato
...
+	// LegacySubspace is used solely for migration of x/params managed parameters
+	LegacySubspace exported.Subspace `optional:"true"`
+}
+
+// Dependency Injection Outputs
+type wasmOutputs struct {
+	depinject.Out
+
+	Module     appmodule.AppModule
+	WasmKeeper *Keeper
```
alpe, line 256: It would be nice, if we can setup and return the `scopedWasmKeeper` as well.

x/wasm/module.go
```diff
@@ -199,6 +215,91 @@ func (AppModule) EndBlock(_ sdk.Context, _ abci.RequestEndBlock) []abci.Validato
...
+}
+
+func provideModule(in wasmInputs) wasmOutputs {
```
alpe, line 259: this function needs to be public or will cause an error on DI
```suggestion
func ProvideModule(in wasmInputs) wasmOutputs {
```

alpe (commented):
Good start, let's add the module to `app_v2_config.go` and remove setup and legacy registration from `app_v2.go`

x/wasm/module.go
alpe, line 256: `app_v2.go` expects as  value of the wasm Keeper. Either change it there or here

faddat:
@alpe & @pinosu sirs, keep the fire.  This is great.

PR #1374, opened 2023-05-03
Add wasm store  to import-export sims

### LINKED ISSUE #1372: import export simulation test for `x/wasm` is missing
Currently, `x/wasm` is not added to store prefix list should be checked inside `TestAppImportExport`. Is any reasons not to add the check for `x/wasm`?
https://github.com/CosmWasm/wasmd/blob/main/app/sim_test.go#L158

By the way, the simulation test happens error as follows since the random genesis state does not generate proper `LastInstanceID`.
https://github.com/CosmWasm/wasmd/blob/main/x/wasm/simulation/genesis.go#L17

The error would be like:
```
compared 2 different key/value pairs between KVStoreKey{0xc001218320, wasm} and KVStoreKey{0xc000302450, wasm}
    sim_test.go:275: 
                Error Trace:    /home/kilem/forbole-dev/desmos/app/sim_test.go:275
                Error:          Not equal: 
                                expected: 2
                                actual  : 0
                Test:           TestAppImportExport
                Messages:       store A  => 
                            
                                store B lastContractId => 
                                store A  => 
                                store B  =>
```

May I take care of this issue if the thing I proposed  is worth to you?

alpe:
Resolves #1372

### DIFF
--- a/app/sim_test.go
+++ b/app/sim_test.go
@@ -9,6 +9,8 @@ import (
 	"strings"
 	"testing"
 
+	wasmtypes "github.com/CosmWasm/wasmd/x/wasm/types"
+
 	dbm "github.com/cometbft/cometbft-db"
 	abci "github.com/cometbft/cometbft/abci/types"
 	"github.com/cometbft/cometbft/libs/log"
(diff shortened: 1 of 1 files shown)


pinosu (approved):
LGTM 👍

alpe:
This addresses only the SDK 47 branch. No need to backport

PR #1380, opened 2023-05-05
Inline ibc packet sender interface and minor chore

alpe:
Follow up on #1375

/cc @assafmo

### DIFF
--- a/x/wasm/types/expected_keepers.go
+++ b/x/wasm/types/expected_keepers.go
@@ -72,11 +72,21 @@ type StakingKeeper interface {
 type ChannelKeeper interface {
 	GetChannel(ctx sdk.Context, srcPort, srcChan string) (channel channeltypes.Channel, found bool)
 	GetNextSequenceSend(ctx sdk.Context, portID, channelID string) (uint64, bool)
-	SendPacket(ctx sdk.Context, channelCap *capabilitytypes.Capability, packet ibcexported.PacketI) error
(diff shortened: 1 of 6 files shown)


x/wasm/types/expected_keepers.go
```diff
@@ -72,13 +72,23 @@ type StakingKeeper interface {
...
 }
 
+// IBCPacketSender defines the method for an IBC data package to be submitted.
+// The interface is implemented by the channel keeper on the lowest level in ibc-go. Middlewares or other abstractions
+// can add functionality on top of it. See ics4Wrapper in ibc-go.
+// It is important to choose the right implementation that is configured for any middleware used in the ibc-stack of wasm.
+//
+// For example, when ics-29 fee middleware is set up for the wasm ibc-stack, then the IBCFeeKeeper should be used, so
+// that they are in sync.
+type IBCPacketSender interface {
```
pinosu, line 88: 👍

pinosu (approved):
LGTM 👍

assafmo:
Instead of `IBCPacketSender`, I suggest naming the interface `ICS4Wrapper`. This is consistent with its naming convention used in other places, such as the `ibc-go` repository and the `ibc-hooks` module in Osmosis.

alpe:
> Instead of IBCPacketSender, I suggest naming the interface ICS4Wrapper.

I understand you motivation but unfortunately the interface was extended in the recent ibc-go versions. I has [2 additional methods](https://github.com/cosmos/ibc-go/blob/v7.0.0/modules/core/05-port/types/module.go#L110) that would be useless here.

Issue #1184, opened 2023-01-27
Improve proposal validation


alpe:
Prevent `0` or duplicate code ids in `PinCodesProposal`,`UnpinCodesProposal`,`UpdateInstantiateConfigProposal`

ruthishvitwit:
Is this issue still open? can i work on it?

pinosu:
Closed by #1402

PR #1402, opened 2023-05-16
Feat : Improved Proposal validations

ruthishvitwit:
closes:  #1184 

Prevented 0 and duplicate Code IDs  in PinCodesProposal, UnpinCodesProposal, UpdateInstantiateConfigProposal

### DIFF
--- a/x/wasm/types/errors.go
+++ b/x/wasm/types/errors.go
@@ -85,6 +85,7 @@ var (
 	ErrNoSuchCodeFn = WasmVMFlavouredErrorFactory(sdkErrors.Register(DefaultCodespace, 28, "no such code"),
 		func(id uint64) error { return wasmvmtypes.NoSuchCode{CodeID: id} },
 	)
+	ErrZero = sdkErrors.Register(DefaultCodespace, 29, "0 is not accepted")
 )
 
 // WasmVMErrorable mapped error type in wasmvm and are not redacted
--- a/x/wasm/types/proposal.go
+++ b/x/wasm/types/proposal.go
@@ -852,6 +852,17 @@ func (p PinCodesProposal) ValidateBasic() error {
 	if len(p.CodeIDs) == 0 {
 		return sdkerrors.Wrap(ErrEmpty, "code ids")
 	}
+	for _, num1 := range p.CodeIDs {
+		if num1 == 0 {
+			return sdkerrors.Wrap(ErrZero, "0 is not accepted")
+
+		}
+		for _, num2 := range p.CodeIDs {
+			if num1 == num2 {
+				return sdkerrors.Wrap(ErrDuplicate, "duplicate")
+			}
+		}
+	}
 	return nil
 }
 
@@ -895,6 +906,17 @@ func (p UnpinCodesProposal) ValidateBasic() error {
 	}
 	if len(p.CodeIDs) == 0 {
 		return sdkerrors.Wrap(ErrEmpty, "code ids")
+	}
+	for _, num1 := range p.CodeIDs {
+		if num1 == 0 {
+			return sdkerrors.Wrap(ErrZero, "0 is not accepted")
+
+		}
+		for _, num2 := range p.CodeIDs {
+			if num1 == num2 {
+				return sdkerrors.Wrap(ErrDuplicate, "duplicate")
+			}
+		}
 	}
 	return nil
 }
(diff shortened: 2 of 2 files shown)


x/wasm/types/errors.go
```diff
@@ -85,6 +85,7 @@ var (
 	ErrNoSuchCodeFn = WasmVMFlavouredErrorFactory(sdkErrors.Register(DefaultCodespace, 28, "no such code"),
 		func(id uint64) error { return wasmvmtypes.NoSuchCode{CodeID: id} },
 	)
+	ErrZero = sdkErrors.Register(DefaultCodespace, 29, "0 is not accepted")
```
alpe, line 88: We moved away from more concrete error types. How about using `ErrInvalid` instead?

x/wasm/types/proposal.go
```diff
@@ -852,6 +852,17 @@ func (p PinCodesProposal) ValidateBasic() error {
 	if len(p.CodeIDs) == 0 {
 		return sdkerrors.Wrap(ErrEmpty, "code ids")
 	}
+	for _, num1 := range p.CodeIDs {
+		if num1 == 0 {
+			return sdkerrors.Wrap(ErrZero, "0 is not accepted")
```
alpe, line 857: No need to add a custom error message when it is the same as the one defined on the type. How about:
```suggestion
			return ErrInvalid.Wrap("0 is not accepted")
```

x/wasm/types/proposal.go
```diff
@@ -852,6 +852,17 @@ func (p PinCodesProposal) ValidateBasic() error {
...
+
+		}
+		for _, num2 := range p.CodeIDs {
+			if num1 == num2 {
```
alpe, line 861: This will always end in a match as you are iterating over the same value set. You need to skip num 1 position.

x/wasm/types/proposal.go
```diff
@@ -852,6 +852,17 @@ func (p PinCodesProposal) ValidateBasic() error {
...
+				return sdkerrors.Wrap(ErrDuplicate, "duplicate")
```
alpe, line 862: The value would be a good error message for the duplicate type. What do you think?

x/wasm/types/proposal.go
```diff
@@ -896,6 +907,17 @@ func (p UnpinCodesProposal) ValidateBasic() error {
...
```
alpe, line 917: same comments as above on value set and error message

x/wasm/types/proposal.go
```diff
@@ -970,6 +992,9 @@ func (p UpdateInstantiateConfigProposal) ValidateBasic() error {
 		if found {
 			return sdkerrors.Wrapf(ErrDuplicate, "duplicate code: %d", codeUpdate.CodeID)
 		}
+		if codeUpdate.CodeID == 0 {
+			return sdkerrors.Wrap(ErrZero, "0 is no accepted")
```
alpe, line 996: Could be ErrInvalid as well

alpe (changes requested):
Thank you very much for your contribution! 💐 

There are some conventions and best practices in wasmd that are not well documented. The error messages for example have evolved from the initial way of many concrete to less but more generic types. I will help you with navigating through that.

Please write tests for any new functionality. We have a lot of tests already and they have proven to ensure quality for the module and safety for developers to not break anything accidentally.
See https://app.circleci.com/pipelines/github/CosmWasm/wasmd/3558/workflows/f55c75ef-0494-4526-b20a-a2bd97bc48f7/jobs/14858 for example.

x/wasm/types/proposal.go
ruthishvitwit: yeah got it i will add it now

alpe (commented):
Thanks for the fixes and tests! The code can be merge now.
I have added some more comments for further improvements. Please let me know if you want to address them or have it merged as it is.

alpe (commented):
Very nice progress! Thanks for cleaning up the test!

x/wasm/types/proposal_test.go
```diff
@@ -1139,3 +1139,100 @@ func TestProposalJsonSignBytes(t *testing.T) {
...
 	}
 }
+
+func TestProposalCodeIDs(t *testing.T) {
+	specs := map[string]struct {
+		src    govtypes.Content
+		expErr bool
+	}{
+
+		"Pin empty code id list": {
```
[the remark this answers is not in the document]
ruthishvitwit: Its my first time writing tests. Thanks for guiding me through this. I will make the required changes now.

alpe (approved):
Very nice! Thank you for your help! 🥇

PR #1412, opened 2023-05-19
Feat : Trimmed label to prevent white spaces

### LINKED ISSUE #1183: Trim label to prevent whitespaces
or other non printable chars

ruthishvitwit:
closes #1183 

The aim of this pr is to prevent white spaces  in labels by either silent trimming before storing  or to fail when the label is not trimmed

### DIFF
--- a/x/wasm/types/proposal.go
+++ b/x/wasm/types/proposal.go
@@ -208,10 +208,7 @@ func (p InstantiateContractProposal) ValidateBasic() error {
 	if p.CodeID == 0 {
 		return errorsmod.Wrap(sdkerrors.ErrInvalidRequest, "code id is required")
 	}
-	//this code can be removed based on the preference between silent trimming the label or validating the label
-	if p.Label != strings.TrimSpace(p.Label) {
-		p.Label = strings.TrimSpace(p.Label)
-	}
+	p.Label = strings.TrimSpace(p.Label)
 
 	if err := ValidateLabel(p.Label); err != nil {
 		return err
--- a/x/wasm/types/proposal.go
+++ b/x/wasm/types/proposal.go
@@ -208,6 +208,10 @@ func (p InstantiateContractProposal) ValidateBasic() error {
 	if p.CodeID == 0 {
 		return errorsmod.Wrap(sdkerrors.ErrInvalidRequest, "code id is required")
 	}
+	//this code can be removed based on the preference between silent trimming the label or validating the label
+	if p.Label != strings.TrimSpace(p.Label) {
+		p.Label = strings.TrimSpace(p.Label)
+	}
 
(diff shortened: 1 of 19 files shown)


x/wasm/types/proposal.go
```diff
@@ -208,6 +208,10 @@ func (p InstantiateContractProposal) ValidateBasic() error {
 	if p.CodeID == 0 {
 		return errorsmod.Wrap(sdkerrors.ErrInvalidRequest, "code id is required")
 	}
+	//this code can be removed based on the preference between silent trimming the label or validating the label
+	if p.Label != strings.TrimSpace(p.Label) {
```
alpe, line 212: Good idea to always trim for gov proposals. So that we are safe to not reject any in flight proposals. The code can be simplified though: 
```suggestion
	p.Label != strings.TrimSpace(p.Label)
```

x/wasm/types/proposal.go
```diff
@@ -208,6 +208,10 @@ func (p InstantiateContractProposal) ValidateBasic() error {
 	if p.CodeID == 0 {
 		return errorsmod.Wrap(sdkerrors.ErrInvalidRequest, "code id is required")
 	}
+	//this code can be removed based on the preference between silent trimming the label or validating the label
```
alpe, line 211: Please do not provide options or todos in code comment. It is hard to follow up on this and it opens another chain for work.You can use the PR description or give review feedback on your own code.

A good code comment describes the motivation or adds a bigger context on why this makes sense.

x/wasm/types/validation.go
```diff
@@ -40,6 +41,9 @@ func ValidateLabel(label string) error {
 	if len(label) > MaxLabelSize {
 		return ErrLimit.Wrapf("cannot be longer than %d characters", MaxLabelSize)
 	}
+	if label != strings.TrimSpace(label) {
+		return ErrInvalid.Wrapf("'%s'This label is not trimmed", label)
```
alpe, line 45: Also very good to reject all labels with whitespaces as default behaviour. This ensures a clean state.
No need to return the original string, just the error context. For example:
```suggestion
		return ErrInvalid.Wrap("label must not start/end with whitespaces")
```

alpe (commented):
Good start!
I added some comments to polish the code. Please add test cases for regression. You can find existing ones in `proposal_test.go` and `tx_test.go`

x/wasm/types/proposal.go
ruthishvitwit: got it

x/wasm/types/proposal.go
ruthishvitwit: do u mean
p.Label = strings.TrimSpace(p.Label)

PR #1417, opened 2023-05-25
chore: various version bumps

faddat:
This PR:

* bumps cometbft-db to v0.8.0
* bumps cosmos/gogoproto to v1.4.10, but replaces it temporarily with v1.4.8 until the release of sdk v0.47.3
* bumps ics23 to v0.10.0
* bumps the gin replace to v1.9.0 as in release/v0.47.x of cosmos-sdk

Please let me know if you'd like this broken into multiple PR's -- I figure it all goes pretty well under the heading "bumps"

Additional work originally found here has been moved into:

* #1420 
* #1419


alpe:
Thanks @faddat for the work. I had quite some hick up with cosmos/gogoproto  already. It is not really required to bump everything to the latest version. Wasmd is a library and we follow the SDK with their versions as a minimum.
This gives the chain architects more options to select their preferred versions.
Please do smaller PRs if you want anything concrete upgraded as golangci-lint  for example. This is normal chore and very much appreciated.
For the v0.40.0 release, I would not include any of the updates. Only the ibc-go patch upgrade for security.

faddat:
10000000% about the issues with cosmos/gogoproto -- @larry0x found a fix, I'll add that to this...

faddat:
@alpe I've got to disagree about the lack of need to update everything to its latest version -- this repository is frequently used as a template, and I think it's by far better to have everything super super fresh, so that when downstreams make new chains, there aren't oddities in dependencies.

faddat:
on second thought, I'm going to revert the changes to the proto folder, makefile and x found here, and make a second pr with those...

alpe:
Except for `github.com/cometbft/cometbft-db v0.8.0` all versions were bumped with the last SDK release. I am closing this now as we are following the SDK with their versions. 
Nevertheless, chain architects are always free to decide on their dependency versions themself via their projects go.mod definitions.

PR #1420, opened 2023-05-27
chore: maintain proto setup

faddat:
This PR:

* uses the latest proto-builder
* pins the versions that we use from the `buf` registry 
* rebuilds the .proto files

### DIFF
--- a/x/wasm/types/authz.pb.go
+++ b/x/wasm/types/authz.pb.go
@@ -5,6 +5,10 @@ package types
 
 import (
 	fmt "fmt"
+	io "io"
+	math "math"
+	math_bits "math/bits"
+
 	_ "github.com/cosmos/cosmos-proto"
 	types "github.com/cosmos/cosmos-sdk/codec/types"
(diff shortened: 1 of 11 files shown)


x/wasm/types/authz.pb.go
```diff
@@ -5,25 +5,22 @@ package types
 
 import (
 	fmt "fmt"
-	io "io"
```
alpe, line 8: Can you run `make format` to have the imports formatted well?
fyi: `make proto-all` should be the preferred target for generation. It includes linter and formatting

alpe (changes requested):
Nice work. Thank you for the help!
Just formatting, then it is good to be merged

x/wasm/types/authz.pb.go
faddat: :) sounds great

pinosu (approved):
Thanks for your contribution! LGTM 👍

alpe (approved):
nice work! 🏅

Issue #1296, opened 2023-03-27
Upgrade to wasmvm v1.3


alpe:
when released

chipshort:
Some new queries and messages have been added there:
* [DistributionQuery::DelegatorWithdrawAddress](https://github.com/CosmWasm/wasmvm/pull/442)
* [BankQuery::AllDenomMetadata as well as BankQuery::DenomMetadata](https://github.com/CosmWasm/wasmvm/pull/430). This one has my [corresponding PR](https://github.com/CosmWasm/wasmd/pull/1426) already
* [MsgFundCommunityPool](https://github.com/CosmWasm/wasmvm/pull/443). Also has a [corresponding PR](https://github.com/CosmWasm/wasmd/pull/1458) already

pinosu:
Completed

PR #1426, opened 2023-05-31
Add AllDenomMetadata BankQuery

chipshort:
See also https://github.com/CosmWasm/wasmvm/pull/430 and https://github.com/CosmWasm/cosmwasm/pull/1703

### DIFF
--- a/x/wasm/keeper/query_plugins_test.go
+++ b/x/wasm/keeper/query_plugins_test.go
@@ -390,7 +390,7 @@ func TestBankQuerierMetadata(t *testing.T) {
 	exp := wasmvmtypes.DenomMetadata{
 		Name: "Test Token",
 		Base: "utest",
-		DenomUnits: []*wasmvmtypes.DenomUnit{
+		DenomUnits: []wasmvmtypes.DenomUnit{
 			{
 				Denom:    "utest",
 				Exponent: 0,
@@ -441,7 +441,7 @@ func TestBankQuerierAllMetadata(t *testing.T) {
 			{
 				Name: "Test Token",
 				Base: "utest",
-				DenomUnits: []*wasmvmtypes.DenomUnit{
+				DenomUnits: []wasmvmtypes.DenomUnit{
 					{
 						Denom:    "utest",
 						Exponent: 0,
--- a/x/wasm/keeper/query_plugins_test.go
+++ b/x/wasm/keeper/query_plugins_test.go
@@ -1,6 +1,7 @@
 package keeper_test
 
 import (
+	"context"
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
@@ -357,6 +358,49 @@ func TestBankQuerierBalance(t *testing.T) {
 }
 
 func TestBankQuerierMetadata(t *testing.T) {
+	metadata := banktypes.Metadata{
+		Name: "Test Token",
+		Base: "utest",
+		DenomUnits: []*banktypes.DenomUnit{
+			{
+				Denom:    "utest",
+				Exponent: 0,
+			},
+		},
+	}
+
+	mock := bankKeeperMock{GetDenomMetadataFn: func(ctx sdk.Context, denom string) (banktypes.Metadata, bool) {
+		if denom == "utest" {
+			return metadata, true
+		} else {
+			return banktypes.Metadata{}, false
+		}
+	}}
+
+	ctx := sdk.Context{}
+	q := keeper.BankQuerier(mock)
+	gotBz, gotErr := q(ctx, &wasmvmtypes.BankQuery{
+		DenomMetadata: &wasmvmtypes.DenomMetadataQuery{
+			Denom: "utest",
+		},
+	})
+	require.NoError(t, gotErr)
+	var got wasmvmtypes.DenomMetadataResponse
+	require.NoError(t, json.Unmarshal(gotBz, &got))
+	exp := wasmvmtypes.DenomMetadata{
+		Name: "Test Token",
+		Base: "utest",
+		DenomUnits: []*wasmvmtypes.DenomUnit{
+			{
+				Denom:    "utest",
+				Exponent: 0,
+			},
+		},
+	}
+	assert.Equal(t, exp, got)
+}
+
+func TestBankQuerierAllMetadata(t *testing.T) {
 	metadata := []banktypes.Metadata{
 		{
 			Name: "Test Token",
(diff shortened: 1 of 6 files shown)


x/wasm/keeper/query_plugins_test.go
```diff
@@ -356,6 +357,94 @@ func TestBankQuerierBalance(t *testing.T) {
...
+					{
+						Denom:    "utest",
+						Exponent: 0,
+					},
+				},
+			},
+		},
+	}
+	assert.Equal(t, exp, got)
+}
```
alpe, line 446: Thanks for writing tests! For legacy reasons, the query plugin has coverage not as high as other parts. Therefore I would love to improve this as we go. I don't want to push all the work on you but in case you are full into testing, it would be good to cover the unknown denom case as well. 
The pagination data from the wasmvm request is also not covered. It would be good to have a test case with and without pagination set. You could capture the value in the bank mock and compare it to match the expectations later, for example

x/wasm/keeper/query_plugins_test.go
```diff
@@ -673,9 +762,11 @@ func (m mockWasmQueryKeeper) GetCodeInfo(ctx sdk.Context, codeID uint64) *types.
...
 
 type bankKeeperMock struct {
-	GetSupplyFn      func(ctx sdk.Context, denom string) sdk.Coin
-	GetBalanceFn     func(ctx sdk.Context, addr sdk.AccAddress, denom string) sdk.Coin
-	GetAllBalancesFn func(ctx sdk.Context, addr sdk.AccAddress) sdk.Coins
+	GetSupplyFn         func(ctx sdk.Context, denom string) sdk.Coin
+	GetBalanceFn        func(ctx sdk.Context, addr sdk.AccAddress, denom string) sdk.Coin
+	GetAllBalancesFn    func(ctx sdk.Context, addr sdk.AccAddress) sdk.Coins
+	GetDenomMetadataFn  func(ctx sdk.Context, denom string) (banktypes.Metadata, bool)
+	GetDenomsMetadataFn func(ctx context.Context, req *banktypes.QueryDenomsMetadataRequest) (*banktypes.QueryDenomsMetadataResponse, error)
```
alpe, line 769: 👍

alpe (commented):
Very nice work! Thanks for moving this forward. 💐 
I have added some comments on testing but otherwise it looks great.
Unfortunately, I can not merge it without the wasmvm dependency resolved first.

x/wasm/keeper/query_plugins_test.go
chipshort: Most of the tests were written by [nik-suri](https://github.com/nik-suri) (the original contributor). I added the unknown denom case now and one test for comparing the supplied pagination with the converted one (I hope that's what you meant).

x/wasm/keeper/query_plugins_test.go
```diff
@@ -445,6 +453,40 @@ func TestBankQuerierAllMetadata(t *testing.T) {
 	assert.Equal(t, exp, got)
 }
 
+func TestBankQuerierAllMetadataPagination(t *testing.T) {
+	var requestedPagination *query.PageRequest
```
alpe, line 457: Nit: to make it more verbose
```suggestion
	var capturedPagination *query.PageRequest
```

x/wasm/keeper/query_plugins_test.go
```diff
@@ -445,6 +453,40 @@ func TestBankQuerierAllMetadata(t *testing.T) {
...
+	mock := bankKeeperMock{GetDenomsMetadataFn: func(ctx context.Context, req *banktypes.QueryDenomsMetadataRequest) (*banktypes.QueryDenomsMetadataResponse, error) {
+		// t.Log("test")
+		// t.Log(req == nil)
```
alpe, line 460: nit, please remove the logs

x/wasm/keeper/query_plugins_test.go
```diff
@@ -445,6 +453,40 @@ func TestBankQuerierAllMetadata(t *testing.T) {
...
+			Metadatas: []banktypes.Metadata{},
+			Pagination: &query.PageResponse{
+				NextKey: nil,
+			},
+		}, nil
+	}}
+
+	ctx := sdk.Context{}
+	q := keeper.BankQuerier(mock)
+	pagination := &wasmvmtypes.PageRequest{
```
alpe, line 472: nit: please remove the object and just use the expected values in the assert lines below:
`assert.Equal(t, uint64(10), requestedPagination.Limit)`

or use `&query.PageRequest{..}` and move the declaration down as described below

x/wasm/keeper/query_plugins_test.go
```diff
@@ -445,6 +453,40 @@ func TestBankQuerierAllMetadata(t *testing.T) {
...
+	_, gotErr := q(ctx, &wasmvmtypes.BankQuery{
+		AllDenomMetadata: &wasmvmtypes.AllDenomMetadataQuery{
+			Pagination: &wasmvmtypes.PageRequest{
+				Key:   []byte("key"),
+				Limit: 10,
+			},
+		},
+	})
+	require.NoError(t, gotErr)
+	require.Equal(t, requestedPagination.Key, pagination.Key)
```
alpe, line 485: nit: prefer `assert` over `require` when you don't want to abort the test already. 
For example `require.NoError(t, gotErr)` makes sense as the next asserts would be wrong anyway.
While assert(key) and assert (limit) can both add value to the test output. You would avoid going slice by slice.

Sometimes is more efficient to just compare the captured object with an object of same type. This is often more readable and ensures changes to the object type are covered as well
For example:
exp := &query.PageRequest{...}
assert.Equal(t, exp, got)

nit: the pattern with `testify` is usually: Equal(t, expect, got) . While it works the other way around as well, the output on a failing test can be very misleading.

x/wasm/keeper/query_plugins_test.go
```diff
@@ -398,6 +398,14 @@ func TestBankQuerierMetadata(t *testing.T) {
...
 	}
 	assert.Equal(t, exp, got.Metadata)
+
+	_, gotErr2 := q(ctx, &wasmvmtypes.BankQuery{
+		DenomMetadata: &wasmvmtypes.DenomMetadataQuery{
+			Denom: "uatom",
+		},
+	})
+	require.Error(t, gotErr2)
+	assert.Contains(t, gotErr2.Error(), "uatom: not found")
```
alpe, line 408: 🎉 Thanks for adding this!

alpe (commented):
Thanks a lot for adding the requested test cases! I appreciate the work a lot and added some detailed feedback on the test code in return. The tests are helping and there is no hard requirement to change them. Just sharing a bit of knowledge from working a lot in Go projects.

x/wasm/keeper/query_plugins_test.go
chipshort: oops, totally missed those

alpe:
Thanks again for the updates. Good work!
We are waiting for the wasmvm release now to get this in

alpe:
Linking https://github.com/CosmWasm/wasmd/issues/1296 to unblock this

PR #1443, opened 2023-06-10
chore: bump ibc to v7.1.0

faddat:
- bump ibc
(edited)


alpe:
Thanks for the upgrade. Using this instead of https://github.com/CosmWasm/wasmd/pull/1441

PR #1455, opened 2023-06-16
Add prune wasm codes proposal

### LINKED ISSUE #999: gov: add prune wasm codes proposal
In order to release storage space for unused contracts a prune proposal makes sense.
When executed wasm codes without an instantiation would be removed from CodeInfo and wasmvm.
The proposal may include a relative height (or fix one in the past?) as a threshold to not delete new uploads.

Follow up on:
https://github.com/CosmWasm/wasmd/issues/840#issuecomment-1118525916

pinosu:
Resolves #999


x/wasm/keeper/keeper.go
```diff
@@ -1213,3 +1218,20 @@ func (h DefaultWasmVMContractResponseHandler) Handle(ctx sdk.Context, contractAd
...
+		}
+
+		info := k.GetCodeInfo(ctx, c)
+		if info == nil {
+			return types.ErrNoSuchCodeFn(c).Wrapf("code id %d", c)
+		}
+		k.wasmVM.RemoveCode(info.CodeHash)
+		k.deleteCodeInfo(ctx, c)
+	}
+	return nil
```
alpe, line 1236: you would also need to ensure that no instance exists for a codeID.

x/wasm/keeper/keeper.go
```diff
@@ -1213,3 +1218,20 @@ func (h DefaultWasmVMContractResponseHandler) Handle(ctx sdk.Context, contractAd
...
+// PruneWasmCodes deletes code info for unpinned codes.
+func (k Keeper) PruneWasmCodes(ctx sdk.Context, codeIDs []uint64) error {
+	for _, c := range codeIDs {
+		if k.IsPinnedCode(ctx, c) {
+			continue
+		}
+
+		info := k.GetCodeInfo(ctx, c)
+		if info == nil {
+			return types.ErrNoSuchCodeFn(c).Wrapf("code id %d", c)
```
alpe, line 1231: let's not abort the operation but skip the number. There is not much we can do in this case.

x/wasm/keeper/msg_server_integration_test.go
```diff
@@ -853,3 +853,90 @@ func TestUpdateInstantiateConfig(t *testing.T) {
...
+			// when
+			msgPruneCodes := &types.MsgPruneWasmCodes{
+				Authority: spec.addr,
+				CodeIDs:   []uint64{result.CodeID},
+			}
+			_, err = wasmApp.MsgServiceRouter().Handler(msgPruneCodes)(ctx, msgPruneCodes)
+
+			// then
+			if spec.expErr {
+				require.Error(t, err)
```
alpe, line 930: just add a return or the code after the else block is executed as well. The else can then be dropped and NoError is on the main path

x/wasm/keeper/keeper.go
```diff
@@ -1213,3 +1218,20 @@ func (h DefaultWasmVMContractResponseHandler) Handle(ctx sdk.Context, contractAd
...
+		}
+		k.wasmVM.RemoveCode(info.CodeHash)
```
alpe, line 1233: ah, we are removing code by checksum and not codeID. It is very likely that multiple codeIDs for the same hash exist. We need to ensure that we only delete by a checksum when no instance for any of codeID exists.

alpe (commented):
Very nice work. 🏄 
Unfortunately the task has some more complexity with duplicate code uploads and instances. This was not well documented in the issue. I have added some notes in the review.

I have some feeling that the list of codeIDs is very hard to generate and validate by voters off chain. How about replacing them with a single max codeID that is ensured to be < current `lastCodeId` sequence. We can then iterate through all codeIDs up to this number and build the list of unused codeIs on proposal execution. This is going to be an expensive operation on Gas but it is executed in an end-blocker.

It would be good to collect some numbers how much gas this would cost and how long this takes for some realistic data like Juno mainnet. But this can be a seconds task

proto/cosmwasm/wasm/v1/tx.proto
```diff
@@ -388,4 +391,20 @@ message MsgStoreAndInstantiateContractResponse {
...
+
+// MsgPruneWasmCodes is the MsgPruneWasmCodes request type.
+message MsgPruneWasmCodes {
+  option (amino.name) = "wasm/MsgPruneWasmCodes";
+  option (cosmos.msg.v1.signer) = "authority";
+
+  // Authority is the address of the governance account.
+  string authority = 1 [ (cosmos_proto.scalar) = "cosmos.AddressString" ];
+  // LatestCodeID references the upper limit until unpinned
+  // code ids will be pruned
```
webmaster128, line 404: Could you be explicit here if there upper limit is included or not? "latest code ID" sounds lit it is included but the implementation seems to be <.

I think <= is a bit simpler to reason about.

x/wasm/keeper/keeper.go
```diff
@@ -1213,3 +1218,37 @@ func (h DefaultWasmVMContractResponseHandler) Handle(ctx sdk.Context, contractAd
 	}
 	return result, nil
 }
+
+// PruneWasmCodes deletes code info for unpinned codes smaller than given code id.
+func (k Keeper) PruneWasmCodes(ctx sdk.Context, codeID uint64) error {
+	pinnedCodesHash := make(map[string]struct{}, 0)
+	unpinnedCodes := make([]uint64, 0)
+
+	for c := uint64(1); c < codeID; c++ {
```
webmaster128, line 1227: Should we also add a minimum here? This can be optional. But then you can clean

First proposal: ID <= 1000
Second proposal 1001 <= ID <= 3000
Third proposal 3001 <= ID <= 5000
Fourth proposal 2001 <= ID <= 4000 (let's say some old codes got unpinned or migrated aways from)

x/wasm/keeper/keeper.go
```diff
@@ -1213,3 +1218,37 @@ func (h DefaultWasmVMContractResponseHandler) Handle(ctx sdk.Context, contractAd
...
+		if info == nil {
+			continue
+		}
+
+		if _, found := pinnedCodesHash[hex.Dump(info.CodeHash)]; found {
+			continue
+		}
+
+		k.deleteCodeInfo(ctx, code)
+		_ = k.wasmVM.RemoveCode(info.CodeHash)
```
webmaster128, line 1251: Do we really want to ignore those errors? It typically means a file could not be removed and indicates invalid state.

webmaster128 (commented):
🙌

x/wasm/keeper/keeper.go
webmaster128: Hmm, looking at the other problem regarding deletion by hash I am not sure if this suggestion makes sense. We need to look through all code IDs anyways to know which checksum is used.

x/wasm/keeper/keeper.go
```diff
@@ -1213,3 +1218,37 @@ func (h DefaultWasmVMContractResponseHandler) Handle(ctx sdk.Context, contractAd
...
+		if info == nil {
+			continue
+		}
+
+		if k.IsPinnedCode(ctx, c) {
+			pinnedCodesHash[hex.Dump(info.CodeHash)] = struct{}{}
+			continue
+		}
+		unpinnedCodes = append(unpinnedCodes, c)
+	}
```
webmaster128, line 1238: Could we loop over codes (or code infos) using a DB iterator instead of an integer? Then we don't need to execute the loop body so often when a lot of codes are already pruned.

x/wasm/keeper/keeper.go
alpe: 🤔 we would aways iterate through all contract-infos and pined codes but when we store the max-code-id of the last run, we could abort early when reached. No need to pass it with the proposal IMHO. 
This would be an optimization, let's do it in a second PR

x/wasm/keeper/keeper.go
alpe: It is either abort the whole operation or not. IMHO there is value in removing the code-info data from the chain state even in case of an error here. There may be individual reasons on single machines, like file permissions that must not affect consensus. On the next state sync, the file would not be transferred to new nodes.
Logging the error makes sense IMHO

alpe (commented):
Thanks for the updates! There is some non optimal code when looping over the code-id range and using single get-pinned calls. It may not be very intuitive to work with iterators but they are essential for bulk operations. Please see my code example for inspiration:
```go
// iteratePinnedCodes iterates over all pinned code ids in ascending order
func (k Keeper) iteratePinnedCodes(ctx sdk.Context, cb func(codeID uint64) bool) {
	store := prefix.NewStore(ctx.KVStore(k.storeKey), types.PinnedCodeIndexPrefix)
	iter := store.Iterator(nil, nil)
	defer iter.Close()

	l := len(types.PinnedCodeIndexPrefix)
	for ; iter.Valid(); iter.Next() {
		codeID := sdk.BigEndianToUint64(iter.Key()[l:])
		if cb(codeID) {
			return
		}
	}
}

// PruneWasmCodes iterates over all contracts and code-infos to remove unused codes that are smaller than given code id.
func (k Keeper) PruneWasmCodes(ctx sdk.Context, maxCodeID uint64) error {
	usedCodeIDs := make(map[uint64]struct{})

	// used by contracts
	k.IterateContractInfo(ctx, func(_ sdk.AccAddress, info types.ContractInfo) bool {
		if info.CodeID < maxCodeID {
			usedCodeIDs[info.CodeID] = struct{}{}
		}
		return false
	})
	// pinned to cache by gov
	k.iteratePinnedCodes(ctx, func(codeID uint64) bool {
		usedCodeIDs[codeID] = struct{}{} // may add maxCodeID but does not matter as we check later
		return codeID >= maxCodeID       // abort early
	})

	// execute prune
	done := make(map[string]struct{})
	var deleteCounter int
	k.IterateCodeInfos(ctx, func(codeID uint64, info types.CodeInfo) bool {
		if codeID >= maxCodeID {
			return true // abort early
		}
		checksum := string(info.CodeHash)
		if _, ok := usedCodeIDs[codeID]; ok {
			done[checksum] = struct{}{}
			return false
		}
		k.deleteCodeInfo(ctx, codeID) // clean up state

		if _, ok := done[checksum]; ok { // do not delete twice on disk
			return false
		}

		done[checksum] = struct{}{}
		if err := k.wasmVM.RemoveCode(info.CodeHash); err != nil {
			k.Logger(ctx).Error("failed to delete wasm file on disk", "checksum", info.CodeHash)
		} else {
			deleteCounter++
		}
		return false
	})
	k.Logger(ctx).Info("deleted unused wasm files", "total", deleteCounter)
	return nil
}
```

x/wasm/keeper/keeper.go
```diff
@@ -1213,3 +1218,68 @@ func (h DefaultWasmVMContractResponseHandler) Handle(ctx sdk.Context, contractAd
...
+
+	// collect pinned code ids
+	k.iteratePinnedCodes(ctx, func(codeID uint64) bool {
+		usedCodeIds[codeID] = struct{}{}
+		return false
+	})
+
+	// check if instances are used only by unpinned code ids <= latestCodeID
+	k.IterateCodeInfos(ctx, func(codeID uint64, info types.CodeInfo) bool {
+		if _, ok := usedCodeIds[codeID]; !ok && codeID <= latestCodeID {
```
alpe, line 1260: The `codeID <= latestCodeID` is not correct. Just remove this part that we iterate over the whole set of code infos

For example, there can be code ID 1 and 10 for the same checksum. If only code ID 10 has an instance then a prune with max code id 2 would miss the instance.

x/wasm/keeper/keeper.go
```diff
@@ -1213,3 +1218,68 @@ func (h DefaultWasmVMContractResponseHandler) Handle(ctx sdk.Context, contractAd
...
+			if _, found := usedChecksums[string(info.CodeHash)]; !found {
```
alpe, line 1261: Interesting solution to store state with the checksum. But is it needed? If you don't set anything here, the result would be the same for `if !usedChecksums[string(info.CodeHash)]`. Or do I miss something?

x/wasm/keeper/keeper.go
```diff
@@ -1213,3 +1218,68 @@ func (h DefaultWasmVMContractResponseHandler) Handle(ctx sdk.Context, contractAd
...
+			return false
+		}
+		usedChecksums[string(info.CodeHash)] = true
+		return false
+	})
+
+	// delete all unpinned code ids <= latestCodeID and all the
+	// instances used only by unpinned code ids <= latestCodeID
+	k.IterateCodeInfos(ctx, func(codeID uint64, info types.CodeInfo) bool {
+		if _, ok := usedCodeIds[codeID]; !ok && codeID <= latestCodeID {
```
alpe, line 1273: The conditions make sense here. I would prefer some better naming for `latestCodeID`. Maybe `maxCodeID`? Latest is a moving target for me as there can be new code-infos stored while the gov proposal is in-flight

x/wasm/keeper/keeper.go
```diff
@@ -1213,3 +1218,68 @@ func (h DefaultWasmVMContractResponseHandler) Handle(ctx sdk.Context, contractAd
...
+
+	// check if instances are used only by unpinned code ids <= latestCodeID
+	k.IterateCodeInfos(ctx, func(codeID uint64, info types.CodeInfo) bool {
+		if _, ok := usedCodeIds[codeID]; !ok && codeID <= latestCodeID {
+			if _, found := usedChecksums[string(info.CodeHash)]; !found {
+				usedChecksums[string(info.CodeHash)] = false
+			}
+			return false
+		}
+		usedChecksums[string(info.CodeHash)] = true
```
alpe, line 1266: A `struct{}` would be a more memory efficient value in `usedChecksums`

x/wasm/keeper/keeper.go
pinosu: I iterate over the whole set of code infos. I need this condition to mark as "false" only the checksums with unpinned code ids <= latestCodeID. This helps because we cannot delete checksums with unpinned code ids > latestCodeID.
These will be marked as "true", so as usedChecksums.. same as if they were pinned.
This also answer to the other comment why we need map[string]bool and it is not enough to have map[string]struct{}

pinosu:
I checked logs and it fails for the issue we already raised: empty validator set. 
`Selected randomly generated parameters for simulated genesis:
{
  stake_per_account: "{824677238976}",
  initially_bonded_validators: "0"
}`
So we can ignore it

x/wasm/keeper/wasmtesting/mock_engine.go
```diff
@@ -167,7 +167,7 @@ func (m *MockWasmer) Unpin(checksum wasmvm.Checksum) error {
 }
 
 func (m *MockWasmer) RemoveCode(checksum wasmvm.Checksum) error {
-	if m.UnpinFn == nil {
+	if m.RemoveCodeFn == nil {
```
alpe, line 170: 👍

x/wasm/keeper/bench_test.go
```diff
@@ -100,3 +102,26 @@ func BenchmarkCompilation(b *testing.B) {
 		})
 	}
 }
+
+// Calculate the time it takes to prune about 100k wasm codes
+func BenchmarkPruningWasmCodes(b *testing.B) {
```
alpe, line 107: Good start to get some numbers.
This runs with a mem db, tiny contracts and no duplicate instances. I can imagine disk IO can have some impact here but let's add another DB to compare them

pinosu:
Closing because performance are very bad (See bench tests).

PR #1460, opened 2023-06-23
Fix prune logic

alpe:


pinosu (approved):
Much cleaner solution! Thank you!

PR #1462, opened 2023-06-27
Ensure history position does not conflict

alpe:
Add extra safety net to ensure no overlap in a mixed environment with 20 + 32 byte contract addresses break the position counter
For a single contract not 2 addresses exists, so this not really an issue


pinosu (approved):
LGTM 👍

PR #1464, opened 2023-06-28
Bump seq in ibctests for failed messages as well

alpe:
Follow up from https://github.com/osmosis-labs/mesh-security-sdk/pull/46 for ibctesing


pinosu (approved):
LGTM 👍

PR #1448, opened 2023-06-13
[Do-not-merge] Upgrade to Cosmos-SDK v0.50.

alpe:
🚧 WIP - early version

See #1438

⚠️ All broken currently ! 🤯 

Failing tests are deactivated. Go mod is setup my local workspace. 

I am waiting for a soution in ibc-go that fixes the: `failed consensus state verification for client (07-tendermint-0): chained membership proof failed to verify membership of value:`

See [upgrade doc](https://github.com/cosmos/cosmos-sdk/blob/release/v0.50.x/UPGRADING.md) in the SDK

I have started with just the go.mod dependencies and some related changes to imports.

This blocked now on the ibc-go release that includes the `capability` module and fixed imports for sdk v0.50. 
* Notional is helping the team with a [PR](https://github.com/cosmos/ibc-go/pull/3883) but still core integration of the testing framework does not work for me.
* Either there are some issues with CometBFT or the way how I would use it for the extended ibc-go testing framework. See my [issue](https://github.com/cometbft/cometbft/issues/1001)

This is also blocked on on some modules in the SDK that are not released. See `go.mod` replacement statements for details

Some notes on the work

#### New modules
- circuit

#### Moved to `cosmossdk.io` path
- store
- x/nft
- x/evidence
- x/feegrant
- x/upgrade

#### Others
"github.com/cosmos/cosmos-sdk/snapshots -> "cosmossdk.io/store/snapshots
"github.com/cosmos/cosmos-sdk/client/grpc/tmservice -> "github.com/cosmos/cosmos-sdk/client/grpc/cmtservice
"github.com/cosmos/cosmos-sdk/x/capability -> "github.com/cosmos/ibc-go/v7/modules/capability

Removed: legacy proposals for x/upgrade
### 
- [x] deprecate MakeEncodingConfig
- [x] wasm keeper: accept runtime.NewKVStoreService for keys[wasm.StoreKey],
- [x] handle storeservice.Get/Set errors 
- [ ] remove non requires sdk modules from begin/ endblocker 
- [ ] replace sdk.Context with stdlib context in keeper
- [x] use `[(cosmos_proto.scalar) = "cosmos.AddressString"]` for addresses in proto
- [x] set `cosmos.msg.v1.service` proto annotation
- [x] upgrade ante-handler
- [x] use new collections framework for params

PR #1468, opened 2023-06-28
Add wasm08 client IBC support

Reecepbcups:
In draft until this goes to IBC-go official version (to remove replace statement)

We need to Juno's v47 testnet soon (mesh security as well), so needed to get this ready now


go.mod
```diff
@@ -2,6 +2,11 @@ module github.com/CosmWasm/wasmd
 
 go 1.19
 
+// Uses this fork of the wasm08 client so PFM works
+// Now uses github.com/cosmos/ibc-go/modules/capability/types
+// go get github.com/crodriguezvega/ibc-go/v7@24d07389e19357ff836f59a60b2a5d3897baaa08
+replace github.com/cosmos/ibc-go/v7 => github.com/crodriguezvega/ibc-go/v7 v7.0.0-20230626145354-24d07389e193
```
alpe, line 8: why not use https://github.com/cosmos/ibc-go/releases/tag/v7.2.0 ?

alpe (commented):
Thank you for your PR!
I have quickly skimmed through the changes and I am not sure how this relates to [ics-008](https://github.com/cosmos/ibc/blob/main/spec/client/ics-008-wasm-client/README.md). I would ask to open an issue with details to discuss your feature request. It is usually a nicer way to collaborate than criticize/ defend code. I consider spikes also very helpful in supporting a conversation but expectations should be clear that a spike is for discussion only and not going to be merged.

In this PR I can see that you are using the capability module from ibc-go. That makes sense and is required for the sdk 50 upgrade anyway. I had included this into #1448 already but as it is blocked other projects currently, I would be happy to have this replaced earlier. 

So why not to a PR to use the official ibc-go v7.2.0 release instead? It would make sense to me to replace all sdk capability imports. No need to rename the import aliases `capabilitykeeper`/ `capabilitytypes`. A mixed setup should be avoided. WDYT?

Reecepbcups:
> Thank you for your PR! I have quickly skimmed through the changes and I am not sure how this relates to [ics-008](https://github.com/cosmos/ibc/blob/main/spec/client/ics-008-wasm-client/README.md).

This PR just pulls in the actual spec (We are using the wasm-client working branch on Juno. Without these changes, wasmd will not even compile) - https://github.com/CosmosContracts/juno/pull/680

> I would ask to open an issue with details to discuss your feature request. It is usually a nicer way to collaborate than criticize/ defend code. I consider spikes also very helpful in supporting a conversation but expectations should be clear that a spike is for discussion only and not going to be merged. I would also like to understand how this relates to mesh-security and the planned Juno testnet.

I do agree, I just made it a draft here so others could see the work was already done. I am more than happy to close if we want to wait until it is in ibc-go main.

Juno's testnet will be upgrading to v47 with the wasm-client which required modifications to: Packet Forward Middleware, ASYNC-ICQ, and this wasmd modules to be compatible. This is a requirement for Composable to connect us with Kusama and Polkadot, as well as needed for mesh security per Ethan's comments. 

I don't think it is the best idea for us to add the mesh-security-sdk in this testnet (Juno's Uni-6), best for us to make that its own testnet most likely since it is early stages. I am more than happy to take lead on that. That testnet would also have this wasm-client support too.

> So why not to a PR to use the official ibc-go v7.2.0 release instead?

The wasm-client branch is built off 7.0.0 still right now as it has been in development for a bit. Once their review is done 7.2 will get merged in and bumped

>  It would make sense to me to replace all sdk capability imports. No need to rename the import aliases `capabilitykeeper`/ `capabilitytypes`. A mixed setup should be avoided. WDYT?

Ahh yes this is a good point. I was thinking backwards compatibility with non migrated scoped keepers, but that would never be the case since you HAVE to migrate because of the new forced types. Thanks for this insight!

go.mod
Reecepbcups: The wasm-client branch is built off 7.0.0 still right now as it has been in development for a bit. Once their review is done 7.2 will get merged in and bumped to v7.3+. This is a requirement to get the wasm-client logic in for Juno for now for testnet

> https://github.com/strangelove-ventures/ibc-go/tree/feat/wasm-clients-main

alpe:
Thank you @Reecepbcups for the additional comments!

> I am more than happy to close if we want to wait until it is in ibc-go main.

It is unclear to me when or even if this code will go into IBC-go main. I do not want to risk our main branch to depend on this custom fork. We did long living feature branches in the past for early SDK upgrades or some special features. This comes with quite some extra work to maintain and requires some good motivation and preparation. Usually an issue is started to discuss this with community members giving feedback.

> Once their review is done 7.2 will get merged in and bumped

Not sure about this either. But I can see your problems with the capability module when you update your Juno `go.mod` to the ibc-go fork. It is quite a mess currently. When we upgrade wasmd to ibc-go 7.2 capability, this should be resolved again.

> ... as well as needed for mesh security per Ethan's comments.

Can you point me to the comments? We [discussed](https://github.com/osmosis-labs/mesh-security/blob/main/docs/ibc/Rewards.md#selected-implementation) ADR-8 support but nothing about wasm-08 in context of MS


To move forward, I would add the spike label now and keep the branch for some time. Just to manage expectations: our capacity is very limited, please do not expect reviews on a spike.

alpe:
fyi: I have asked and the capability module will be shipped with the SDK 50 upgrade in IBC-Go. Well get an early alpha or RC tag to start integration.

Reecepbcups:
@alpe 

> It is unclear to me when or even if this code will go into IBC-go main. I do not want to risk our main branch to depend on this custom fork. 

Based off my discussions with Strangelove, they are actively working to get it into IBC go main. Right now it looks like it is targeted for IBC v7.3.X, these commits get it to latest so it works.  I will start an issue once we get it up on Juno's testnet here in the next week or 2 (pending reviews from Core-1 right now)

> as well as needed for mesh security per Ethan's comments
Disregard this comment now sorry. I talked with the Core team and we are going to put the mesh security SDK into its own testnet since this v47 upgrade + wasm client is already quite large. I let them know I will happily manage and spin this up when it comes time

> our capacity is very limited, please do not expect reviews on a spike
Understood, I am just using it here as a placeholder since it is required for our upgrade. I do not expect any reviews as this is bleeding edge implementations

alpe:
We will be working with the IBC-go team on wasm-08 client support in the future.
The code has been moved from stranglove to the ibc-go repo already, afaik

PR #1471, opened 2023-06-29
Remove check for wasm limit size in state sync

### LINKED ISSUE #1467: Accept wasm data from state-sync as valid
The max code size limit must not prevent a legacy contract to be stored when it was accepted before:
Follow up on https://twitter.com/pwnfoo/status/1674052643214307329

pinosu:
Resolves #1467

### DIFF
--- a/x/wasm/ioutils/ioutil.go
+++ b/x/wasm/ioutils/ioutil.go
@@ -11,22 +11,8 @@ import (
 )
 
 // Uncompress expects a valid gzip source to unpack or fails. See IsGzip
-func Uncompress(gzipSrc []byte) ([]byte, error) {
-	zr, err := gzip.NewReader(bytes.NewReader(gzipSrc))
-	if err != nil {
-		return nil, err
-	}
-	zr.Multistream(false)
-	defer zr.Close()
-
-	bz, err := io.ReadAll(zr)
-	return bz, err
-}
-
-// UncompressWithLimit fails if wasm size exceeds the limit.
-// It expects a valid gzip source to unpack or fails. See IsGzip
-func UncompressWithLimit(gzipSrc []byte, limit uint64) ([]byte, error) {
-	if uint64(len(gzipSrc)) > limit {
+func Uncompress(gzipSrc []byte, limit int64) ([]byte, error) {
+	if int64(len(gzipSrc)) > limit {
 		return nil, types.ErrLimit.Wrapf("max %d bytes", limit)
 	}
 	zr, err := gzip.NewReader(bytes.NewReader(gzipSrc))
@@ -35,7 +21,7 @@ func UncompressWithLimit(gzipSrc []byte, limit uint64) ([]byte, error) {
 	}
 	zr.Multistream(false)
 	defer zr.Close()
-	bz, err := io.ReadAll(LimitReader(zr, int64(limit)))
+	bz, err := io.ReadAll(LimitReader(zr, limit))
 	if types.ErrLimit.Is(err) {
 		return nil, errorsmod.Wrapf(err, "max %d bytes", limit)
 	}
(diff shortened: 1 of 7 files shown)


x/wasm/ioutils/ioutil.go
```diff
@@ -11,7 +11,21 @@ import (
...
+	zr.Multistream(false)
+	defer zr.Close()
+
+	bz, err := io.ReadAll(zr)
+	return bz, err
+}
+
+// UncompressWithLimit fails if wasm size exceeds the limit.
+// It expects a valid gzip source to unpack or fails. See IsGzip
+func UncompressWithLimit(gzipSrc []byte, limit uint64) ([]byte, error) {
```
alpe, line 28: This will break api compatibility easily. Let's keep the old name just in case. The param is self explaining

x/wasm/ioutils/ioutil.go
```diff
@@ -11,7 +11,21 @@ import (
 )
 
 // Uncompress expects a valid gzip source to unpack or fails. See IsGzip
-func Uncompress(gzipSrc []byte, limit uint64) ([]byte, error) {
+func Uncompress(gzipSrc []byte) ([]byte, error) {
```
alpe, line 14: No need to copy the code. Instead of adding a new method, you can simply pass `math.MaxUint` as limit. This makes it also more readable on the caller, IMHO.

x/wasm/ioutils/ioutil_test.go
```diff
@@ -23,6 +23,51 @@ func TestUncompress(t *testing.T) {
...
+		t.Run(msg, func(t *testing.T) {
+			r, err := Uncompress(spec.src)
+			require.True(t, errors.Is(spec.expError, err), "exp %v got %+v", spec.expError, err)
+			if spec.expError != nil {
+				return
+			}
+			assert.Equal(t, spec.expResult, r)
+		})
+	}
+}
```
alpe, line 62: good test. But we would need one on the snapshotter to ensure a file > MaxWasmSize is accepted.

alpe (commented):
Good start but let's do less ;-)
An open question for me is the genesis import. Do we want to ensure the `MaxWasmSize` or accept what comes in as in the state sync example? It feels that the cases are very similar. WDYT?

x/wasm/keeper/genesis_test.go
```diff
@@ -133,6 +135,12 @@ func TestGenesisExportImport(t *testing.T) {
 	srcIT := srcCtx.KVStore(wasmKeeper.storeKey).Iterator(nil, nil)
 	dstIT := dstCtx.KVStore(dstKeeper.storeKey).Iterator(nil, nil)
 
+	t.Cleanup(func() {
+		types.MaxWasmSize = 800 * 1024
```
alpe, line 139: Great to do proper housekeeping here! I am a bit afraid that we will easily miss this line whenever the default would be touched in the future. Can you use a temp variable for the original value instead and then use it here? 
like:
```go
origMaxWasmSize := types.MaxWasmSize
types.MaxWasmSize = 1
...
types.MaxWasmSize = origMaxWasmSize
```

x/wasm/keeper/snapshotter_integration_test.go
```diff
@@ -67,6 +67,11 @@ func TestSnapshotter(t *testing.T) {
 			require.NoError(t, err)
 			assert.NotNil(t, snapshot)
 
+			types.MaxWasmSize = 1
+			t.Cleanup(func() {
+				types.MaxWasmSize = 800 * 1024
```
alpe, line 72: As commented before, please use a temp variable to not hard code the value here

alpe (approved):
Great work! Please see my comments on restoring the original size limit. Otherwise 💯 🏄

PR #1473, opened 2023-06-30
Export empty byte code when code hash is empty

inon-man:
## Description
It seems that Terra made the decision to abandon the code of all contracts when upgrading from Columbus-4 to Columbus-5 because of incompatibility. It also removed CodeHash and CodeBytes when exporting the Columbus-4 Genesis to reduce unnecessary space usage. This resulted in the addition of [exception handling](https://github.com/classic-terra/core/blob/v2.0.1/x/wasm/genesis.go#L70C1-L76) to the GenesisExport function of Columbus-5.

The purpose of this PR is to resolve the panic issue that occurs during the export of Genesis in Classic Terra Core v2.1.1 with using CosmWasm/wasmd v0.30.0

Related Issue: https://github.com/classic-terra/core/issues/296

### DIFF
--- a/x/wasm/keeper/genesis.go
+++ b/x/wasm/keeper/genesis.go
@@ -78,9 +78,13 @@ func ExportGenesis(ctx sdk.Context, keeper *Keeper) *types.GenesisState {
 	genState.Params = keeper.GetParams(ctx)
 
 	keeper.IterateCodeInfos(ctx, func(codeID uint64, info types.CodeInfo) bool {
-		bytecode, err := keeper.GetByteCode(ctx, codeID)
-		if err != nil {
-			panic(err)
+		var bytecode []byte
+		var err error
+		if len(info.CodeHash) != 0 {
+			bytecode, err = keeper.GetByteCode(ctx, codeID)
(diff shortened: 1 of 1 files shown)


alpe:
Thank you for your PR!
The changes look straight forward but I am wondering if this is addresses the issue well enough. I can see that the genesis import requires a valid wasm code and ensures that the checksum matches. 
I would like to avoid a scenario where we discover this slice by slice and have conditions in the code that originate from invalid state of a single blockchain.

IMHO the only **sustainable** process is cleaning up the data. I don't have the full context and I also don't know what else was modified for Columbus-5 but I think that invalid code-infos, contract-infos and contract-history should be removed.
You can do this via genesis export/ cleanup/ import , a state migration during a chain upgrade or in the export process similar to the [zero height cleanup](https://github.com/CosmWasm/wasmd/blob/v0.40.1/app/export.go#L50) .
The export process would probably be the fastest method to start with as you fully own that code in your codebase.

Please let me know if this makes sense for you.

alpe:
Closing this now as there is no further conversation

PR #1481, opened 2023-07-04
Remove trace param in system tests

alpe:
Testing system tests without `--trace` param on node start
This resolves the issues with failing tests. Thanks to @julienrbrt for digging into this and the help! 🤗


pinosu (approved):
LGTM 💯 👏

PR #1482, opened 2023-07-04
Gov authorization propagation for sub-messages

### LINKED ISSUE #1207: Gov: support contracts that dynamically instantiate other contracts
Currently, only the first contract call is executed in a gov authorization context. On a restricted chain, submessages would not be able to create instances.

As a simple solution for this feature, gov proposal executions would pass their privileged [GovAuthorizationPolicy](https://github.com/CosmWasm/wasmd/blob/v0.31.0-rc0/x/wasm/keeper/authz_policy.go#L46) to submessages. Without further restrictions, this policy would not only bypass chain restrictions on instantiate but also [admin checks](https://github.com/CosmWasm/wasmd/blob/v0.31.0-rc0/x/wasm/keeper/authz_policy.go#L57-L62) for migrate or access config increasing the risk of dangerous operations.

A more restrictive solution, would consider the original intent of the gov proposal and forward a finer grained authorization. The original intent though is not always easy to spot with sudo/ execute proposals
Current [proposal types](https://github.com/CosmWasm/wasmd/blob/v0.31.0-rc0/proto/cosmwasm/wasm/v1/proposal.proto):

- InstantiateContractProposal  
- InstantiateContract2Proposal
- StoreAndInstantiateContractProposal
- MigrateContractProposal
- SudoContractProposal
- ExecuteContractProposal

alpe:
Resolves #1207 

I have focused on instantiate and migrate to propagate to sub-messages. `Instantiate` is enabled by default as this comes with low risk compared to propagating `migrate` gov permissions.
This can be enabled/ disabled via the `WitGovSubMsgAuthZPropagated` option in the app setup.

The `sudo` call though does not include gov permission propagation for security reasons. It can be enabled for native Go calls by passing an `AutorizationPolicy` with the `sdk.Context`. I have added `WithSubMsgAuthzPolicy` to support this.


x/wasm/keeper/keeper_cgo.go
```diff
@@ -54,7 +54,10 @@ func NewKeeper(
 		gasRegister:          NewDefaultWasmGasRegister(),
 		maxQueryStackSize:    types.DefaultMaxQueryStackSize,
 		acceptedAccountTypes: defaultAcceptedAccountTypes,
-		authority:            authority,
+		propagateGovAuthorization: map[types.AuthorizationPolicyAction]struct{}{
+			types.AuthZActionInstantiate: {},
```
alpe, line 58: Setting the default

x/wasm/keeper/options.go
```diff
@@ -161,6 +161,20 @@ func WithAcceptedAccountTypesOnContractInstantiation(accts ...authtypes.AccountI
 	})
 }
 
+// WitGovSubMsgAuthZPropagated overwrites the default gov authorization policy for sub-messages
+func WitGovSubMsgAuthZPropagated(entries ...types.AuthorizationPolicyAction) Option {
```
alpe, line 165: Constructor option to modify defaults

x/wasm/keeper/msg_server.go
```diff
@@ -357,9 +357,12 @@ func (m msgServer) StoreAndInstantiateContract(goCtx context.Context, req *types
 	}, nil
 }
 
-func (m msgServer) selectAuthorizationPolicy(actor string) AuthorizationPolicy {
+func (m msgServer) selectAuthorizationPolicy(ctx sdk.Context, actor string) types.AuthorizationPolicy {
 	if actor == m.keeper.GetAuthority() {
-		return GovAuthorizationPolicy{}
+		return newGovAuthorizationPolicy(m.keeper.propagateGovAuthorization)
+	}
+	if policy, ok := types.SubMsgAuthzPolicy(ctx); ok {
```
alpe, line 364: Reads policy from context when set

x/wasm/keeper/keeper.go
```diff
@@ -356,6 +350,7 @@ func (k Keeper) instantiate(
 		sdk.NewAttribute(types.AttributeKeyCodeID, strconv.FormatUint(codeID, 10)),
 	))
 
+	ctx = types.WithSubMsgAuthzPolicy(ctx, authPolicy.SubMessageAuthorizationPolicy(types.AuthZActionInstantiate))
```
alpe, line 353: Pass sub-message policy with context

x/wasm/keeper/submsg_test.go
```diff
@@ -550,3 +557,157 @@ func TestDispatchSubMsgConditionalReplyOn(t *testing.T) {
 		})
 	}
 }
+
+func TestInstantiateGovSubMsgAuthzPropagated(t *testing.T) {
+	mockWasmVM := &wasmtesting.MockWasmer{}
```
pinosu, line 562: Very nice tests!

pinosu (approved):
Great work 🥇 LGTM 👍

PR #1484, opened 2023-07-05
Deprecate alias.go

### LINKED ISSUE #1456: Deprecate alias.go
Deprecate everything in /x/wasm/alias.go to get rid of this file in a future version.

This task includes removing all references to types in this file in /app and other packages

pinosu:
Resolves #1456


alpe (commented):
Thanks for taking this tasks. A lot of work 🥇 
Can you bring back the `alias.go` and make everything deprecated there? We would delete it with the next version. 
Some people may use wamd as a library and this would help them with a softer transition. 🙏

alpe (approved):
Not very rewarding but important work! Thank you for doing this!

PR #1486, opened 2023-07-05
Upgrade to wasmvm 1.3.0-rc.0

pinosu:
Part of #1296 

- [x] use new checksum [function](https://github.com/CosmWasm/wasmvm/pull/410) in gov_tx.go instead of `sha256.Sum256(raw)`; check for other places
- [x] Update go.mod
- [x] Update download links and checksum in Dockerfile
- [x] Update compatibility [matrix](https://github.com/CosmWasm/wasmd/#compatibility)
- [x] Use [StoreCodeUnchecked](https://github.com/CosmWasm/wasmvm/issues/434) in state sync
- [x] Deprecate wasmVM.Create and use wasmVM.StoreCode instead


x/wasm/keeper/keeper.go
```diff
@@ -217,7 +217,7 @@ func (k Keeper) importCode(ctx sdk.Context, codeID uint64, codeInfo types.CodeIn
 			return types.ErrCreateFailed.Wrap(errorsmod.Wrap(err, "uncompress wasm archive").Error())
 		}
 	}
-	newCodeHash, err := k.wasmVM.Create(wasmCode)
+	newCodeHash, err := k.wasmVM.StoreCodeUnchecked(wasmCode)
```
alpe, line 220: nice!

x/wasm/keeper/snapshotter.go
```diff
@@ -106,7 +106,7 @@ func restoreV1(_ sdk.Context, k *Keeper, compressedCode []byte) error {
 	}
 
 	// FIXME: check which codeIDs the checksum matches??
-	_, err = k.wasmVM.Create(wasmCode)
+	_, err = k.wasmVM.StoreCodeUnchecked(wasmCode)
```
alpe, line 109: 👍

x/wasm/keeper/wasmtesting/mock_engine.go
```diff
@@ -88,6 +89,20 @@ func (m *MockWasmer) Create(codeID wasmvm.WasmCode) (wasmvm.Checksum, error) {
 	return m.CreateFn(codeID)
 }
 
+func (m *MockWasmer) StoreCode(codeID wasmvm.WasmCode) (wasmvm.Checksum, error) {
```
alpe, line 92: I was wondering if we want to use `Create` at all now in this mock. The method is required to be compatible with the interface but why not always panic with "Deprecated: use StoreCode instead"? There should not be any reason to use it in tests anymore. WDYT?

Please add a deprecated annotation to the method, too.

x/wasm/types/test_fixtures.go
```diff
@@ -2,14 +2,18 @@ package types
...
 	"encoding/hex"
 	"encoding/json"
 	"math/rand"
 
+	wasmvm "github.com/CosmWasm/wasmvm"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
+//go:embed testdata/reflect.wasm
+var wasmCode []byte
```
alpe, line 15: it would be good to rename this to `reflectWasmCode` or something more verbose. There are some contracts used in this project

x/wasm/types/wasmer_engine.go
```diff
@@ -19,8 +19,24 @@ type WasmerEngine interface {
...
 
+	// Create will compile the wasm code, and store the resulting pre-compile
+	// as well as the original code. Both can be referenced later via CodeID
+	// This must be done one time for given code, after which it can be
+	// instatitated many times, and each instance called many times.
+	// It does the same as StoreCodeUnchecked plus the static checks.
+	StoreCode(code wasmvm.WasmCode) (wasmvm.Checksum, error)
+
+	// Create will compile the wasm code, and store the resulting pre-compile
+	// as well as the original code. Both can be referenced later via CodeID
```
alpe, line 34: nit: here and above
```suggestion
	// as well as the original code. Both can be referenced later via checksum
```

alpe (approved):
Great work! Thanks a lot for taking this task. 💐 
I have added some comment about the test mock to discuss. Otherwise excellent!

x/wasm/keeper/wasmtesting/mock_engine.go
pinosu: Good idea!

webmaster128:
LGTM. What iss missing for 1.3 is support for the new message and query types: https://github.com/CosmWasm/wasmd/issues/1296#issuecomment-1621509531

PR #1494, opened 2023-07-10
Allow better override of wasmVM in x/wasm keeper

ethanfrey:
The current code always creates one VM and then possibly overrides it with the options. However, creating the wasmVm will also create directories and initialize Rust structs that are not cleanly deleted. It would be better to never create it in the first place if it is overridden later.

This PR pushes the initialization towards the end of NewKeeper for this. A concrete use-case is in creating a wasmvm.VM externally and passing it in for more control, which will likely be necessary with ibc wasm light client.

### DIFF
--- a/x/wasm/keeper/keeper_cgo.go
+++ b/x/wasm/keeper/keeper_cgo.go
@@ -35,15 +35,11 @@ func NewKeeper(
 	authority string,
 	opts ...Option,
 ) Keeper {
-	wasmer, err := wasmvm.NewVM(filepath.Join(homeDir, "wasm"), availableCapabilities, contractMemoryLimit, wasmConfig.ContractDebugMode, wasmConfig.MemoryCacheSize)
-	if err != nil {
-		panic(err)
-	}
 
 	keeper := &Keeper{
 		storeKey:             storeKey,
 		cdc:                  cdc,
-		wasmVM:               wasmer,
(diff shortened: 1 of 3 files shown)


x/wasm/keeper/keeper_cgo.go
```diff
@@ -63,6 +59,16 @@ func NewKeeper(
 	for _, o := range opts {
 		o.apply(keeper)
 	}
+	// only set the wasmvm if no one set this in the options
+	// NewVM does a lot, so better not to create it and silently drop it.
```
alpe, line 63: Thanks for bringing this up! Unfortunately, the implementation conflicts with the [`WithWasmEngineDecorator`](https://github.com/CosmWasm/wasmd/blob/v0.41.0-rc.0/x/wasm/keeper/options.go#L28) now that requires the object to exist. 
I will come up with some solution

faddat:
Hey guys, I want to mention -- I susepct that some of this is actually addressed by moving to wazero, but I don't really have a decent starting place. 

The thinking is like this:

* cgo bad
* type conversions error prone

therefore move to a pure go way of running the wasm.

alpe:
@faddat thanks for the head up! The use case Ethan wants to solve is:
>  creating a wasmvm.VM externally and passing it in ... which will likely be necessary with ibc wasm light client.

PR #1503, opened 2023-07-14
Make system tests extendable for other app binaries

alpe:
Support other binaries via parameter
+ less noise on test error output


pinosu (approved):
Nice work! 💯 LGTM 👍

PR #1505, opened 2023-07-16
Add a method to do raw query in ibctesting

larry0x:
This may be useful to check things such as cw2 version info.

I use it in my project: https://github.com/larry0x/ics999/blob/main/tests/suite_test.go#L361-L370


alpe (approved):
Very Nice! Thanks for adding this 💐

PR #1509, opened 2023-07-17
update golang docker

faddat:
- update go version used in dockerfile

I thought that the lint PR was merged, but, maybe not?  Seems this has some linting changes (afaik these are specific to the 50 upgrade)  If you want only the dockerfile changes, 

* https://github.com/CosmWasm/wasmd/pull/1509/commits/5819ece46d6f97148adfac780c26f43122274ecd

is the commit for you :)

It also rebuilds the .proto files


faddat:
Cool :)

I'll keep trucking on ibc-go, since I'm not too sure of the scope of 0.41

Thanks for the feedback!

PR #1526, opened 2023-07-20
Revisit circleCI workflow for tags

alpe:
The docker tag and release step must not depend on other steps that are not enabled for tags.
No need to re-run them as people would not tag a broken build ...


pinosu (approved):
LGTM 👍

PR #1531, opened 2023-07-21
[WIP] SDK 50 beta 0 upgrade

alpe:
🚧 Early version

After #1528 

Requires compatible IBC Go version to move on

PR #1532, opened 2023-07-24
bump versions to the working ibc8

faddat:
This PR bumps ibc-go to a version that passes every test.  

This should resolve the blockage in #1531.

It also seems to pass all tests when running `make test` which is pretty great, too. 

## Note:

I'm still not really sure how to fix the Dockerfile.


faddat:
I think this was buggy.

PR #1537, opened 2023-07-28
style: lint wasmd in the same manner as cosmos-sdk

faddat:
This PR updates linter settings in wasmd so that they exactly match those in the cosmos-sdk so that the style is 
more consistent and less error prone.


pinosu (approved):
Thank you for your contribution! LGTM 👍

PR #1539, opened 2023-08-02
Add StoreAndMigrateContract proposal

### LINKED ISSUE #1463: StoreAndMigrate
Following `StoreAndInstantiate` see https://github.com/CosmWasm/wasmd/issues/785, I think it becomes very useful to do the same for migrate operations since that also requires 2 governance cycles.

pinosu:
Resolves #1463


proto/cosmwasm/wasm/v1/tx.proto
```diff
@@ -430,4 +434,29 @@ message MsgRemoveCodeUploadParamsAddresses {
 
 // MsgRemoveCodeUploadParamsAddressesResponse defines the response
 // structure for executing a MsgRemoveCodeUploadParamsAddresses message.
-message MsgRemoveCodeUploadParamsAddressesResponse {}
\ No newline at end of file
+message MsgRemoveCodeUploadParamsAddressesResponse {}
+
+// MsgStoreAndMigrateContract is the MsgStoreAndMigrateContract
+// request type.
+message MsgStoreAndMigrateContract {
```
alpe, line 441: let's add a `// Since: 0.42` comment here and for the response

x/wasm/client/cli/gov_tx.go
```diff
@@ -928,3 +928,55 @@ func getProposalInfo(cmd *cobra.Command) (client.Context, string, string, sdk.Co
 
 	return clientCtx, proposalTitle, summary, deposit, nil
 }
+
+func ProposalStoreAndMigrateContractCmd() *cobra.Command {
```
alpe, line 932: the command was not added to `SubmitProposalCmd()`

alpe (commented):
The changes look good. Please add the command so that it is available on the CLI.
You can add some more data to the response but other than this, great work.

x/wasm/keeper/msg_server.go
```diff
@@ -469,6 +469,8 @@ func (m msgServer) StoreAndMigrateContract(goCtx context.Context, req *types.Msg
 	}
 
 	return &types.MsgStoreAndMigrateContractResponse{
-		Data: data,
+		CodeID:   codeID,
+		Checksum: checksum,
+		Data:     data,
```
alpe, line 474: Thanks for adding!

alpe (approved):
Good work! 🥇

PR #1546, opened 2023-08-04
Diff reduction: run linter autofixes

faddat:
The linter produced much more diff than actually exists.  

This PR runs golangci-lint run ./... against the v50 branch, so that earlier PRs I made like:

* #1544
* #1453

Don't seem like they contain excessively large chainsets. 

It's also meant to keep things easy to review :).

-Jacob


faddat:
I don't know why the test system works here, but I'll take it.

Issue #1439, opened 2023-06-08
Split project into separate Go modules


alpe:
Use https://github.com/osmosis-labs/mesh-security-sdk as a template

PR #1547, opened 2023-08-04
update system tests

faddat:
This PR updates the system tests in wasmd to:

* use go 1.20
* use the sdk's math library
* use a go workspace so that they're easier to work with


pinosu:
Thanks for your contribution! It looks great 💯 
This work is part of #1439, planned for next milestone so we will keep it on hold for the moment.

alpe:
Closing this now as we moved on in the codebase already. I have applied your changes for `sdk.NewInt`

PR #1553, opened 2023-08-11
Separate types into separate module

### LINKED ISSUE #1552: Separate types into separate package
I'd like to use just the type definitions and compiled proto files.  My team is running into recurring issues by depending on the full wasmd package due to the use of the precompiled shared library.  We'd like to deploy to lightweight cloud environments like Google App Engine, but due to the precompiled share lib, it's a no go to use wasmd as a dependency.  

However, if we can split x/wasm/types into a separate package, problem is solved for us.  Some of the imported wasmvm types would also need to be moved.

Would this project accept a PR to make the split?

conorpp:
My team have been working on a custody solution and have made great use of wasmd to integrate a rust based MPC sidecar into our custody chain :). Really appreciate this project and hope we can contribute.  

One recurring pain point we've had is running into issues with the precompiled shared object when we don't really need it.  Often this is just when we want to _query_ cosmos chains that are using wasmd and act purely as a client.  So we only need the types.

If we can separate the types into a separate package, that would totally make building our client software much easier.

I've done this in 3 steps.
1. Move all of the types using wasmvm into their own package, now `wasmd/x/vmtypes`
2. Rename all affected references to use `vmtypes.` instead of `types.`
3. Create a go.mod and go.sum file in `wasmd/x/types`.

Please let me know what else you'd like to see if anything to get this merged.  Fixes #1552


conorpp:
Can just set `CGO_ENABLED=0` which will remove libvmwasm dependency.

PR #1555, opened 2023-08-14
Fix label validation error

### LINKED ISSUE #1554: Instantiate Label validation error is misleading
When the instantiate msg's label is validated, the [error it returns is thrown away](https://github.com/CosmWasm/wasmd/blob/4611d6daf3afd986af2f9dfee6d87c4037471670/x/wasm/types/tx.go#L105-L107) and instead a "label is required" is thrown.
This is confusing for the contract author, since they provided a label, but an invalid one, with no indication of what is wrong with it. This is the case for `Instantiate`, `Instantiate2` and `StoreAndInstantiate`.

I think the error from `ValidateLabel` should be wrapped / returned instead.

[Original issue](https://github.com/CosmWasm/cosmwasm/issues/1810)

pinosu:
Resolves #1554

### DIFF
--- a/x/wasm/types/tx.go
+++ b/x/wasm/types/tx.go
@@ -104,7 +104,7 @@ func (msg MsgInstantiateContract) ValidateBasic() error {
 	}
 
 	if err := ValidateLabel(msg.Label); err != nil {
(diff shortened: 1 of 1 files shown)


chipshort (approved):
LGTM

alpe (approved):
LGTM 👍

Issue #1536, opened 2023-07-28
[Bug] The wasmd querier GasConsumed() is not adaptive wasmvm, maybe cause an attack。


lyh169:
1. in wasmd
when Execute, need generate the querier.
```
	// prepare querier
	querier := k.newQueryHandler(ctx, contractAddress)
	gas := k.runtimeGasForContract(ctx)
	res, gasUsed, execErr := k.wasmVM.Execute(codeInfo.CodeHash, env, info, msg, prefixStore, cosmwasmAPI, querier, k.gasMeter(ctx), gas, costJSONDeserialization)
	k.consumeRuntimeGas(ctx, gasUsed)
	if execErr != nil {
		return nil, errorsmod.Wrap(types.ErrExecuteFailed, execErr.Error())
	}

```
the newQueryHandler generate a object **QueryHandler** that implement the **Querier** interface as follow
```
type Querier interface {
	Query(request QueryRequest, gasLimit uint64) ([]byte, error)
	GasConsumed() uint64
}
```
the **QueryHandler** Gasconsumed is as follow.
```
func (q QueryHandler) GasConsumed() uint64 {
	return q.Ctx.GasMeter().GasConsumed()
}
```

2. But in wasmvm the cQueryExternal get the GasConsumed as follow, 
```
func cQueryExternal(ptr *C.querier_t, gasLimit cu64, usedGas *cu64, request C.U8SliceView, result *C.UnmanagedVector, errOut *C.UnmanagedVector) (ret C.GoError) {
	defer recoverPanic(&ret)
        
........
	// query the data
	querier := *(*Querier)(unsafe.Pointer(ptr))
	req := copyU8Slice(request)

	gasBefore := querier.GasConsumed()
	res := types.RustQuery(querier, req, uint64(gasLimit))
	gasAfter := querier.GasConsumed()
	*usedGas = (cu64)(gasAfter - gasBefore)
........
	return C.GoError_None
}
```
the *usedGas will be acquired by wasmvm/libwasmvm/src/querier.rs/query_raw, as follow
 ```fn query_raw(
        &self,
        request: &[u8],
        gas_limit: u64,
    ) -> BackendResult<SystemResult<ContractResult<Binary>>> {
        let mut output = UnmanagedVector::default();
        let mut error_msg = UnmanagedVector::default();
        let mut used_gas = 0_u64;
        let go_result: GoError = (self.vtable.query_external)(
            self.state,
            gas_limit,
[...]

PR #1567, opened 2023-08-22
Fix gas calculation

chipshort:
closes #1536

PR #1568, opened 2023-08-22
[Do not merge] Reproduction of #1536

chipshort:
Just added a few assertions to show the problem.
If we want to merge this at some point, we probably want to remove them and create separate tests instead (but that requires some more changes).
Currently uses an unreleased version of wasmvm to get access to the gas data.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/CosmWasm/wasmd
 go 1.20
 
 require (
-	github.com/CosmWasm/wasmvm v1.3.0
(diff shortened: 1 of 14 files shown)


webmaster128:
Great stuff! Did you check if the fix from #1567 solves the problem?

Looking at the diff I doubt we should merge https://github.com/CosmWasm/wasmvm/pull/448. It just makes the API more complex and exposes internals to wasmd that we don't need there for the happy path.

chipshort:
Closing this, as the fix is merged and we won't merge this

Issue #1574, opened 2023-08-28
Panic because WasmVM cache metrics collector cannot be registered properly


dadamu:
## Description

WasmVM metrics now is not registered properly and it occurs panics when querying the `/metrics` endpoint. The panic message is as follows:

```bash
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x1f5b9dd]

goroutine 148 [running]:
github.com/CosmWasm/wasmd/x/wasm/keeper.(*WasmVMMetricsCollector).Collect(0xc0009fa660, 0xc000d25f60?)
	github.com/CosmWasm/wasmd/x/wasm/keeper/metrics.go:56 +0x3d
github.com/prometheus/client_golang/prometheus.(*Registry).Gather.func1()
	github.com/prometheus/client_golang@v1.16.0/prometheus/registry.go:455 +0x10d
created by github.com/prometheus/client_golang/prometheus.(*Registry).Gather
	github.com/prometheus/client_golang@v1.16.0/prometheus/registry.go:547 +0xbac
```

## How to repoduce

1. setup `wasmd` testnet
2. enable `instrumentation.prometheus` inside `~/.wasmd/config/config.toml`
3. enable `telemetry.enabled` inside `~/wasmd/config/app.toml`
4. start wasmd localnet by `wasmd start`
5. query `/metrics` by `curl localhost:26660/metrics`
6. node panics

## How to fix

Fix `WithVMCacheMetrics` to use `postOptsFn` instead of `optsFn` to make sure wasmVM has already been setup.

```github
func WithVMCacheMetrics(r prometheus.Registerer) Option {
-	  return optsFn(func(k *Keeper) {
+	  return postOptsFn(func(k *Keeper) {
		NewWasmVMMetricsCollector(k.wasmVM).Register(r)
	})
}
```

https://github.com/CosmWasm/wasmd/blob/main/x/wasm/keeper/options.go#L129

PR #1575, opened 2023-08-28
fix: make sure wasmvm cache metrics collector registration after VM setup

dadamu:
Close: #1574

This PR intends to fix the bug mentioned by #1574, to make sure wasmvm cache metrics collector registration after VM setup.


alpe (approved):
Thanks a lot for reporting and fixing this!
This option was missed during the refactoring. I will add a unit test so that we have this covered in the future!

I have to apologize for the issue!

PR #1576, opened 2023-08-28
Fail fast on empty wasmvm cache metric setup

alpe:
Follow up on #1574 + #1575

Test and panic to be on the safe side


pinosu (approved):
LGTM 👍

PR #1582, opened 2023-08-29
Add unit test for gas calculation

pinosu:


alpe (approved):
Thanks for adding the tests! 🏄

PR #1585, opened 2023-08-30
Rename key store key attribute to be more concrete

### LINKED ISSUE #1182: Rename `lastIDKey` key
The name can be missleading. In autoincrementID() it is actually the next id key but there are other places to revisit, too

alpe:
Resolves #1182


pinosu (approved):
LGTM 👍

PR #1591, opened 2023-08-31
Add store code authz

### LINKED ISSUE #1584: Custom StoreCode Authorization for authz module
Currently signing store code transactions with multisigs can only be done if you use non-ledger keys due to the memory limit on such devices it makes impossible for signers and for this reason I think it would be very useful to have a contract upload authorization for authz. 

- Authorize grantee to upload any contract (although this can already be achieved by the generic grant authorization)
- Authorize grantee to upload any contract from a list of given hashes  so the granter can verify and be confident the grantee only uploads pre-approved contracts, 

This can even  be useful for DAOs or CW4 multisigs giving access to an uploader by creating a custom authorization.

pinosu:
Resolves #1584


alpe (commented):
Good start! I have added some quick feedback

x/wasm/types/authz.go
```diff
@@ -16,12 +18,92 @@ import (
...
+// MsgTypeURL implements Authorization.MsgTypeURL.
+func (a StoreCodeAuthorization) MsgTypeURL() string {
+	return sdk.MsgTypeURL(&MsgStoreCode{})
+}
+
+// Accept implements Authorization.Accept.
+func (a *StoreCodeAuthorization) Accept(ctx sdk.Context, msg sdk.Msg) (authztypes.AcceptResponse, error) {
+	var code []byte
+	var permission AccessConfig
+	switch msg := msg.(type) {
```
alpe, line 44: pp: a switch can be very powerful but when used only for a single condition, it is not very readable to me. How about using an ordinary `if` as guard instead?
```go
	storeMsg, ok := msg.(*MsgStoreCode)
	if !ok{
		return authztypes.AcceptResponse{}, sdkerrors.ErrInvalidRequest.Wrap("unknown msg type")
	}
	code := storeMsg.WASMByteCode
	permission := *storeMsg.InstantiatePermission
```

x/wasm/types/authz.go
```diff
@@ -16,12 +18,92 @@ import (
...
+	case *MsgStoreCode:
+		code = msg.WASMByteCode
+		permission = *msg.InstantiatePermission
+	default:
+		return authztypes.AcceptResponse{}, sdkerrors.ErrInvalidRequest.Wrap("unknown msg type")
+	}
+
+	checksum, err := wasmvm.CreateChecksum(code)
```
alpe, line 52: the binary in `code` is likely zipped. You need to check and unpack as we do in keeper.create before getting the checksum. This would include charging the gas

x/wasm/types/authz.go
```diff
@@ -16,12 +18,92 @@ import (
...
+	if err != nil {
+		return authztypes.AcceptResponse{}, sdkerrors.ErrInvalidRequest.Wrap("checksum")
```
alpe, line 54: 👍

x/wasm/types/authz.go
```diff
@@ -16,12 +18,92 @@ import (
...
+	}
+
+	for _, grant := range a.Grants {
+		if grant.Accept(checksum, permission) {
+			return authztypes.AcceptResponse{Accept: true}, nil
+		}
+	}
+	return authztypes.AcceptResponse{Accept: false}, nil
```
alpe, line 62: When there is no grant, then it is rejected. Makes sense!

x/wasm/types/authz.go
```diff
@@ -16,12 +18,92 @@ import (
...
+}
+
+// ValidateBasic implements Authorization.ValidateBasic.
+func (a StoreCodeAuthorization) ValidateBasic() error {
+	if len(a.Grants) == 0 {
```
alpe, line 67: Grants can not be empty. Looks good to me

x/wasm/types/authz.go
```diff
@@ -16,12 +18,92 @@ import (
...
+		if err := g.InstantiatePermission.ValidateBasic(); err != nil {
+			return errorsmod.Wrap(err, "instantiate permission")
+		}
+	}
+	return nil
+}
+
+// Accept checks if checksum and permission match the grant
+func (g CodeGrant) Accept(checksum []byte, permission AccessConfig) bool {
+	if !bytes.Equal(g.CodeHash, []byte("*")) && !bytes.Equal(g.CodeHash, checksum) {
```
alpe, line 101: pp: If you would extract `*` to a constant with verbose naming like: `codehashWildcard`, then it would be more readable.  Just place it before the `if` if you don't to make it global

x/wasm/types/authz.go
```diff
@@ -16,12 +18,92 @@ import (
...
+		return false
+	}
+	return permission.IsSubset(*g.InstantiatePermission)
```
alpe, line 104: This part is not fully clear to me. `InstantiatePermission` is optional in the grant. When nothing is set, then I would expect this to be a wildcard, which allows any permission. WDYT? 
Please update the doc on the `CodeGrant` type to be very specific on the cases

x/wasm/types/authz_test.go
```diff
@@ -744,3 +745,184 @@ func mustGrant(contract sdk.AccAddress, limit ContractAuthzLimitX, filter Contra
...
+		"nil code hash": {
+			codeHash:              nil,
+			instantiatePermission: AllowEverybody,
+			expErr:                true,
+		},
+		"invalid permission": {
+			codeHash:              []byte("ABC"),
+			instantiatePermission: AccessConfig{Permission: AccessTypeUnspecified},
+			expErr:                true,
+		},
```
alpe, line 773: good tests. Please add a case with empty permission so that we have this covered for regression.

x/wasm/types/authz_test.go
```diff
@@ -744,3 +745,184 @@ func mustGrant(contract sdk.AccAddress, limit ContractAuthzLimitX, filter Contra
...
 }
+
+func TestValidateCodeGrant(t *testing.T) {
+	specs := map[string]struct {
+		codeHash              []byte
+		instantiatePermission AccessConfig
+		expErr                bool
+	}{
+		"all good": {
+			codeHash:              []byte("ABC"),
```
alpe, line 756: nit: `"ABC"` is ok but if you want to be very verbose, let's make this something like `"any_valid_checksum"` (here and other placs)

x/wasm/types/authz_test.go
```diff
@@ -744,3 +745,184 @@ func mustGrant(contract sdk.AccAddress, limit ContractAuthzLimitX, filter Contra
...
+	invalidGrant, err := NewCodeGrant(nil, AllowEverybody)
+	require.NoError(t, err)
+
+	specs := map[string]struct {
+		setup  func(t *testing.T) validatable
+		expErr bool
+	}{
+		"all good": {
+			setup: func(t *testing.T) validatable {
+				t.Helper()
```
alpe, line 802: I found this `t.Helper()` not very helpful in table tests therefore I [removed](https://github.com/CosmWasm/wasmd/commit/086c052c1820f9330eb9bef9d785e9963312434c) the linter. Let's keep `setup` functions to the minimum

x/wasm/types/authz_test.go
```diff
@@ -744,3 +745,184 @@ func mustGrant(contract sdk.AccAddress, limit ContractAuthzLimitX, filter Contra
...
+			},
+			expErr: true,
+		},
+		"empty grants": {
+			setup: func(t *testing.T) validatable {
+				t.Helper()
+				return NewStoreCodeAuthorization()
+			},
+			expErr: true,
+		},
```
alpe, line 825: Duplicate grants can be more complex. Please add more cases (and doc) to ensure that this is well defined:
` *, AllowEverybody` and `ABC, AllowEverybody` must not be allowed for example
Not sure about `ABC, AllowEverybody` and `ABC, unset` or other cases where only the accessConfig differs.
There is also the double wildcard with `*,unset` to consider

x/wasm/types/authz.go
```diff
@@ -16,12 +18,92 @@ import (
...
+func NewCodeGrant(codeHash []byte, instantiatePermission AccessConfig) (*CodeGrant, error) {
+	return &CodeGrant{
+		CodeHash:              codeHash,
+		InstantiatePermission: &instantiatePermission,
+	}, nil
+}
+
+// ValidateBasic validates the grant
+func (g CodeGrant) ValidateBasic() error {
+	if len(g.CodeHash) == 0 {
```
alpe, line 88: I am often confused if a hash is upper/lowercase. Do you want to enforce a rule here or normalize it when comparing? the `GetChecksum` returns lower case.
Please add a test that ensures this works as expected

x/wasm/types/authz_test.go
```diff
@@ -744,3 +745,184 @@ func mustGrant(contract sdk.AccAddress, limit ContractAuthzLimitX, filter Contra
...
+				CodeID:   1,
+				Msg:      []byte(`{"foo":"bar"}`),
+			},
+			expResult: authztypes.AcceptResponse{
+				Accept: false,
+			},
+			expErr: sdkerrors.ErrInvalidRequest,
+		},
+	}
+	for name, spec := range specs {
```
alpe, line 916: 👍 good test cases

x/wasm/types/codec.go
```diff
@@ -56,6 +56,7 @@ func RegisterLegacyAminoCodec(cdc *codec.LegacyAmino) {
 	cdc.RegisterConcrete(&MaxFundsLimit{}, "wasm/MaxFundsLimit", nil)
 	cdc.RegisterConcrete(&CombinedLimit{}, "wasm/CombinedLimit", nil)
 
+	cdc.RegisterConcrete(&StoreCodeAuthorization{}, "wasm/StoreCodeAuthorization", nil)
```
alpe, line 59: 👍 important to register

alpe (commented):
Thanks for the updates. Good progress!
I have added some nits but also comments to spec out some edge cases.

x/wasm/ioutils/ioutil.go
```diff
@@ -6,14 +6,14 @@ import (
 	"io"
 
 	errorsmod "cosmossdk.io/errors"
-
-	"github.com/CosmWasm/wasmd/x/wasm/types"
 )
 
+var errLimit = errorsmod.Register("wasm", 29, "exceeds limit")
```
alpe, line 11: let's not register a new error type outside of types/errors . You can just make this a regular stdlib error. In the method call you can check for the concrete type if needed and use a registered type.  Please make sure that this error type is public

x/wasm/types/authz.go
```diff
@@ -39,16 +43,23 @@ func (a StoreCodeAuthorization) MsgTypeURL() string {
...
 	}
 
+	code := storeMsg.WASMByteCode
+	permission := *storeMsg.InstantiatePermission
+
+	if ioutils.IsGzip(code) {
+		// TODO: consume gas
+		wasmCode, err := ioutils.Uncompress(code, int64(MaxWasmSize))
+		if err != nil {
+			return authztypes.AcceptResponse{}, sdkerrors.ErrInvalidRequest.Wrap(errorsmod.Wrap(err, "uncompress wasm archive").Error())
```
alpe, line 58: This line would work with a Go stdlib error without a change. The error message is wrapped into a `ErrInvalidRequest` type. The registered `errLimit` namespace/code is ignored anyway.

x/wasm/types/authz.go
```diff
@@ -67,11 +78,20 @@ func (a StoreCodeAuthorization) ValidateBasic() error {
...
-		if err := v.ValidateBasic(); err != nil {
+	for i := 0; i < len(a.Grants); i++ {
+		if strings.EqualFold(string(a.Grants[i].CodeHash), codehashWildcard) && len(a.Grants) > 1 {
+			return sdkerrors.ErrInvalidRequest.Wrap("cannot have multiple grants when wildcard grant is one of them")
+		}
+		if err := a.Grants[i].ValidateBasic(); err != nil {
 			return errorsmod.Wrapf(err, "position %d", i)
 		}
+		for j := i + 1; j < len(a.Grants); j++ {
+			if strings.EqualFold(string(a.Grants[i].CodeHash), string(a.Grants[i].CodeHash)) {
```
alpe, line 89: This is always true. One of the grants should be found by `j`. 
Alternatively, you can store the codehashes in a map and compare the len to be equal len(grants)

x/wasm/types/authz.go
```diff
@@ -15,7 +16,10 @@ import (
 	authztypes "github.com/cosmos/cosmos-sdk/x/authz"
 )
 
-const gasDeserializationCostPerByte = uint64(1)
+const (
+	gasDeserializationCostPerByte = uint64(1)
+	codehashWildcard              = "*"
```
alpe, line 21: let's make the wildcard public so that it can be used from other packages if needed

alpe (commented):
Good progress!

x/wasm/ioutils/ioutil.go
```diff
@@ -3,17 +3,16 @@ package ioutils
...
 	"compress/gzip"
+	"fmt"
 	"io"
-
-	errorsmod "cosmossdk.io/errors"
-
-	"github.com/CosmWasm/wasmd/x/wasm/types"
 )
 
+var errLimit = fmt.Errorf("exceeds limit")
```
alpe, line 10: Nit: as there is no params to fmt you can use:
```suggestion
var errLimit = errors.New("exceeds limit")
```

x/wasm/ioutils/ioutil.go
```diff
@@ -3,17 +3,16 @@ package ioutils
...
+
 // Uncompress expects a valid gzip source to unpack or fails. See IsGzip
 func Uncompress(gzipSrc []byte, limit int64) ([]byte, error) {
 	if int64(len(gzipSrc)) > limit {
-		return nil, types.ErrLimit.Wrapf("max %d bytes", limit)
+		return nil, fmt.Errorf("%s: max %d bytes", errLimit, limit)
```
alpe, line 15: Here an others: wrapping the error via stdlib is fine. Please use `%w` for this. 
```suggestion
		return nil, fmt.Errorf("%w: max %d bytes", errLimit, limit)
```
Alternatively you can also use:
```suggestion
		return nil, errorsmod.Wrapf(errLimit, "max %d bytes", errLimit, limit)
```

x/wasm/ioutils/ioutil.go
```diff
@@ -22,14 +21,14 @@ func Uncompress(gzipSrc []byte, limit int64) ([]byte, error) {
 	zr.Multistream(false)
 	defer zr.Close()
 	bz, err := io.ReadAll(LimitReader(zr, limit))
-	if types.ErrLimit.Is(err) {
-		return nil, errorsmod.Wrapf(err, "max %d bytes", limit)
+	if err == errLimit {
```
alpe, line 24: Nit: my IDE complained and suggested this to cover wrapped errors, too
```suggestion
	if errors.Is(err, errLimit) {
```

app/ante.go
```diff
@@ -46,6 +47,7 @@ func NewAnteHandler(options HandlerOptions) (sdk.AnteHandler, error) {
 		ante.NewSetUpContextDecorator(), // outermost AnteDecorator. SetUpContext must be called first
 		wasmkeeper.NewLimitSimulationGasDecorator(options.WasmConfig.SimulationGasLimit), // after setup context to enforce limits early
 		wasmkeeper.NewCountTXDecorator(options.TXCounterStoreKey),
+		wasmkeeper.NewGasRegisterDecorator(options.WasmKeeper.GetGasRegister()),
```
alpe, line 50: 👍 this give a reference to the correct gas register

x/wasm/keeper/ante.go
```diff
@@ -95,3 +95,21 @@ func (d LimitSimulationGasDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simu
...
+}
+
+// NewGasRegisterDecorator constructor.
+func NewGasRegisterDecorator(gr GasRegister) *GasRegisterDecorator {
+	return &GasRegisterDecorator{gasRegister: gr}
+}
+
+// AnteHandle adds the gas register to the context.
+func (g GasRegisterDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (sdk.Context, error) {
+	if simulate {
```
alpe, line 111: In simulation calls messages are executed. Therefore the gas register must be set or we run into an error

x/wasm/keeper/keeper.go
```diff
@@ -145,6 +145,11 @@ func (k Keeper) GetAuthority() string {
 	return k.authority
 }
 
+// GetGasRegister returns the x/wasm module's gas register.
+func (k Keeper) GetGasRegister() GasRegister {
```
alpe, line 149: 👍 makes sense

x/wasm/types/authz.go
```diff
@@ -11,17 +12,127 @@ import (
...
+	storeMsg, ok := msg.(*MsgStoreCode)
+	if !ok {
+		return authztypes.AcceptResponse{}, sdkerrors.ErrInvalidRequest.Wrap("unknown msg type")
+	}
+
+	code := storeMsg.WASMByteCode
+	permission := *storeMsg.InstantiatePermission
+
+	if ioutils.IsGzip(code) {
+		gasRegister, ok := GasRegister(ctx)
```
alpe, line 56: 👍

x/wasm/types/context.go
```diff
@@ -14,8 +14,15 @@ const (
 	contextKeyQueryStackSize contextKey = iota
 	// authorization policy for sub-messages
 	contextKeySubMsgAuthzPolicy = iota
+	// gas register
+	contextKeyGasRegister = iota
 )
 
+type gasRegister interface {
```
alpe, line 21: There is an interface `GasRegister` in the keeper package. Let's move this to `types` package and use it here. Otherwise it can get confusing to have both.

alpe (commented):
Very good progress. 🏄

proto/cosmwasm/wasm/v1/authz.proto
```diff
@@ -4,12 +4,24 @@ package cosmwasm.wasm.v1;
...
 import "cosmos_proto/cosmos.proto";
 import "cosmos/base/v1beta1/coin.proto";
+import "cosmwasm/wasm/v1/types.proto";
 import "google/protobuf/any.proto";
 import "amino/amino.proto";
 
 option go_package = "github.com/CosmWasm/wasmd/x/wasm/types";
 option (gogoproto.goproto_getters_all) = false;
 
+// StoreCodeAuthorization defines authorization for wasm code upload.
```
alpe, line 14: // Since: wasmd 0.42

alpe (approved):
Great job! 🥇

proto/cosmwasm/wasm/v1/authz.proto
```diff
@@ -34,6 +47,18 @@ message ContractMigrationAuthorization {
       [ (gogoproto.nullable) = false, (amino.dont_omitempty) = true ];
 }
 
+// CodeGrant a granted permission for a single code
+message CodeGrant {
+  // CodeHash is the unique identifier created by wasmvm
+  // Wildcard "*" is used to specify any kind of grant.
+  bytes code_hash = 1;
```
webmaster128, line 54: Is there a reason why this is not a string? All checksums are lower case hex string in the CosmWasm stack.

What kind of bytes are expected here? 32 raw bytes hash output or the hex string?

pinosu: yes, it was kept as bytes to keep consistency with other code_hash/checksum attributes in the proto files such as:
``` types.proto
message CodeInfo {
  // CodeHash is the unique identifier created by wasmvm
  bytes code_hash = 1;
  ```
  and
  ``` tx.proto
  message MsgStoreCodeResponse {
  // CodeID is the reference to the stored WASM code
  uint64 code_id = 1 [ (gogoproto.customname) = "CodeID" ];
  // Checksum is the sha256 hash of the stored code
  bytes checksum = 2;
}
```

pinosu: Thanks for pointing this out. I will be more careful in the future.
Maybe in future releases we can fix all these fields and have migrations for them. I don't think it will be too difficult.

PR #1599, opened 2023-09-04
Remove Wasmer references

### LINKED ISSUE #1596: Remove all references to "Wasmer"
In the codebase there are a lot of references to Wasmer. I assume for historic reasons. Wasmer is the engine used by cosmwasm-vm/wasmvm to execute WebAssembly. However, this is an implementation detail and nothing wasmd should worry about. A different engine could be used any time or even multiple could be used in differen circumstances.

To avoid confusion and improve overall code quality, pretty much all references to Wasmer should be removed from here. In a quick search I found those, but the list is probably not complete:
- `WasmerEngine interface`
- `MockWasmer`
- `NewIBCContractMockWasmer`
- "wasmer gas" should be ["CosmWasm gas"](https://github.com/CosmWasm/cosmwasm/blob/main/docs/GAS.md)
- "Wasmer function execution"

A full text search in the codebase should find them all.

pinosu:
Resolves #1596

### DIFF
--- a/x/wasm/ibc_integration_test.go
+++ b/x/wasm/ibc_integration_test.go
@@ -48,7 +48,7 @@ func TestOnChanOpenInitVersion(t *testing.T) {
 			var (
 				chainAOpts = []wasmkeeper.Option{
 					wasmkeeper.WithWasmEngine(
-						wasmtesting.NewIBCContractMockWasmer(myContract)),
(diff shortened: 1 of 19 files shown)


alpe (approved):
👍 very nice. Good balance to keep references to wasmer where it still makes sense

Issue #1559, opened 2023-08-17
Question about TestQueryContractListByCodeOrdering


170210:
I am currently reading the tests for the msg server, and I have a question about the [TestQueryContractListByCodeOrdering](https://github.com/CosmWasm/wasmd/blob/main/x/wasm/keeper/querier_test.go#L268). This test doesn't seem to be verifying the ordering of contracts.

Additionally, I attempted to comment out the [block setting](https://github.com/170210-private/wasmd/blob/issue/order/x/wasm/keeper/querier_test.go#L294-L296), and it seems to return contracts in reverse order. 

We can get below address when manage some realistic block settings

>	cosmos14hj2tavq8fpesdwxxcu44rty3hh90vhujrvcmstl4zr3txmfvw9s4hmalr
	cosmos1suhgf5svhu4usrurvxzlgn54ksxmn8gljarjtxqnapv8kjnp4nrs2zhgh2
	cosmos1yyca08xqdgvjz0psg56z67ejh9xms6l436u8y58m82npdqqhmmtq8xrd4s

But after I comment out the setting, the result become reverse order

>	cosmos1yyca08xqdgvjz0psg56z67ejh9xms6l436u8y58m82npdqqhmmtq8xrd4s
	cosmos1suhgf5svhu4usrurvxzlgn54ksxmn8gljarjtxqnapv8kjnp4nrs2zhgh2
	cosmos14hj2tavq8fpesdwxxcu44rty3hh90vhujrvcmstl4zr3txmfvw9s4hmalr

Why does the block setting affect the order of contracts?
Could anyone please tell me the reason?

alpe:
Good questions!
> This test doesn't seem to be verifying the ordering of contracts.

This observation is correct. We need to add the proper assertion.

> Why does the block setting affect the order of contracts?

To support pagination of the result set, we store an artificial "timestamp" of height/gas_used in the secondary index on [instantiation](https://github.com/CosmWasm/wasmd/blob/v0.41.0/x/wasm/keeper/keeper.go#L342).
When you use the same block settings for all example entries, then the order is not by creation date anymore but by address. New elements would not be added at the end but at more or less random positions which breaks pagination for the clients. Please note that we have a second address generation with `instantiate2`

170210:
Thanks for your quick reply. I got it.

> We need to add the proper assertion for full coverage.

Is it necessary to make a pull request about this issue. 

[...]

PR #1607, opened 2023-09-07
test: add test cases in ContractsByCode

170210:
Close: #1559 
This PR intends to add test cases in ContractsByCode to cover the result set.


alpe (approved):
👍 very nice! thank you for taking care! 💐

PR #1612, opened 2023-09-07
chore: update proto builder and protos

faddat:
This PR updates the proto builder image and rebuilds the .proto files


alpe:
Cherry picked

PR #1613, opened 2023-09-07
chore: update ibc-go

faddat:
This PR updates ibc-go module paths to v8


alpe:
Thanks for the help but please bare with me for some days. We are getting the v0.42 release out of the door and then the sdk 50 work branch into main. There is a lot of rebasing and `push -f` currently that makes it hard to contribute. I also have not the capacity to go into details or do reviews on big PRs.
As with other major sdk version upgrades, this is some work that we can not split or outsource easily. We need to understand all the changes and the implications that may affect the wasmvm to ensure quality.
There is usually a lot of internal communication with the sdk or ibc-go team in the background. 
In order to help most at this stage, please do reviews. If I would know that people are looking at a branch, I would not rebase it.

alpe:
I have cherry-picked the commit. 🍒 Thank you!

PR #1618, opened 2023-09-11
Allow empty attribute values

### LINKED ISSUE #1617: Why can't the value in add_attribute be “”?
I think the need for event's value to be "" should still be quite high.
https://github.com/CosmWasm/wasmd/blob/main/x/wasm/keeper/events.go#L57-L61

https://github.com/CosmWasm/cosmwasm/issues/1869

https://github.com/CosmWasm/wasmd/pull/1618

I don't know why it can't be `""`, I think it's still necessary to remove that part of the code.

If it's a judgment about add_attribute not being able to be "" it shouldn't be in `wasmd`, you should put it in `cosmwasm-std`.
https://github.com/CosmWasm/cosmwasm/blob/7b2e753e4752eeb2466a7411ec8e177147581c40/packages/std/src/results/events.rs

99Kies:
Resolves #1617


alpe (approved):
Thank you for the contribution!  💐 
Nice work. The tests are also very much appreciated!

PR #1620, opened 2023-09-12
Start rework channel query

### LINKED ISSUE #1597: Bug in IbcQuery::ListChannels implementation when port is unset
According to the API specification, an unset port ID falls back to the contract's port ID:

> If `port_id` is omitted, this list all channels bound to the contract's port.

(https://github.com/CosmWasm/cosmwasm/blob/v1.3.3/packages/std/src/query/ibc.rs#L16-L20), and

> If `PortID` is unset, this list all channels bound to the contract's port.

(https://github.com/CosmWasm/wasmvm/blob/v1.3.0/types/queries.go#L165-L172). This specification never changed since the API's existence.

However, in https://github.com/CosmWasm/wasmd/pull/439/files#diff-47293308b181e6cdf9a982870e2de9941dfb405553ed239f7a192ec46963ec58R180 the implemetation changed to list channels for all ports in that case, instead of the contract's port. This is a bug in wasmd.

The motivation for changing the implementation instead of documenting the current behaviour is the following:
- The behaviour of falling back to the contract's port is consistent with `IbcQuery::Channel`
- Listing all channels of the chain is hardly something that makes sense to do in a smart contract
- It seems like the query is hardly used. After asking around [only one user was found: astroport](https://github.com/CosmWasm/cosmwasm/discussions/1849#discussioncomment-6879716). That user can probably better use a different query.

alpe:
Resolves #1597

This may change the result set for `list-channel` queries when no port is passed. 
Before we returned all open channels but now we fallback to the contracts ibc port or return an empty list for non ibc-contracts.


x/wasm/keeper/query_plugin_integration_test.go
```diff
@@ -642,6 +643,194 @@ func TestDistributionQuery(t *testing.T) {
...
+			},
+		},
+		"non ibc contract - with query portID empty": {
+			contract: nonIbcExample.Contract,
+			setup:    withChannelsStored(myIBCPortID, myExampleChannels...),
+			query:    &wasmvmtypes.IBCQuery{ListChannels: &wasmvmtypes.ListChannelsQuery{}},
+			assert: func(t *testing.T, d []byte) {
+				rsp := unmarshalReflect[wasmvmtypes.ListChannelsResponse](t, d)
+				assert.Nil(t, rsp.Channels)
+			},
```
pinosu, line 797: really good tests!

x/wasm/keeper/query_plugins.go
```diff
@@ -249,11 +249,18 @@ func IBCQuerier(wasm contractMetaDataSource, channelKeeper types.ChannelKeeper)
 		}
 		if request.ListChannels != nil {
 			portID := request.ListChannels.PortID
-			channels := make(wasmvmtypes.IBCChannels, 0)
-			channelKeeper.IterateChannels(ctx, func(ch channeltypes.IdentifiedChannel) bool {
-				// it must match the port and be in open state
-				if (portID == "" || portID == ch.PortId) && ch.State == channeltypes.OPEN {
-					newChan := wasmvmtypes.IBCChannel{
+			if portID == "" { // then fallback to contract port address
+				portID = wasm.GetContractInfo(ctx, caller).IBCPortID
```
pinosu, line 253: 👍

pinosu (approved):
LGTM 💯

PR #1630, opened 2023-09-18
Fix genesis import with predictable addresses

### LINKED ISSUE #1629: Ensure genesis import works with both address generators
There is currently an issue with the sanity check on sequence values when predicatble addresses are used. 
See https://github.com/osmosis-labs/wasmd/pull/19 for more details on the problem

alpe:
Resolves #1629

### DIFF
--- a/x/wasm/keeper/genesis.go
+++ b/x/wasm/keeper/genesis.go
@@ -41,7 +41,6 @@ func InitGenesis(ctx sdk.Context, keeper *Keeper, data types.GenesisState) ([]ab
 		}
 	}
 
-	var maxContractID int
 	for i, contract := range data.Contracts {
 		contractAddr, err := sdk.AccAddressFromBech32(contract.ContractAddress)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


pinosu (approved):
LGTM 👍

czarcas7ic:
This fix doesn't confirm that contracts with predictable addresses were properly exported though right?

alpe:
> This fix doesn't confirm that contracts with predictable addresses were properly exported though right?

All contracts are stored in the same model and exported. This fix ensures that predictable addresses do not cause issues on genesis import. The contract id sequence had a constraint for sanity that ensured `contractID_Seq > count(contract)`. With any predictable address though the `count(contract)` is always `> contractID_Seq`. Therefore the fix is either 
a) bump contractID_Seq for every contract instantiation
b) export `count(contract) +1` as `contractID_Seq`
c) replace the import constraint with something better

My motivation for c)
a) comes with some extra storage IO and gas costs
b) creates a different state on export/import in the store and will cause simulations and other tests to fail
c) the constraint this is only a safety net to fail fast on inconsistent manual editions to the genesis. We achieve the same by checking that the next contract address is available and not used

PR #1640, opened 2023-09-27
Add msg update contract label

### LINKED ISSUE #1601: Make contract lable updatable
Currently, the admin of a contract is able to migrate the contract (i.e. change the code), and change or remove the admin but isn't able to change the contract's label. 

From a theoretical point of view, it just looks a bit strange (for me) that the only things that can't be changed by the admin are the address (which is obvious) and the label of the contract. While an address is essentially the thing that defines the contract instance the label is just metadata that is carved in stone while even the code/admin can be changed (thus the purpose of the contract can be changed which might make the initial label misleading).

From the practical point of view, Neutron organized DAO contracts with the hierarchical naming convention that helps with e.g. querying the contracts from the chain. Since labels aren't editable any mistake made is going to be preserved and there's no way to update the convention (even while DAO is the admin of those contracts).

I propose to consider making contract labels updatable by the contract's admin.

pinosu:
Resolves #1601


x/wasm/keeper/keeper.go
```diff
@@ -653,6 +653,26 @@ func (k Keeper) setContractAdmin(ctx context.Context, contractAddress, caller, n
...
+	if !authZ.CanModifyContract(contractInfo.AdminAddr(), caller) {
+		return errorsmod.Wrap(sdkerrors.ErrUnauthorized, "can not modify contract")
+	}
+	contractInfo.Label = newLabel
+	k.mustStoreContractInfo(sdkCtx, contractAddress, contractInfo)
+	sdkCtx.EventManager().EmitEvent(sdk.NewEvent(
+		types.EventTypeUpdateContractLabel,
+		sdk.NewAttribute(types.AttributeKeyContractAddr, contractAddress.String()),
+		sdk.NewAttribute(types.AttributeKeyNewLabel, newLabel),
+	))
```
alpe, line 671: nice!

x/wasm/keeper/keeper_test.go
```diff
@@ -2489,6 +2489,58 @@ func TestGasConsumed(t *testing.T) {
...
+		"update label - unauthorized": {
+			newLabel: "new label",
+			caller:   RandomAccountAddress(t),
+			policy:   DefaultAuthorizationPolicy{},
+			expErr:   true,
+		},
+		"update label - gov policy": {
+			newLabel: "new label",
+			policy:   GovAuthorizationPolicy{},
+			caller:   example.CreatorAddr,
```
alpe, line 2517: When you use a random caller address, then it would ensure the different behaviour of the gov policy

x/wasm/keeper/keeper_test.go
```diff
@@ -2489,6 +2489,58 @@ func TestGasConsumed(t *testing.T) {
...
+			require.Len(t, em.Events(), 1)
+			assert.Equal(t, "update_contract_label", em.Events()[0].Type)
+			exp := map[string]string{
+				"_contract_address": example.Contract.String(),
+				"new_label":         spec.newLabel,
+			}
+			assert.Equal(t, exp, attrsToStringMap(em.Events()[0].Attributes))
+		})
+	}
+}
```
alpe, line 2542: Good tests! For bonus points, there is also the unknown contract address scenario that must not panic.

x/wasm/keeper/msg_server.go
```diff
@@ -462,3 +462,26 @@ func (m msgServer) StoreAndMigrateContract(goCtx context.Context, req *types.Msg
...
+		return nil, errorsmod.Wrap(err, "contract")
+	}
+
+	policy := m.selectAuthorizationPolicy(ctx, msg.Sender)
+
+	if err := m.keeper.setContractLabel(ctx, contractAddr, senderAddr, msg.NewLabel, policy); err != nil {
+		return nil, err
+	}
+
+	return &types.MsgUpdateContractLabelResponse{}, nil
```
alpe, line 486: straight forward 👍

x/wasm/keeper/msg_server_integration_test.go
```diff
@@ -1144,3 +1144,73 @@ func TestStoreAndMigrateContract(t *testing.T) {
...
+	}{
+		"authority can update contract label": {
+			addr:   authority,
+			expErr: false,
+		},
+		"admin can update contract label": {
+			addr:   myAddress.String(),
+			expErr: false,
+		},
+		"other address cannot update contract label": {
```
alpe, line 1170: 🤔 good to have the scenarios covered but there is some overlap with the keeper tests. 
The integration tests give more confidence on the whole process. If you see value in both, then we can keep them.

x/wasm/keeper/msg_server_integration_test.go
```diff
@@ -1144,3 +1144,73 @@ func TestStoreAndMigrateContract(t *testing.T) {
...
+			addr:   otherAddr.String(),
+			expErr: true,
+		},
+	}
```
alpe, line 1174: Please check for invalid data

x/wasm/keeper/msg_server_integration_test.go
```diff
@@ -1144,3 +1144,73 @@ func TestStoreAndMigrateContract(t *testing.T) {
...
+			if spec.expErr {
+				require.Error(t, err)
+				require.Equal(t, "old label", wasmApp.WasmKeeper.GetContractInfo(ctx, contractAddr).Label)
+			} else {
+				require.NoError(t, err)
+				require.Equal(t, "new label", wasmApp.WasmKeeper.GetContractInfo(ctx, contractAddr).Label)
+			}
+		})
+	}
+}
```
alpe, line 1216: Nice test!

x/wasm/types/codec.go
```diff
@@ -79,6 +80,7 @@ func RegisterInterfaces(registry types.InterfaceRegistry) {
 		&MsgAddCodeUploadParamsAddresses{},
 		&MsgRemoveCodeUploadParamsAddresses{},
 		&MsgStoreAndMigrateContract{},
+		&MsgUpdateContractLabel{},
```
alpe, line 83: Important!

x/wasm/types/tx_test.go
```diff
@@ -1326,3 +1326,71 @@ func TestMsgStoreAndMigrateContractValidation(t *testing.T) {
...
+		"bad contract addr": {
+			src: MsgUpdateContractLabel{
+				Sender:   goodAddress,
+				NewLabel: "new label",
+				Contract: badAddress,
+			},
+			expErr: true,
+		},
+	}
+	for msg, spec := range specs {
```
alpe, line 1386: Excellet!

alpe (approved):
Great work. 🎸 
Only minor nits. Feel free to merge when you are happy with the code.

I would add the backport label to the PR but the backport will need some work from you, too

PR #1649, opened 2023-10-03
fix: propagate funds validation errors

aelesbao:
The validation functions on the `tx` type masquerade the root error message for the `Funds` validation. Having the original error helps to save time when debugging the cause for a failed tx.

One example is if someone sends multiple funds to a contract execution without sorting the denoms, which is one of the validations in the `Coins.Validate` method. With the error propagation, the developer can quickly determine why the tx failed.

### DIFF
--- a/x/wasm/types/tx.go
+++ b/x/wasm/types/tx.go
@@ -103,8 +103,8 @@ func (msg MsgInstantiateContract) ValidateBasic() error {
 		return errorsmod.Wrap(err, "label")
 	}
 
-	if !msg.Funds.IsValid() {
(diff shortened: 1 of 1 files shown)


alpe (commented):
Thanks for the PR!
Better error messages make sense. 👍 
The deprecated gov beta1 types were removed this morning. Can you rebase your PR please?

aelesbao:
Thanks for the reply, @alpe. I've rebased the branch and updated the pull request description.

alpe (approved):
LGTM! Thanks for doing this 💐

PR #1650, opened 2023-10-05
Add label validation for non printable chars

### LINKED ISSUE #1623: Prevent non printable characters in label
The label was designed for humans. Non printable characters should be rejected therefore

pinosu:
Resolves #1623

### DIFF
--- a/x/wasm/types/tx_test.go
+++ b/x/wasm/types/tx_test.go
@@ -116,6 +116,24 @@ func TestInstantiateContractValidation(t *testing.T) {
 			},
 			valid: false,
 		},
+		"white space ending label": {
+			msg: MsgInstantiateContract{
+				Sender: goodAddress,
+				CodeID: firstCodeID,
+				Label:  "foo ",
+				Msg:    []byte("{}"),
+			},
+			valid: false,
+		},
+		"non printable chars in label": {
+			msg: MsgInstantiateContract{
+				Sender: goodAddress,
+				CodeID: firstCodeID,
+				Label:  "foo\v",
+				Msg:    []byte("{}"),
+			},
+			valid: false,
+		},
 		"label too long": {
 			msg: MsgInstantiateContract{
 				Sender: goodAddress,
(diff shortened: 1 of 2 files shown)


x/wasm/types/validation.go
```diff
@@ -45,6 +46,12 @@ func ValidateLabel(label string) error {
 	if label != strings.TrimSpace(label) {
 		return ErrInvalid.Wrap("label must not start/end with whitespaces")
 	}
+	labelWithoutNonPrintableChars := strings.TrimFunc(label, func(r rune) bool {
+		return !unicode.IsGraphic(r)
+	})
+	if label != labelWithoutNonPrintableChars {
+		return ErrInvalid.Wrap("label must not have non printable characters")
```
alpe, line 53: nit: let's avoid double negative ;-)
```suggestion
		return ErrInvalid.Wrap("label must have printable characters only")
```

x/wasm/types/tx_test.go
```diff
@@ -116,6 +116,24 @@ func TestInstantiateContractValidation(t *testing.T) {
...
+		},
+		"non printable chars in label": {
+			msg: MsgInstantiateContract{
+				Sender: goodAddress,
+				CodeID: firstCodeID,
+				Label:  "foo\v",
+				Msg:    []byte("{}"),
+			},
+			valid: false,
+		},
```
alpe, line 136: We need to ensure that there are only printable characters in the whole string. So `f\voo` should be rejected as well

alpe (commented):
Nice start. Please see my comment on the test

x/wasm/types/validation.go
```diff
@@ -45,6 +46,15 @@ func ValidateLabel(label string) error {
 	if label != strings.TrimSpace(label) {
 		return ErrInvalid.Wrap("label must not start/end with whitespaces")
 	}
+	labelWithPrintableCharsOnly := strings.Map(func(r rune) rune {
```
alpe, line 49: very nice use of strings.Map 👍

alpe (approved):
Good work! 🏌️

PR #1661, opened 2023-10-09
Genesis migration system tests

alpe:
🚧 WIP: early version


alpe:
Closing this: genesis export/import is not a recommend way for a chain upgrade

Related: https://github.com/cosmos/cosmos-sdk/issues/18021#issuecomment-1754706981

PR #1666, opened 2023-10-12
Classic address tests

DariuszDepta:
Added tests for `BuildContractAddressClassic` function.
And small spelling ;-)

### DIFF
--- a/x/wasm/keeper/addresses_test.go
+++ b/x/wasm/keeper/addresses_test.go
@@ -17,8 +17,8 @@ func TestBuildContractAddressClassic(t *testing.T) {
 	t.Cleanup(func() {
 		sdk.GetConfig().SetBech32PrefixForAccount(x, y)
 	})
-	// set custom Bech32 settings
 	sdk.GetConfig().SetBech32PrefixForAccount("purple", "purple")
+
 	// prepare test data
 	type Spec struct {
 		In struct {
@@ -39,7 +39,7 @@ func TestBuildContractAddressClassic(t *testing.T) {
 			gotAddr := BuildContractAddressClassic(spec.In.CodeId, spec.In.InstanceId)
 			// then
 			require.Equal(t, spec.Out.Address.String(), gotAddr.String())
-			//require.NoError(t, sdk.VerifyAddressFormat(gotAddr))
+			require.NoError(t, sdk.VerifyAddressFormat(gotAddr))
 		})
 	}
 }
--- a/x/wasm/keeper/addresses_test.go
+++ b/x/wasm/keeper/addresses_test.go
@@ -3,9 +3,9 @@ package keeper
 import (
 	"encoding/json"
 	"fmt"
+	tmbytes "github.com/cometbft/cometbft/libs/bytes"
 	"testing"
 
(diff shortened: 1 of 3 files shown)


x/wasm/keeper/addresses_test.go
```diff
@@ -3,20 +3,82 @@ package keeper
...
 		sdk.GetConfig().SetBech32PrefixForAccount(x, y)
+		sdk.SetAddrCacheEnabled(c)
 	})
+
+	// set custom Bech32 settings
 	sdk.GetConfig().SetBech32PrefixForAccount("purple", "purple")
+	// disable address cache
+	// AccAddress -> String conversion is then slower, but does not lead to errors like this:
+	//   runtime error: invalid memory address or nil pointer dereference
+	sdk.SetAddrCacheEnabled(false)
```
pinosu, line 28: you could extract this piece of code and have it into a separate private function e.g. `setup(t *testing.T)` since it is same for both tests.

x/wasm/keeper/addresses_test.go
```diff
@@ -3,20 +3,82 @@ package keeper
...
+		codeId:     1,
+		instanceId: 0,
+		address:    "purple1mzdhwvvh22wrt07w59wxyd58822qavwkx5lcej7aqfkpqqlhaqfs5efvjk",
+	},
+	{
+		codeId:     1,
+		instanceId: 1,
+		address:    "purple14hj2tavq8fpesdwxxcu44rty3hh90vhujrvcmstl4zr3txmfvw9smc2vxm",
+	},
+}
```
pinosu, line 67: you could declare and initialize this directly inside the test function because it is closely related to it and definitely we are not using somewhere else.
Moreover, as personal preference I would call `expAddress` instead of `address`, just to make it clear what are the inputs and what are the expected values.

pinosu (changes requested):
Very good! 💯 
There are some small changes that would improve a bit the code, but very nice to have also some tests for the address generation.

x/wasm/keeper/addresses_test.go
DariuszDepta: Done.

x/wasm/keeper/addresses_test.go
DariuszDepta: Done.

x/wasm/keeper/addresses_test.go
```diff
@@ -3,21 +3,75 @@ package keeper
...
 	"testing"
 
-	tmbytes "github.com/cometbft/cometbft/libs/bytes"
 	"github.com/stretchr/testify/require"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
-func TestBuildContractAddress(t *testing.T) {
+func PrepareCleanup(t *testing.T) {
```
pinosu, line 14: this function shuld be private
```suggestion
func prepareCleanup(t *testing.T) {
```

pinosu (approved):
LGTM 🥇 
Just make the function private and then feel free to merge it!

x/wasm/keeper/addresses_test.go
DariuszDepta: Done.

PR #1679, opened 2023-10-18
Rename some more "features" to "capabilities"

webmaster128:


x/wasm/keeper/addresses_test.go
```diff
@@ -3,9 +3,9 @@ package keeper
 import (
 	"encoding/json"
 	"fmt"
-	tmbytes "github.com/cometbft/cometbft/libs/bytes"
 	"testing"
 
+	tmbytes "github.com/cometbft/cometbft/libs/bytes"
```
alpe, line 8: This change should not be in here. Please run `make format`

x/wasm/keeper/reflect_test.go
```diff
@@ -24,8 +24,8 @@ import (
 )
 
 const (
-	CyberpunkFeatures = "staking,mask,stargate,cosmwasm_1_1,cosmwasm_1_2,cosmwasm_1_3,cosmwasm_1_4"
-	ReflectFeatures   = CyberpunkFeatures
+	CyberpunkCapabilities = "staking,mask,stargate,cosmwasm_1_1,cosmwasm_1_2,cosmwasm_1_3,cosmwasm_1_4"
+	ReflectCapabilities   = CyberpunkCapabilities
```
alpe, line 28: 👍

alpe (approved):
Thanks for the housekeeping. Please run the formatter.

x/wasm/keeper/addresses_test.go
webmaster128: This is the result of `make format`. It's in a separate commit. Feel free to kick it out if needed

PR #1697, opened 2023-11-08
Prevent empty channel version

### LINKED ISSUE #1693: OnChanOpenInit Implementation may not follow IBCModule docs
Maybe this is not relevant in practice, but it seems to me that [our implementation of OnChanOpenInit](https://github.com/CosmWasm/wasmd/blob/606289ab56dbe803bc28941a4b4c361b26435d37/x/wasm/ibc.go#L40) doesn't uphold the contract specified in the [doc comment of the IBCModule
](https://github.com/cosmos/ibc-go/blob/9c7212198d0ef82b8219ea66cee9c96b40e7981d/modules/core/05-port/types/module.go#L23-L24):
> If the version string is empty, OnChanOpenInit is expected to
return a default version string representing the version(s) it supports.
If there is no default version string for the application,
it should return an error if provided version is empty string.

In our implementation, we let the [contract return a version](https://github.com/CosmWasm/wasmd/blob/606289ab56dbe803bc28941a4b4c361b26435d37/x/wasm/ibc.go#L73) or [default to the provided one](https://github.com/CosmWasm/wasmd/blob/606289ab56dbe803bc28941a4b4c361b26435d37/x/wasm/ibc.go#L77-L79) if the contract does not return a version, but we do not handle the case mentioned above, where the contract does not return a version and the provided version is empty.

alpe:
Resolves #1693

### DIFF
--- a/x/wasm/ibc.go
+++ b/x/wasm/ibc.go
@@ -75,6 +75,9 @@ func (i IBCHandler) OnChanOpenInit(
 		return "", err
 	}
 	if acceptedVersion == "" { // accept incoming version when nothing returned by contract
+		if version == "" {
+			return "", types.ErrEmpty.Wrap("version")
+		}
 		acceptedVersion = version
 	}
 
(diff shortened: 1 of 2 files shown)


pinosu (approved):
LGTM 👍

PR #1714, opened 2023-11-13
Fix local test scripts

alpe:
Adjust for sdk 50


pinosu (approved):
tested locally and works well! Nice work 💯

PR #1719, opened 2023-11-15
Pass empty attribute value through stack

alpe:
Abusing ping-pong test as hack to fail fast when Event attributes are handled differently by the IBC + SDK stack.
Relates to #1617


pinosu (approved):
LGTM 👍

PR #1732, opened 2023-11-17
Handle legacy OnlyAddress parameter

alpe:
See # 1725

I have manually tested this with a v033 chain upgrade to v04x on my box


pinosu (approved):
LGTM 👍

PR #1748, opened 2023-11-30
Remove codeowners

alpe:
Today is my last day at Confio.
I am very thankful for the opportunity to work on this project in the past and the great support that I got from the community!  💪 🤗

I am deleting  the codeowners file as suggested by @webmaster128


webmaster128 (approved):
Thank you Alex for all your efforts. You'll be missed

PR #1753, opened 2023-12-10
Add Build Address gRPC Query

NotJeremyLiu:
Closes #1705

PR #1765, opened 2023-12-19
Add RPC bindings to build-address

codehans:
This PR adds basic RPC bindings to allow user-interfaces to query addresses with Instantiate2. Our use-case specifically is to support creation of a UI to support instantiation of multiple dependent contracts in a single governance proposal.


pinosu:
Thanks for your PR!
There is already a PR ongoing which adds the same feature. Please take a look at #1753 and feel free to review it/add comments.
If youagree, I will close this pr in favor of #1753.

codehans:
Ah yeah perfect @pinosu , not sure how I missed that, thanks!

PR #1776, opened 2024-01-14
fix: use protos from sdk v0.50.x

faddat:
This PR ensures that we are using the protos from the SDK v0.50.x series


pinosu:
Hey @faddat, thanks for your contribution! 💐 
Could you please run the command `make proto-all` , which runs also the formatter to keep code more readable?

pinosu (changes requested):
See comment above

pinosu:
Closing in favor of #1795

PR #1797, opened 2024-01-25
Rename InstantiateContractCosts -> SetupContractCost; Remove NewContractInstanceCosts

webmaster128:
This PR makes to changes to improve clarity on how starting a contract execution is charged. They are independent an in one commit each:

1. **Rename InstantiateContractCosts -> SetupContractCost.** The term "instantiate" has two different meanings in our stack: (a) the contract instance identified by a blockchain address (b) the Wasmer instance created before calling an entry point. Here "InstantiateContract" was referring to the later, which is more an implementation detail. To avoid confusion we can use a different term that was in the codebase already: "setup".
2. **Remove NewContractInstanceCosts.** This was just an alias that was used for one entry point (MsgInstantiateContract/MsgInstantiateContract2). However, we do not maintain entry-point specific gas values for all other cases. So having this extra is confusing. If we want entrypoint specific functions here, we should create them for all entry points.

Also in general we should migrate "costs" -> "cost" when we get the chance since this the the price for a single operation.

### DIFF
--- a/x/wasm/keeper/keeper.go
+++ b/x/wasm/keeper/keeper.go
@@ -249,8 +249,8 @@ func (k Keeper) instantiate(
 		return nil, nil, types.ErrEmpty.Wrap("creator")
 	}
 	sdkCtx := sdk.UnwrapSDKContext(ctx)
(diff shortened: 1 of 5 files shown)


pinosu (approved):
Nice work! LGTM 👍

PR #1800, opened 2024-01-31
[idea, unfinished] Add flattened msgResponses to SubMsgResponse

webmaster128:
This is a proposal for how to fill the msg_responses vector in the SubMsgResponse for CosmWasm 2.0. It preserves the two dimensional loop for flexibility in wasmd (also debugging and stuff) and flattens the result for the contract since we only expect 1 element anyways in the vast majority of cases.


webmaster128:
Merged as part of https://github.com/CosmWasm/wasmd/pull/1804

PR #1804, opened 2024-02-06
Update to wasmvm 2.0.0-rc.1

chipshort:
Very WIP, main goal is to get it to compile and pass current tests

TODOs:
- [x] supported capabilities are now `[]string`
- [x] nested Result changes
- [x] renaming
- [x] StoreCode gas
- [x] ValidateAddress
- [x] `Array[C]` type replacements
- [x] Adjust gas multiplier
- [x] TransferMsg.Memo
- [x] QueryRequest.Grpc
- [x] MsgResponses (see #1800)
- [x] IBCReceiveResponse.Acknowledgement == nil
- [x] Payload
- [x] Make `redactError` call conditional in `func (d MessageDispatcher) DispatchSubmessages`


x/wasm/keeper/keeper.go
```diff
@@ -169,7 +169,9 @@ func (k Keeper) create(ctx context.Context, creator sdk.AccAddress, wasmCode []b
 	}
 
 	sdkCtx.GasMeter().ConsumeGas(k.gasRegister.CompileCosts(len(wasmCode)), "Compiling wasm bytecode")
-	checksum, err = k.wasmVM.StoreCode(wasmCode)
+	gas := k.runtimeGasForContract(sdkCtx)
```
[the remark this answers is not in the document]
chipshort: I think gasLeft explains the intent better here. I was just keeping the naming the same as in the other functions.

x/wasm/keeper/api.go
```diff
@@ -22,25 +24,32 @@ const (
...
 
-func canonicalAddress(human string) ([]byte, uint64, error) {
+func canonicalizeAddress(human string) ([]byte, uint64, error) {
 	bz, err := sdk.AccAddressFromBech32(human)
 	return bz, costCanonical, err
 }
 
+func validateAddress(human string) (uint64, error) {
+	_, err := sdk.AccAddressFromBech32(human)
+	return costValidate, err
```
webmaster128, line 48: This implementation does not yet check that the address is normalized since bech32 allows both uppercase and lower case inputs.

For a generic normalization check we need to convert it back and and check if we get the same result as the input.

chipshort:
@webmaster128 Any idea why the CI docker-image libwasmvm check is failing?
When I locally run `(cd cmd/wasmd && go run . query wasm libwasmvm-version)`, I get `2.0.0-rc.1`. Is the docker image an old version?

webmaster128:
> @webmaster128 Any idea why the CI docker-image libwasmvm check is failing?
> When I locally run `(cd cmd/wasmd && go run . query wasm libwasmvm-version)`, I get `2.0.0-rc.1`. Is the docker image an old version?

I did not check the error yet but you need to update those lines for sure: https://github.com/CosmWasm/wasmd/blob/main/Dockerfile#L18-L21

webmaster128:
I think it would be valuable to do as many of the _Probably in follow-up PRs:_ items in here as well. This PR then serves as a go-to-place for every other user of wasmvm (which right now is primarily wasmd forks and the Wasm Light Client).

pinosu:
Changes look good so far! To make the CI pass you need to update the[ Dockerfile](https://github.com/CosmWasm/wasmd/blob/main/Dockerfile#L18-L21) both version and checksum for libwasmvm.

chipshort:
@pinosu @webmaster128 Does one of you know the difference between the different `reflect.wasm` files and `reflect_1_1.wasm` used in tests? Because of the `Stargate` to `Any` rename, we either need to update them to the CW 2.0 version or we need to continue calling them with `stargate` json. Any opinions on that?

webmaster128:
I think it would be pretty cool to test both, the old (stargate) and the new (any) message type to ensure wasmd supports both. I don't know about the 1_1 one. it was added here: https://github.com/CosmWasm/wasmd/pull/1055/files

webmaster128:
Could you add the cosmwasm_2_0 capability to the README (on top of #1805)?

chipshort:
I rebased on main and added the capability to the readme.
I also fixed some more tests and added an updated reflect contract that is now used in most tests, as well as one test that uses both the old and new reflect contract for a `Stargate` / `Any` msg (`TestGroupWithContract` / `TestGroupWithNewReflectContract`). That should cover the backwards compatibility with the old `Stargate` variant.

To get the tests to pass, [this PR in wasmvm](https://github.com/CosmWasm/wasmvm/pull/519/files) is needed because the `Payload` is currently serialized incorrectly on the Go side. After that, the tests should pass and I can start working on the other new features.

chipshort:
Tests are passing 🎉

chipshort:
I added some code for the `redactError` change. Not sure if this is the best way to do it or if it would be better to have a completely separate error type for that to match against (like it is done for the `SystemError`)

app/app.go
```diff
@@ -646,7 +644,7 @@ func NewWasmApp(
 		app.GRPCQueryRouter(),
 		wasmDir,
 		wasmConfig,
-		availableCapabilities,
+		AllCapabilities(),
```
pinosu, line 647: 👍

app/sim_test.go
```diff
@@ -331,6 +332,8 @@ func TestAppStateDeterminism(t *testing.T) {
 				logger = log.NewNopLogger()
 			}
 
+			appOptions.SetDefault(flags.FlagHome, t.TempDir()) // ensure a unique folder per run
```
pinosu, line 335: nice! makes sense 👍

Dockerfile
```diff
@@ -15,13 +15,10 @@ RUN apk add git
 WORKDIR /code
 COPY . /code/
 # See https://github.com/CosmWasm/wasmvm/releases
-ADD https://github.com/CosmWasm/wasmvm/releases/download/v1.5.2/libwasmvm_muslc.aarch64.a /lib/libwasmvm_muslc.aarch64.a
-ADD https://github.com/CosmWasm/wasmvm/releases/download/v1.5.2/libwasmvm_muslc.x86_64.a /lib/libwasmvm_muslc.x86_64.a
-RUN sha256sum /lib/libwasmvm_muslc.aarch64.a | grep e78b224c15964817a3b75a40e59882b4d0e06fd055b39514d61646689cef8c6e
-RUN sha256sum /lib/libwasmvm_muslc.x86_64.a | grep e660a38efb2930b34ee6f6b0bb12730adccb040b6ab701b8f82f34453a426ae7
-
-# Copy the library you want to the final location that will be found by the linker flag `-lwasmvm_muslc`
-RUN cp /lib/libwasmvm_muslc.${arch}.a /lib/libwasmvm_muslc.a
```
pinosu, line 24: is this not needed anymore?

tests/e2e/group_test.go
```diff
@@ -74,3 +74,58 @@ func TestGroupWithContract(t *testing.T) {
...
+	propMsg, err := group.NewMsgSubmitProposal(policyAddr.String(), []string{contractAddr.String()}, payload, "my proposal", group.Exec_EXEC_TRY, "my title", "my description")
+	require.NoError(t, err)
+
+	rsp = e2e.MustExecViaAnyReflectContract(t, chain, contractAddr, propMsg)
+	var execRsp types.MsgExecuteContractResponse
+	chain.UnwrapExecTXResult(rsp, &execRsp)
+
+	var groupRsp group.MsgSubmitProposalResponse
+	require.NoError(t, chain.Codec.Unmarshal(execRsp.Data, &groupRsp))
+	// require.NotEmpty(t, groupRsp.ProposalId)
```
pinosu, line 125: why this is commented? please remove if not needed

tests/e2e/group_test.go
```diff
@@ -74,3 +74,58 @@ func TestGroupWithContract(t *testing.T) {
 	expBalanceAmount := sdk.NewCoin(sdk.DefaultBondDenom, sdkmath.OneInt())
 	assert.Equal(t, expBalanceAmount.String(), recipientBalance.String())
 }
+
+func TestGroupWithNewReflectContract(t *testing.T) {
```
pinosu, line 78: Nice test!

tests/e2e/reflect_helper.go
```diff
@@ -16,9 +18,18 @@ import (
 	"github.com/CosmWasm/wasmd/x/wasm/types"
 )
 
-// InstantiateReflectContract store and instantiate a reflect contract instance
+// InstantiateStargateReflectContract stores and instantiates a pre 2.0 reflect contract instance.
+// This instance still expects the old CosmosMsg.Stargate variant instead of the new CosmosMsg.Any.
+func InstantiateStargateReflectContract(t *testing.T, chain *ibctesting.TestChain) sdk.AccAddress {
+	codeID := chain.StoreCodeFile("../../x/wasm/keeper/testdata/reflect.wasm").CodeID
```
pinosu, line 24: Should we modify also the name of the file to give a meaningful name? Something like this
```suggestion
	codeID := chain.StoreCodeFile("../../x/wasm/keeper/testdata/stargate_reflect.wasm").CodeID
```

tests/e2e/reflect_helper.go
```diff
@@ -37,12 +48,38 @@ type sdkMessageType interface {
...
+	for i, m := range msgs {
+		bz, err := chain.Codec.Marshal(m)
+		require.NoError(t, err)
+		// json is built manually because the wasmvm CosmosMsg does not have the `Stargate` variant anymore
+		vmMsgs[i] = fmt.Sprintf("{\"stargate\":{\"type_url\":\"%s\",\"value\":\"%s\"}}", sdk.MsgTypeURL(m), base64.StdEncoding.EncodeToString(bz))
+	}
+	// build the complete reflect message
+	reflectSendBz := []byte(fmt.Sprintf("{\"reflect_msg\":{\"msgs\":%s}}", vmMsgs))
+
+	// {"reflect_msg":{"msgs":[{"stargate":{"type_url":"mytype","value":"EjQ="}}]}}
```
pinosu, line 63: remove

x/wasm/keeper/api.go
```diff
@@ -22,25 +26,39 @@ const (
...
 	costCanonical           = DefaultGasCostCanonicalAddress * types.DefaultGasMultiplier
+	costValidate            = DefaultGasCostCanonicalAddress * types.DefaultGasMultiplier
 	costJSONDeserialization = wasmvmtypes.UFraction{
 		Numerator:   DefaultDeserializationCostPerByte * types.DefaultGasMultiplier,
 		Denominator: 1,
 	}
 )
 
-func humanAddress(canon []byte) (string, uint64, error) {
+func humanizeAddress(canon []byte) (string, uint64, error) {
```
pinosu, line 36: 👍

x/wasm/keeper/handler_plugin.go
```diff
@@ -173,67 +175,69 @@ func NewIBCRawPacketHandler(ics4Wrapper types.ICS4Wrapper, channelKeeper types.C
...
 	moduleLogger(ctx).Debug("ibc packet set", "seq", seq)
 
+	var msgResponse [][]*codectypes.Any
 	resp := &types.MsgIBCSendResponse{Sequence: seq}
 	val, err := resp.Marshal()
 	if err != nil {
-		return nil, nil, errorsmod.Wrap(err, "failed to marshal IBC send response")
+		return nil, nil, nil, errorsmod.Wrap(err, "failed to marshal IBC send response")
 	}
+	// TODO: fill msgResponse
```
pinosu, line 206: is this pending?

x/wasm/keeper/handler_plugin_encoders_test.go
```diff
@@ -421,13 +419,50 @@ func TestEncoding(t *testing.T) {
...
 			},
 			expError: true,
 		},
+	}
+	encodingConfig := MakeEncodingConfig(t)
+	for name, tc := range cases {
+		t.Run(name, func(t *testing.T) {
+			var ctx sdk.Context
+			encoder := DefaultEncoders(encodingConfig.Codec, wasmtesting.MockIBCTransferKeeper{})
+			res, err := encoder.Encode(ctx, tc.sender, "", tc.srcMsg)
```
pinosu, line 435: probably the variable could be removed
```suggestion
			res, err := encoder.Encode(sdk.Context{}, tc.sender, "", tc.srcMsg)
```

x/wasm/keeper/handler_plugin_encoders_test.go
```diff
@@ -421,13 +419,50 @@ func TestEncoding(t *testing.T) {
...
+
+func TestEncodeIbcMsg(t *testing.T) {
+	var (
+		addr1 = RandomAccountAddress(t)
+		addr2 = RandomAccountAddress(t)
+	)
+	valAddr := make(sdk.ValAddress, types.SDKAddrLen)
+	valAddr[0] = 12
+	valAddr2 := make(sdk.ValAddress, types.SDKAddrLen)
+	valAddr2[1] = 123
```
pinosu, line 454: is this used anywhere?

pinosu (changes requested):
Great work! 💯 
I did a first round review. I added some minor comments.
I will review it again before approving and merging it.

Dockerfile
chipshort: The new version of wasmvm now links against `wasmvm_muslc.x86_64` / `wasmvm_muslc.aarch64` instead of `wasmvm_muslc`, so the renaming is not needed anymore.
It was changed in this PR: https://github.com/CosmWasm/wasmvm/pull/514/files

tests/e2e/reflect_helper.go
chipshort: Funny, I was thinking something similar while implementing this. Will do!

x/wasm/keeper/handler_plugin.go
chipshort: Nice catch! Completely overlooked that

x/wasm/keeper/handler_plugin_encoders_test.go
chipshort: No, leftover from copy-pasting. Removed

Dockerfile
webmaster128: Chris, could you add this change to https://github.com/CosmWasm/wasmvm/blob/main/docs/MIGRATING.md?

README.md
```diff
@@ -37,6 +37,7 @@ The following table shows which of the [latest capabilities](https://github.com/
 | cosmwasm_1_2 | x       | x       | x       |         |      |
 | cosmwasm_1_3 | x       | x       |         |         |      |
 | cosmwasm_1_4 | x       |         |         |         |      |
+| cosmwasm_2_0 |         |         |         |         |      |
```
webmaster128, line 40: Could you add a column that includes cosmwasm_2_0 for wasmd >= 0.51?

tests/e2e/reflect_helper.go
```diff
@@ -16,9 +18,18 @@ import (
 	"github.com/CosmWasm/wasmd/x/wasm/types"
 )
 
-// InstantiateReflectContract store and instantiate a reflect contract instance
+// InstantiateStargateReflectContract stores and instantiates a pre 2.0 reflect contract instance.
+// This instance still expects the old CosmosMsg.Stargate variant instead of the new CosmosMsg.Any.
+func InstantiateStargateReflectContract(t *testing.T, chain *ibctesting.TestChain) sdk.AccAddress {
+	codeID := chain.StoreCodeFile("../../x/wasm/keeper/testdata/stargate_reflect.wasm").CodeID
```
webmaster128, line 24: Woudn't it be easier to manage if we used the file name `reflect_1_4.wasm` here (and download file from https://github.com/CosmWasm/cosmwasm/releases/tag/v1.4.3). the file itself should not care too much why it is used here in the tests. No strong opinion on 1.2, 1.3, 1.4, or 1.5 – just something old.

x/wasm/keeper/api.go
```diff
@@ -14,6 +16,8 @@ const (
 	DefaultGasCostHumanAddress = 5
 	// DefaultGasCostCanonicalAddress is how much SDK gas we charge to convert to a canonical address format
 	DefaultGasCostCanonicalAddress = 4
+	// DefaultGasCostValidateAddress is how much SDK gas we charge to validate an address
+	DefaultGasCostValidateAddress = 9
```
webmaster128, line 20:
```suggestion
	DefaultGasCostValidateAddress = DefaultGasCostHumanAddress + DefaultGasCostCanonicalAddress
```

does this work too?

x/wasm/keeper/api.go
```diff
@@ -22,25 +26,39 @@ const (
 var (
 	costHumanize            = DefaultGasCostHumanAddress * types.DefaultGasMultiplier
 	costCanonical           = DefaultGasCostCanonicalAddress * types.DefaultGasMultiplier
+	costValidate            = DefaultGasCostCanonicalAddress * types.DefaultGasMultiplier
```
webmaster128, line 29:
```suggestion
	costValidate            = DefaultGasCostValidateAddress * types.DefaultGasMultiplier
```

x/wasm/keeper/api.go
```diff
@@ -22,25 +26,39 @@ const (
...
+func validateAddress(human string) (uint64, error) {
+	canonicalized, err := sdk.AccAddressFromBech32(human)
+	if err != nil {
+		return costValidate, err
+	}
+	// AccAddressFromBech32 already calls VerifyAddressFormat, so we can just humanize and compare
+	if canonicalized.String() != human {
+		return costValidate, errors.New("address not normalized")
+	}
+	return costValidate, err
```
webmaster128, line 57: The end of the function is supposed to be the success case, right?

```suggestion
	return costValidate, nil
```

x/wasm/keeper/api.go
```diff
@@ -22,25 +26,39 @@ const (
...
 	return sdk.AccAddress(canon).String(), costHumanize, nil
 }
 
-func canonicalAddress(human string) ([]byte, uint64, error) {
+func canonicalizeAddress(human string) ([]byte, uint64, error) {
 	bz, err := sdk.AccAddressFromBech32(human)
 	return bz, costCanonical, err
 }
 
+func validateAddress(human string) (uint64, error) {
```
webmaster128, line 48: Do we have a place to write unit tests for those functions?

webmaster128 (commented):
Good stuff! Not done reviwwing but here is a start

chipshort:
The test-system job seems to fail sporadically. Is it just a timing thing?

x/wasm/keeper/api.go
chipshort: Nice catch!

x/wasm/keeper/api.go
chipshort: Currently not, but I can add an `api_test.go` and add tests there.

webmaster128: Yeah that would be great. At least for the new addition

chipshort: I added a very basic test here.

pinosu:
> The test-system job seems to fail sporadically. Is it just a timing thing?

Yes, unfortunately since we added the system test for the upgrade, sporadically it fails for timeout

webmaster128:
If you look at the logs and the implementation of `AwaitBlockHeight` we see that we wait for block number 21, current block is 9 and block time is 1s with a timeout of 14s. Finalizing 12 one-second blocks within 14 seconds is very optimistic given that the blockTime parameter is used for `"--commit-timeout=" + s.blockTime.String(),`. But commit timeout is not the block time but a lower bound for the block time (https://docs.tendermint.com/v0.34/tendermint-core/configuration.html).  We use e.g.

```
			// Consensus settings
			config.Consensus.TimeoutPropose = 2000 * time.Millisecond
			config.Consensus.TimeoutProposeDelta = 500 * time.Millisecond
			config.Consensus.TimeoutPrecommitDelta = 500 * time.Millisecond
			config.Consensus.TimeoutPrevote = 1 * time.Second
			config.Consensus.TimeoutPrecommit = 1 * time.Second
			config.Consensus.TimeoutPrecommitDelta = 500 * time.Millisecond
			config.Consensus.TimeoutCommit = 1750 * time.Millisecond
```

to get block times of 2.11s in production or

```
# Custom settings for very fast blocks in CI
# We target 250ms block times with one validator.
sed -i"" \
  -e 's/^timeout_propose =.*$/timeout_propose = "100ms"/' \
  -e 's/^timeout_propose_delta =.*$/timeout_propose_delta = "100ms"/' \
  -e 's/^timeout_prevote =.*$/timeout_prevote = "100ms"/' \
  -e 's/^timeout_prevote_delta =.*$/timeout_prevote_delta = "100ms"/' \
  -e 's/^timeout_precommit =.*$/timeout_precommit = "100ms"/' \
  -e 's/^timeout_precommit_delta =.*$/timeout_precommit_delta = "100ms"/' \
  -e 's/^timeout_commit =.*$/timeout_commit = "230ms"/' \
  "config/config.toml"
```

to aim for 250ms block times in CI. So I think the `--commit-timeout` should be something like 90% of the desired block time.

x/wasm/types/errors.go
```diff
@@ -10,6 +10,10 @@ import (
 var (
 	DefaultCodespace = ModuleName
 
+	// ContractErrorCodes are the error codes for errors returned by the contract
+	// Since contract execution is deterministic, the errors are also deterministic
+	ContractErrorCodes = []uint32{InstantiateErrorCode, ExecuteErrorCode, QueryErrorCode, MigrateErrorCode}
```
webmaster128, line 15: I think with this approach the chances are high that one of
- `ErrInstantiateFailed`
- `ErrExecuteFailed`
- `ErrQueryFailed`
- `ErrMigrationFailed`

is used in more than one place. E.g. `ErrExecuteFailed` is used many times throughout the codebase for vaious different things. `ErrMigrationFailed` is used in two places as well.

I think the safer approach is to explicitly mark errors as "Contract error" (as in the error case of

```go
// ContractResult is the raw response from the instantiate/execute/migrate calls.
// This is mirrors Rust's ContractResult<Response>.
type ContractResult struct {
	Ok  *Response `json:"ok,omitempty"`
	Err string    `json:"error,omitempty"`
}
```

in the places where we know for sure they are coming from the contract. Everything else by default remains redacted.

This would also play nicely with https://github.com/CosmWasm/wasmd/issues/1813 where we get migrations that don't even call into the contract, such that the idea of a "migration error" can be different things on two different levels.

webmaster128: The term "contract error" is also used here: https://github.com/CosmWasm/cosmwasm/pull/1646

x/wasm/keeper/msg_dispatcher.go
```diff
@@ -189,6 +215,12 @@ func redactError(err error) error {
 	// sdk/5 is insufficient funds (on bank send)
 	// (we can theoretically redact less in the future, but this is a first step to safety)
 	codespace, code, _ := errorsmod.ABCIInfo(err, false)
+
+	// Also do not redact any errors that are coming from the contract,
+	// as they are always deterministic
+	if codespace == types.DefaultCodespace && contains(types.ContractErrorCodes, code) {
+		return err
```
webmaster128, line 222: Should we unwrap what was wrapped before? Not sure if this is possible though

x/wasm/types/errors.go
chipshort: That makes sense. So, we would add a new `ContractError` type and check against that? I was trying to do that at first, but I was not sure how to add the error codes and codespace with that.
After looking at it again, we could handle this in a similar way as the `WasmVMFlavouredError`.

webmaster128: > So, we would add a new ContractError type and check against that?

That was my first idea. However, I think I missed that the error must work well in the two scenarios:
1. created in a submessage and handled in reply
2. created in a transaction requiring code and codespace

> After looking at it again, we could handle this in a similar way as the WasmVMFlavouredError.

Interesting, I did not look into that before. However, I wonder if there is a way to attach some sort of "is deterministic" property to the normal Cosmos SDK error type that we can inspect in  `redactError` to keep it simpler.

chipshort: I think that should be possible. I'll experiment a bit.

chipshort: I added a proposal in 4e7e53422c1b49fcaf482c61524cba0d88abe625. It's a wrapper type around any error that marks it as deterministic for `redactError`. I only marked the errors coming from the contract as that for now (in theory most errors produced by wasmd should be deterministic as long as we don't change the message between versions.

x/wasm/types/errors.go
```diff
@@ -164,3 +152,32 @@ func (e WasmVMFlavouredError) Wrap(desc string) error { return errorsmod.Wrap(e,
...
+	return DeterministicError{error: e}
+}
+
+// Unwrap implements the built-in errors.Unwrap
+func (e DeterministicError) Unwrap() error {
+	return e.error
+}
+
+// Cause is the same as unwrap but used by ABCIInfo
+// By returning the wrapped error here, we ensure
```
webmaster128, line 180: This sentence seems broken

x/wasm/types/errors.go
```diff
@@ -164,3 +152,32 @@ func (e WasmVMFlavouredError) Wrap(desc string) error { return errorsmod.Wrap(e,
...
+func MarkErrorDeterministic(e error) DeterministicError {
+	return DeterministicError{error: e}
+}
+
+// Unwrap implements the built-in errors.Unwrap
+func (e DeterministicError) Unwrap() error {
+	return e.error
+}
+
+// Cause is the same as unwrap but used by ABCIInfo
```
webmaster128, line 179: Is there an interface we can check this implementation against?

x/wasm/keeper/msg_dispatcher.go
```diff
@@ -209,18 +209,19 @@ func redactError(err error) error {
 		return err
 	}
 
+	// If it is a DeterministicError, we can safely return it without redaction.
+	// We only check the top level error to avoid changes in the error chain becoming
+	// consensus-breaking.
+	if _, ok := err.(types.DeterministicError); ok {
+		return err
```
webmaster128, line 216: Should we unwrap here?

x/wasm/types/errors_test.go
```diff
@@ -94,3 +94,19 @@ func TestWasmVMFlavouredError(t *testing.T) {
 		t.Run(name, spec.exec)
 	}
 }
+
+func TestDeterministicError(t *testing.T) {
+	inner := ErrInstantiateFailed
+	err := MarkErrorDeterministic(inner)
+
+	// behaves like a wrapper around inner error
+	assert.Equal(t, inner.Error(), err.Error())
```
webmaster128, line 103: Could we add a sentence for this property in the `DeterministicError` docs? I think this is very powerful feature to have.

webmaster128 (commented):
Very very nice solution

x/wasm/types/errors.go
webmaster128: Love it

> I only marked the errors coming from the contract as that for now (in theory most errors produced by wasmd should be deterministic as long as we don't change the message between versions).

We can add more and more errors over time which makes this solution extra powerful and elegant. Changing error text is usually not consensus breaking and happens in Cosmos SDK patch releases. But we can highlight that some error texts must be stable to get the extra benefit of having them non-redacted in the replies.

x/wasm/types/errors.go
chipshort: Unfortunately not. ABCIInfo has a [private interface](https://github.com/cosmos/cosmos-sdk/blob/35c4243da9c8c285a4b65117238dec9b9f7921c8/errors/errors.go#L300) for that and uses it to [iterate through the error chain](https://github.com/cosmos/cosmos-sdk/blob/35c4243da9c8c285a4b65117238dec9b9f7921c8/errors/abci.go#L92) until it finds an error with codespace / code.

x/wasm/types/errors.go
chipshort: oops, I just stopped mid-sentence and forgot about it 😆

x/wasm/keeper/msg_dispatcher.go
chipshort: We could, but it doesn't make much of a difference since the DeterministicError has the same `Error()` message as its wrapped error.

x/wasm/types/errors.go
webmaster128: okay, fine, thanks for checking

x/wasm/keeper/msg_dispatcher.go
webmaster128: I see, thanks

chipshort:
> Do we have an integration test somewhere that shows how a submessage error is returned to a contract unredacted?

Good point, I added a test now that instantiates two reflect instances and makes one send a submessage that fails to the other one. It then checks the error message the first contract got.

x/wasm/keeper/relay.go
```diff
@@ -194,7 +200,11 @@ func (k Keeper) OnAckPacket(
 	if execErr != nil {
 		return errorsmod.Wrap(types.ErrExecuteFailed, execErr.Error())
 	}
-	return k.handleIBCBasicContractResponse(ctx, contractAddr, contractInfo.IBCPortID, res)
+	if res.Err != "" {
```
pinosu, line 203: should we add a check that `res != nil`?

chipshort: Right now wasmvm only returns `nil` in case of an `err`, but it's probably better to add a check here anyways, just in case.

PR #1816, opened 2024-02-29
Fix misleading error message for duplicate address on instantiate2

webmaster128:
Chaning the label does not help avoiding the collision. Only creator, checksum and salt is typically used.


chipshort (approved):
Nice comment 👍

pinosu (approved):
LGTM 👍

PR #1817, opened 2024-02-29
IBC Callbacks

chipshort:


tests/e2e/ibc_callbacks_test.go
```diff
@@ -0,0 +1,200 @@
...
+	// and on chain B
+	codeIDonB := chainB.StoreCodeFile("./testdata/ibc_callbacks.wasm").CodeID
+	contractAddrB := chainB.InstantiateContract(codeIDonB, []byte(`{}`))
+	require.NotEmpty(t, contractAddrB)
+
+	// ExecuteMsg is the ibc-callbacks contract's execute msg
+	type ExecuteMsg struct {
+		ToAddress      string             `json:"to_address"`
+		ChannelID      string             `json:"channel_id"`
+		TimeoutSeconds wasmvmtypes.Uint64 `json:"timeout_seconds"`
```
webmaster128, line 71: Since we only need this for the difference from now in seconds, we can also jut use uint64 and JSON numbers for this field. Even uint32 would be sufficient

app/app.go
```diff
@@ -647,10 +652,10 @@ func NewWasmApp(
...
 	)
 
-	// Create Transfer Stack
-	var transferStack porttypes.IBCModule
-	transferStack = transfer.NewIBCModule(app.TransferKeeper)
-	transferStack = ibcfee.NewIBCMiddleware(transferStack, app.IBCFeeKeeper)
+	// Create fee enabled wasm ibc Stack
+	var wasmStack porttypes.IBCModule
+	wasmStackIBCHandler := wasm.NewIBCHandler(app.WasmKeeper, app.IBCKeeper.ChannelKeeper, app.IBCFeeKeeper)
+	wasmStack = ibcfee.NewIBCMiddleware(wasmStackIBCHandler, app.IBCFeeKeeper)
```
webmaster128, line 658: MaxIBCCallbackGas is not supposed to be used for this stack?

webmaster128 (commented):
🐎

app/app.go
chipshort: No, this is our module, wrapped in the ibcfee middleware (not callbacks). The diff just looks a bit confusing on this.

We could implement support for the callbacks middleware at some point in the future though. But that requires some more thought on how it should work.

tests/e2e/ibc_callbacks_test.go
chipshort: Done

tests/e2e/ibc_callbacks_test.go
```diff
@@ -0,0 +1,200 @@
...
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/CosmWasm/wasmd/app"
+	"github.com/CosmWasm/wasmd/tests/e2e"
+	wasmibctesting "github.com/CosmWasm/wasmd/x/wasm/ibctesting"
+	"github.com/CosmWasm/wasmd/x/wasm/types"
+)
+
+func TestIBCCallbacks(t *testing.T) {
```
srdtrk, line 27: nit: This test would be more readable if you turned it into a table test. So that test cases are clearly labeled with expSuccess/expError/expTimeout and etc.

srdtrk (commented):
Thank you for this amazing work.

app/app.go
chipshort: Not sure I understand. The fee middleware is already in main for a long time. I only moved the wasm stack up because I need a reference to the `wasmStackIBCHandler` for the other stacks.

x/wasm/ibc.go
```diff
@@ -326,22 +327,146 @@ func (i IBCHandler) OnTimeoutPacket(ctx sdk.Context, packet channeltypes.Packet,
...
+	}
+	senderAddress, err := sdk.AccAddressFromBech32(senderAddr)
+	if err != nil {
+		return nil, errorsmod.Wrapf(err, "packet sender address")
+	}
+
+	// We only allow the contract that sent the message to receive source chain callbacks for it.
+	if !contractAddress.Equals(senderAddress) {
+		return nil, errorsmod.Wrapf(types.ErrExecuteFailed, "contract address %s does not match packet sender %s", contractAddr, senderAddress)
+	}
```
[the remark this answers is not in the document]
chipshort: Doesn't that fail for different casing? Or is there some guarantee that both are e.g. lower case?
But I guess, even if not I can just use `EqualFold`, so good point.

x/wasm/ibc.go
```diff
@@ -326,22 +327,146 @@ func (i IBCHandler) OnTimeoutPacket(ctx sdk.Context, packet channeltypes.Packet,
...
+
+	return contractAddress, nil
+}
+
+func newIBCPacket(packet ibcexported.PacketI) wasmvmtypes.IBCPacket {
```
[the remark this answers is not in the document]
chipshort: The `IBCReceivePacketCallback` only provides me with a `PacketI`, not a `Packet`, so to be able to use `newIBCPacket` there, I had to change it here too.

app/app.go
```diff
@@ -141,6 +142,10 @@ const appName = "WasmApp"
 var (
 	NodeDir      = ".wasmd"
 	Bech32Prefix = "wasm"
+	// MaxIBCCallbackGas is the maximum gas that an IBC callback can use.
+	// If the callback uses more gas, it will be out of gas and will not be reverted.
+	// This is passed to the callbacks middleware
+	MaxIBCCallbackGas = uint64(1000000)
```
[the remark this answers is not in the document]
chipshort: That number is relatively arbitrary. I just looked at some normal on-chain transactions and put a safety margin around that. If we want to export this to chains as a default, we need to put more thought into it.

app/app.go
srdtrk: Sorry, then everything is all good.

go.mod
```diff
@@ -46,6 +45,7 @@ require (
 	cosmossdk.io/x/upgrade v0.1.1
 	github.com/cometbft/cometbft v0.38.5
 	github.com/cosmos/cosmos-db v1.0.0
+	github.com/cosmos/ibc-go/modules/apps/callbacks v0.2.1-0.20231113120333-342c00b0f8bd
```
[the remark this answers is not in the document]
chipshort: I believe that's the commit we are using here, so yes, that works. Will keep this open as a reminder to update to the full release.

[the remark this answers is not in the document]
srdtrk: We decided to keep using git commits for now since we didn't come up with a long term solution. So you may proceed with this

x/wasm/ibc.go
```diff
@@ -326,22 +327,146 @@ func (i IBCHandler) OnTimeoutPacket(ctx sdk.Context, packet channeltypes.Packet,
...
+
+// IBCReceivePacketCallback implements the IBC Callbacks ContractKeeper interface
+// see https://github.com/cosmos/ibc-go/blob/main/docs/architecture/adr-008-app-caller-cbs.md#contractkeeper
+func (i IBCHandler) IBCReceivePacketCallback(
+	cachedCtx sdk.Context,
+	packet ibcexported.PacketI,
+	ack ibcexported.Acknowledgement,
+	contractAddress string,
+) error {
+	// sender validation makes no sense here, as the receiver is never the sender
```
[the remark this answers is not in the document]
chipshort: I added a paragraph about this to the code docs in cosmwasm-std and will make sure to mention this in our currently ongoing documentation rewrite.

x/wasm/ibc.go
```diff
@@ -18,6 +20,11 @@ import (
 	"github.com/CosmWasm/wasmd/x/wasm/types"
 )
 
+// DefaultMaxIBCCallbackGas is the default value of maximum gas that an IBC callback can use.
+// If the callback uses more gas, it will be out of gas and will not be reverted.
```
webmaster128, line 24: "will not be reverted" is this correct? Sounds scary if some half done callback is not reverted

x/wasm/ibc.go
```diff
@@ -18,6 +20,11 @@ import (
...
+// Pass this to the callbacks middleware or choose a custom value.
+const DefaultMaxIBCCallbackGas = uint64(1000000)
```
webmaster128, line 26:
```suggestion
const DefaultMaxIBCCallbackGas = uint64(1_000_000)
```

x/wasm/ibc.go
```diff
@@ -326,22 +333,142 @@ func (i IBCHandler) OnTimeoutPacket(ctx sdk.Context, packet channeltypes.Packet,
...
+	}
+
+	return nil
+}
+
+func validateSender(contractAddr, senderAddr string) (sdk.AccAddress, error) {
+	// We only allow the contract that sent the message to receive source chain callbacks for it.
+	if !strings.EqualFold(contractAddr, senderAddr) {
+		return nil, errorsmod.Wrapf(types.ErrExecuteFailed, "contract address %s does not match packet sender %s", contractAddr, senderAddr)
+	}
```
webmaster128, line 445: Looks correct, but if we can avoid it I'd not do string comparisions on bech32 inputs. Can we just decode both and then compare the bytes?

webmaster128 (commented):
Looks great as far as I can tell

x/wasm/ibc.go
chipshort: Ok, the sentence is not good. What I mean is: The transaction is not reverted. The contract state is reverted though.
It basically refers to this paragraph of the [ADR-8 spec](https://github.com/cosmos/ibc-go/blob/8c79fac7e69b67ed697bc299657cd26ab2439407/docs/architecture/adr-008-app-caller-cbs.md?plain=1#L255-L256): 
> If the relayer-defined gas limit is exceeded before the user-defined gas limit or global callback gas limit is exceeded, then the entire transaction is reverted to allow for resubmission. If the chain-defined or user-defined gas limit is reached, 
the callback state changes are reverted and the transaction is committed.

I'll change the wording to explain that better.

webmaster128 (approved):
🔥

chipshort:
I rebased on main. Can be merged if CI is happy

PR #1819, opened 2024-03-03
deps: bump ecosystem libraries

faddat:
# testnet time

Hi Simon, we last spoke on Twitter and you were looking for a person/team willing to do a test network for wasmvm v2.0.0.

Just so happens I've been harboring a vision for a very, very boring permissionless CosmWasm chain, and timing has lined up well for that. 

To be on the safe side, this testnet isn't going to use vote extensions, and likely the mainnet won't either. 

If there is anything in particular that you'd like contract authors to test, please let me know :)

## Changes in this PR

* bump ibc to v8.1.0
* bump comet to v0.38.5 (the versions used here previously are retracted)
* bump cosmos-sdk to v0.50.4
* use cometbft-db v0.11.0
* use go v1.22
* update golangci-lint to v1.56.2

## reasoning for changes

Basically this is everything I'd bump right before starting a network.


go.mod
```diff
@@ -1,14 +1,14 @@
 module github.com/CosmWasm/wasmd
 
-go 1.21
+go 1.22
```
webmaster128, line 3: Why this requirement change?

PR #1842, opened 2024-03-26
Updated Fixed Potential slashing evasion during re-delegation

bangtabil:
## Summary
An issue was identified in the slashing mechanism that may allow for the evasion of slashing penalties during a slashing event. If a delegation contributed to byzantine behavior of a validator, and the validator has not yet been slashed, it may be possible for that delegation to evade a pending slashing penalty through re-delegation behavior. Additional validation logic was added to restrict this behavior.

### DIFF
--- a/tests/system/go.mod
+++ b/tests/system/go.mod
@@ -4,12 +4,12 @@ go 1.21
 
 require (
 	github.com/cosmos/cosmos-proto v1.0.0-beta.4 // indirect
-	github.com/cosmos/cosmos-sdk v0.50.4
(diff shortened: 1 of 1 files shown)


pinosu (changes requested):
Thanks for your PR. Could you please also update the `go.mod` file for the project ?
Please also run `go mod tidy` `go mod vendor` commands for both folders to update the `go.sum` files.

pinosu:
Closing since cosmos-sdk upgrade to v0.50.5 is included in #1844

PR #1846, opened 2024-03-29
[DO NOT MERGE] Allow more gas efficient migrations

### LINKED ISSUE #1813: Optional migrate entry point – allow more gas efficient migrations
When CosmWasm migrations were designed, we had 1 instance – 1 dev team in mind. The whole system assumes a small number of instances to handle. Recently we learned that more and more use cases are poping up in which many instances of a contract are created, dozens of pools, hundreds or DAOs, thousands of accounts. Having to send a `MsgMigrateContract` cand calling the contract for each message is gas intense and hard to manage.

While this ticket does no solve the full problem at once, it is a first step and actionable. The idea is this: not every migration needs a state change. Often times state changes can be done lazily or are not required at all. In those cases we can avoid calling the `migrate` entry point of the contract. Just swap out the code and call it a day.

This means not executing those lines:
https://github.com/CosmWasm/wasmd/blob/7ea00e2ea858ed599141e322bd68171998a3259a/x/wasm/keeper/keeper.go#L475-L479

In such cases the `msg` field of `MsgMigrateContract` becomes optional.

If a call to the `migrate` entry point is needed or not should be decided by the new code. There the contract author can decide if the state needs to change or not. Luckily we have `Entrypoints` in `AnalysisReport` and can implement the following logic:
- if new code contains the `migrate` entry point, require `msg` to be non-empty JSON (current logic)
- otherwise require `msg` to be empty

This avoids situations in which an expected msg is missing or a provided msg is ignored.

pinosu:
Resolves #1813

### DIFF
--- a/x/wasm/keeper/keeper.go
+++ b/x/wasm/keeper/keeper.go
@@ -466,8 +466,10 @@ func (k Keeper) migrate(
 		return nil, errorsmod.Wrap(sdkerrors.ErrUnauthorized, "to use new code")
 	}
 
+	report, err := k.wasmVM.AnalyzeCode(newCodeInfo.CodeHash)
+
 	// check for IBC flag
-	switch report, err := k.wasmVM.AnalyzeCode(newCodeInfo.CodeHash); {
+	switch {
 	case err != nil:
 		return nil, errorsmod.Wrap(types.ErrVMError, err.Error())
 	case !report.HasIBCEntryPoints && contractInfo.IBCPortID != "":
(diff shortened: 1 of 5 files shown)


x/wasm/keeper/keeper.go
```diff
@@ -482,26 +484,17 @@ func (k Keeper) migrate(
...
-		return nil, types.MarkErrorDeterministic(errorsmod.Wrap(types.ErrMigrationFailed, res.Err))
+	// check for migrate endpoint
+	hasMigrateEndpoint := false
+	switch {
+	case contains(report.Entrypoints, "migrate") && msg != nil:
+		hasMigrateEndpoint = true
+	case !contains(report.Entrypoints, "migrate") && msg == nil:
+		hasMigrateEndpoint = false
+	default:
+		return nil, errorsmod.Wrap(types.ErrMigrationFailed, "invalid migration setup")
```
chipshort, line 495: I think the error message could be a bit more descriptive. Maybe something like "invalid migration parameters". Still not great, but it's hard to word it in a way that describes both failure states ("no migrate + msg" and "migrate + no msg")

x/wasm/keeper/keeper.go
```diff
@@ -482,26 +484,17 @@ func (k Keeper) migrate(
...
-	}
-	if res.Err != "" {
-		return nil, types.MarkErrorDeterministic(errorsmod.Wrap(types.ErrMigrationFailed, res.Err))
+	// check for migrate endpoint
+	hasMigrateEndpoint := false
+	switch {
+	case contains(report.Entrypoints, "migrate") && msg != nil:
+		hasMigrateEndpoint = true
+	case !contains(report.Entrypoints, "migrate") && msg == nil:
+		hasMigrateEndpoint = false
```
webmaster128, line 493: - Wouldn't the flow become easier if we set `hasMigrateEndpoint := contains(report.Entrypoints, "migrate")` first and then check for valid combinations? No need to execute the `contains` twice.
- I would use `len(msg) == 0` for the message check since both nil and an empty byte slice means there is no message set

chipshort (approved):
LGTM

webmaster128:
Thank you! Closing in favour of this solution as discussed: #1924

PR #1855, opened 2024-04-12
Add BuiltInCapabilities() to x/wasm/keeper and deprecate AllCapabilities()

webmaster128:
`AllCapabilities()` is currently used by chains
1. without additions (https://github.com/cosmos/gaia/pull/3051/files#diff-1ac782d2fab1d48ad7f7ef06cf2d58b37c99b6b4c98879a5ad86a985b73fad58R456)
2. with additions (https://github.com/CosmosContracts/juno/blob/e98863bf7112f4b117a2114e22f7482367362764/app/keepers/keepers.go#L114)

This is good since it avoids bugs and adds newly added capabilities autimatically in an upgrade.

This PR does not change functionality but improves things in the following ways:
- Better documentation
- Better function name
- As Pino highlighted, the symbols from `app` should not be imported by chains using x/wasm

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -642,7 +642,7 @@ func NewWasmApp(
 		app.GRPCQueryRouter(),
 		wasmDir,
 		wasmConfig,
-		AllCapabilities(),
+		wasmkeeper.BuiltInCapabilities(),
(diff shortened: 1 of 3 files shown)


pinosu (approved):
Very nice!!!
LGTM 👍

PR #1876, opened 2024-05-03
Async Ack

### LINKED ISSUE #1891: Possible problem with the `nil` ack
Hello, we are using `IbcReceiveResponse::without_ack` to return a `nil` ack. But we are hitting [this](https://github.com/cosmos/ibc-go/blob/main/modules/core/04-channel/keeper/packet.go#L293) error. Our initial investigation shows us that `nil` ack returns non-nil ack with empty data. That's why we are hitting this error.

As far as I understand, this feature is meant to return a `nil` ack and cause `WriteAcknowledgement` to be deferred. Is there something that we are missing or is this a bug?

chipshort:
also fixes #1891 

I had to do changes in quite a few places


x/wasm/ibc.go
```diff
@@ -280,10 +280,16 @@ func (i IBCHandler) OnRecvPacket(
...
+	} else if ack == nil || ack.Acknowledgement() == nil {
 		// emit all contract and submessage events on success
 		// nil ack is a success case, see: https://github.com/cosmos/ibc-go/blob/v7.0.0/modules/core/keeper/msg_server.go#L453
 		ctx.EventManager().EmitEvents(em.Events())
+		// contract wants async acknowledgement, so store the packet for later
+		i.keeper.StoreAsyncAckPacket(ctx, packet)
+		ack = nil
+	} else if ack.Success() {
+		// emit all contract and submessage events on success
+		ctx.EventManager().EmitEvents(em.Events())
```
srdtrk, line 292: This might be easier to read if it is not chained after the error handling, for example something like:
```suggestion
        types.EmitAcknowledgementEvent(ctx, contractAddr, ack, err)
        return ack
	}
	
	switch {
	case ack == nil || ack.Acknowledgement() == nil:
		// emit all contract and submessage events on success
		// nil ack is a success case, see: https://github.com/cosmos/ibc-go/blob/v7.0.0/modules/core/keeper/msg_server.go#L453
		ctx.EventManager().EmitEvents(em.Events())
		// contract wants async acknowledgement, so store the packet for later
		i.keeper.StoreAsyncAckPacket(ctx, packet)
		ack = nil
	case ack.Success():
		// emit all contract and submessage events on success
		ctx.EventManager().EmitEvents(em.Events())
    default:
        panic("unreachable")
    }
```
it also has the added benefit of panicing in an unexpected scenario

x/wasm/keeper/handler_plugin.go
```diff
@@ -161,54 +163,115 @@ func (m MessageHandlerChain) DispatchMsg(ctx sdk.Context, contractAddr sdk.AccAd
...
+			return nil, nil, nil, errorsmod.Wrap(err, "failed to marshal IBC send response")
+		}
+		any, err := codectypes.NewAnyWithValue(resp)
+		if err != nil {
+			return nil, nil, nil, errorsmod.Wrap(err, "failed to convert IBC send response to Any")
+		}
+		msgResponses := [][]*codectypes.Any{{any}}
+
+		return nil, [][]byte{val}, msgResponses, nil
+	case msg.IBC.WriteAcknowledgement != nil:
```
srdtrk, line 218: I'm assuming only the contract itself can send this message

x/wasm/keeper/handler_plugin.go
```diff
@@ -161,54 +163,115 @@ func (m MessageHandlerChain) DispatchMsg(ctx sdk.Context, contractAddr sdk.AccAd
...
-	msgResponses := [][]*codectypes.Any{{any}}
+}
+
+var _ ibcexported.Acknowledgement = asyncAck{}
+
+type asyncAck wasmvmtypes.IBCAcknowledgement
+
+func (w asyncAck) Success() bool {
+	// We don't give the contract to supply a success flag because it's not part of the IBC spec.
+	// Currently, ibc-go also just ignores this flag for `WriteAcknowledgement`.
```
srdtrk, line 269: I believe this is ignored when handling an acknowledgement asynchronously, which is the case here. However, this is used if the acknowledgement is synchronous, see here. We discard the state changes due to application callbacks if this is false. Since the entire contract state is in the application callback, even if the ack was synchronous, you'd want to always return true. Right?

srdtrk (commented):
Lgtm. Added some comments :)

x/wasm/keeper/handler_plugin.go
chipshort: Yes, the success value is ignored for async acks. This type is specifically for async acks. For synchronous acks, there is a separate type `ContractConfirmStateAck`, which also just returns `true`, so I guess that could be reused here instead.

x/wasm/keeper/handler_plugin.go
chipshort: Yes, this is part of the MessageDispatcher chain that is used to execute messages returned by a contract entrypoint. It's not a normal sdk message and the ibc port id that is used here is the one of the contract who sent this.

x/wasm/ibc.go
chipshort: I like the early return and the switch, but that default case is actually reachable. That `keeper.OnRecvPacket` call translates contract errors to error ack already. Tbh, this whole split between `IBCHandler.OnRecvPacket` and `Keeper.OnRecvPacket` seems very confusing to me.

x/wasm/keeper/relay.go
```diff
@@ -169,6 +170,15 @@ func (k Keeper) OnRecvPacket(
 		// submessage errors result in error ACK with state reverted. Error message is redacted
 		return nil, err
 	}
+
+	if data == nil {
+		// contract wants async acknowledgement, so store the packet for later
+		err = k.StoreAsyncAckPacket(ctx, convertPacket(msg.Packet))
```
chipshort, line 176: I moved this call into the keeper now (was in IBCHandler before). That makes the whole thing a little easier to understand, but now we have to convert the packet back into a `channeltypes.Packet`

x/wasm/keeper/handler_plugin.go
```diff
@@ -161,54 +162,101 @@ func (m MessageHandlerChain) DispatchMsg(ctx sdk.Context, contractAddr sdk.AccAd
...
+		}
+
+		err = h.ics4Wrapper.WriteAcknowledgement(ctx, channelCap, packet, ContractConfirmStateAck(msg.IBC.WriteAcknowledgement.Ack.Data))
+		if err != nil {
+			return nil, nil, nil, errorsmod.Wrap(err, "acknowledgement")
+		}
 
-	return nil, [][]byte{val}, msgResponses, nil
+		// delete the packet from the store after acknowledgement
+		h.wasmKeeper.DeleteAsyncAckPacket(ctx, contractIBCPortID, contractIBCChannelID, msg.IBC.WriteAcknowledgement.PacketSequence)
```
webmaster128, line 242:
```suggestion
		// Delete the packet from the store after acknowledgement.
		// This ensures WriteAcknowledgement can only be used once per packet
		// such that overriding the acknowledgement later on is not possible.
		h.wasmKeeper.DeleteAsyncAckPacket(ctx, contractIBCPortID, contractIBCChannelID, msg.IBC.WriteAcknowledgement.PacketSequence)
```

x/wasm/keeper/relay.go
```diff
@@ -169,6 +170,15 @@ func (k Keeper) OnRecvPacket(
 		// submessage errors result in error ACK with state reverted. Error message is redacted
 		return nil, err
 	}
+
+	if data == nil {
+		// contract wants async acknowledgement, so store the packet for later
```
webmaster128, line 175:
```suggestion
		// Protocol might never write acknowledgement or contract
		// wants async acknowledgements, we don't know.
		// So store the packet for later.
```

x/wasm/types/keys.go
```diff
@@ -60,6 +63,23 @@ func GetContractStorePrefix(addr sdk.AccAddress) []byte {
...
 }
 
+// GetAsyncPacketKey returns the key for a packet that is acknowledged asynchronously
+func GetAsyncPacketKey(destChannel string, sequence uint64) []byte {
+	// key is a concatenation of length-prefixed destination channel and sequence
+	channel := []byte(destChannel)
+	channelLen := make([]byte, 4)
+	binary.LittleEndian.PutUint32(channelLen, uint32(len(channel)))
+	seq := make([]byte, 8)
+	binary.LittleEndian.PutUint64(seq, sequence)
```
webmaster128, line 73: Any reason for using LE instead of BE?

If we want to iterate though all packets of a given (port_id, destinationChannel) this would give us a strange order

webmaster128 (commented):
Great stuff.

I am just wondering about one thing: how to cleanup this temporary packet store at some point for the packets that get never acknowledged. Would be great to have some sort of configurable lifetime.

x/wasm/types/keys.go
chipshort: Nope, no reason other than all the example code I found being little endian. Will change to big endian

docs/proto/proto-docs.md
```diff
@@ -40,6 +40,7 @@
     - [MsgIBCCloseChannel](#cosmwasm.wasm.v1.MsgIBCCloseChannel)
     - [MsgIBCSend](#cosmwasm.wasm.v1.MsgIBCSend)
     - [MsgIBCSendResponse](#cosmwasm.wasm.v1.MsgIBCSendResponse)
+    - [MsgIBCWriteAcknowledgementResponse](#cosmwasm.wasm.v1.MsgIBCWriteAcknowledgementResponse)
```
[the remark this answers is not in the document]
chipshort: Not sure. What is that even used for?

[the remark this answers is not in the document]
webmaster128: Okay, interesting. Seems like we can remove `MsgIBCSend` and don't break anything. But that's another story. Closing here.

chipshort:
> how to cleanup this temporary packet store at some point for the packets that get never acknowledged. Would be great to have some sort of configurable lifetime.

That's a general problem with async acks, even on the ibc-go layer. If there is no ack, the packet commitment is not deleted there either.
Adding a lifetime opens a whole different can of worms. When would we clean it up?
Also, I'm not sure how a valid contract should handle that case. After deleting the stored packet, there is no way to ack the packet anymore, so that can in theory cause a packet that would otherwise have been acknowledged to get stuck.

I'm not convinced there's a good solution for this, so my take is to just live with the fact that incorrect contracts will write packets that are never cleaned up (just like they can write stuff into their own state that never gets cleaned up).

webmaster128:
> my take is to just live with the fact that incorrect contracts will write packets that are never cleaned up

Right, sounds good. In that case should we add a storage write fee to `StoreAsyncAckPacket`? Or does `prefixStore.Set(key, packetBz)` charge gas automatically?

chipshort:
Yes `prefixStore.Set(key, packetBz)` will charge that. You can see an example in the test in [relay_test.go](https://github.com/CosmWasm/wasmd/pull/1876/files#diff-44578a53fec2e61d283949550afd5632581e30744b7dc6810fbd4e103b375d67).

webmaster128 (approved):
Beautiful

PR #1894, opened 2024-06-03
bump ibc-go version to 8.3.1

jhernandezb:
Replacement of https://github.com/CosmWasm/wasmd/pull/1886 because ica host keeper now enforces the querier being passed when the MsgServer is constructed


srdtrk (approved):
thanks, lgtm. I would've opened the same PR :)

pinosu (approved):
Thanks for your PR! LGTM 💯

PR #1904, opened 2024-06-19
Mark QuerySmart contract error as deterministic

chipshort:
We apparently didn't mark this as deterministic in #1804.

### DIFF
--- a/x/wasm/keeper/contract_keeper_test.go
+++ b/x/wasm/keeper/contract_keeper_test.go
@@ -202,5 +202,5 @@ func TestQuerierError(t *testing.T) {
 	require.Error(t, err)
 
 	// we expect the contract's "reply 1 not found" to be in there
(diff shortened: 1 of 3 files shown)


x/wasm/keeper/contract_keeper_test.go
```diff
@@ -169,3 +171,36 @@ func TestInstantiate2(t *testing.T) {
...
+				},
+			},
+		},
+	}
+	query := mustMarshal(t, queryType)
+	_, err := keepers.WasmKeeper.QuerySmart(parentCtx, contract.Contract, query)
+	require.Error(t, err)
+
+	// we expect the contract's "reply 1 not found" to be in there
+	assert.Equal(t, "Generic error: Querier contract error: reply 1 not found: query wasm contract failed: query wasm contract failed", err.Error())
```
pinosu, line 205: What about having `assert.Contains(t, err.Error(), "reply 1 not found")` instead of `assert.Equal`?

pinosu (approved):
LGTM 👍 

just left a small comment that might make the test more robust

Issue #1740, opened 2023-11-26
Feature request: show text along with codespace numbers


taitruong:
afaik sei chain does this, but many Cosmos chains doesnt. Ngl, would save a lot of time for devs.
It's quite tedious figuring out all the time what these codespace errors mean:
https://github.com/CosmWasm/wasmd/blob/7ea00e2ea858ed599141e322bd68171998a3259a/x/wasm/types/errors.go

alpe:
Thanks for the feature request! 
Errors are a bigger topic. We have to redact them for contracts to not contain host specific and therefore consensus breaking information. It is also quite easy to break backwards compatibility if a contract builds upon an error message and then this text is changed for some reason. Other than this, errors are handled by the SDK for IO. 
Can you elaborate more on your scenario so that I can understand the use case better?
I could imagine a CLI command for example that can help to translate the codespace/ code into the error message...

BHIIKTOR:
i think that's what Sei did (but im not 100% sure), they used the CLI to interpret the error msgs, instead of adding them directly to the chain.

i could not agree more with @alpe solution of adding it to the CLI instead of anywhere else
this way we can keep things more agnostic at chain level

webmaster128:
There are two related topics here and we need to address them separately:

## Errors in submessage replies

Currently redacted due to consensus problems. This will change for all errors coming from CosmWasm contracts with 2.0. The change is [already implemented here](https://github.com/CosmWasm/wasmvm/pull/465) and will be integrated into wasmd soon.

## Errors in acknowledgements

`channeltypes.NewErrorAcknowledgement` from ibc-go which is used here does not even contain the codespace. We only get a code `X` and don't know if it is sdk/X or wasm/X or any other module:

```
func NewErrorAcknowledgement(err error) Acknowledgement {
	// the ABCI code is included in the abcitypes.ResponseDeliverTx hash
	// constructed in Tendermint and is therefore deterministic
	_, code, _ := errorsmod.ABCIInfo(err, false) // discard non-determinstic codespace and log values

	return Acknowledgement{
		Response: &Acknowledgement_Error{
			Error: fmt.Sprintf("ABCI code: %d: %s", code, ackErrorString),
		},
	}
}
```

This is because [codespace is not part of consesus](https://github.com/cometbft/cometbft/issues/844) for some reason.

[...]

PR #1911, opened 2024-06-27
Add codespace to error acknowledgement

webmaster128:
This adds the codespace to error acknowledgements by switching from NewErrorAcknowledgement to NewErrorAcknowledgementWithCodespace.

Also creates the initial infrastrucrture to further customize error acknowledgement creation.

Part of #1740


pinosu (approved):
LGTM! 👍 
Let's update all the tests to that CI passes and then it's good to go

PR #1915, opened 2024-06-28
Mark wasm queries with module_query_safe

### LINKED ISSUE #1903: Mark wasmd queries with `module_query_safe`
## Description

[`module_query_safe`](https://github.com/cosmos/cosmos-sdk/blob/27a231ae4816fd8c3ee39a1cc02eccf977fb1b79/proto/cosmos/query/v1/query.proto#L36) tag, introduced in CosmosSDK v0.47, is used mark queries that are deterministic and consume gas properly. In ibc-go, we allow such queries to be executed by interchain accounts. Since wasm queries are deterministic and also track their gas usage, they should be marked with this tag. This would allow wasm queries to be executed by interchain accounts and could also provide additional benefits if #1840 is implemented.

## Example

[Here](https://github.com/cosmos/cosmos-sdk/blob/v0.50.7/proto/cosmos/bank/v1beta1/query.proto#L19) is a bank query marked with this tag in CosmosSDK v0.50

pinosu:
Resolves #1903


proto/cosmwasm/wasm/v1/query.proto
```diff
@@ -17,65 +18,77 @@ service Query {
...
       returns (QueryContractsByCreatorResponse) {
+    option (cosmos.query.v1.module_query_safe) = true;
     option (google.api.http).get =
         "/cosmwasm/wasm/v1/contracts/creator/{creator_address}";
   }
 
   // BuildAddress builds a contract address
   rpc BuildAddress(QueryBuildAddressRequest)
       returns (QueryBuildAddressResponse) {
+    option (cosmos.query.v1.module_query_safe) = true;
```
chipshort, line 91: This query doesn't charge any gas. The option docs say:
> 2. has its gas tracked, to avoid the attack vector where no gas is accounted for on potentially high-computation queries.

I guess this query is so computationally cheap that it's not an issue, but would be good to check somehow.

chipshort (commented):
Looks good, just wondering about the gas

proto/cosmwasm/wasm/v1/query.proto
pinosu: good 👀 ! let's check what @webmaster128  thinks about this

webmaster128: This is for creating Instantiate2 addresses, right?

I think this query should be removed as the algorithm is stateless and fixed. There are multiple implementations in Rust, TypeScript and Go. I don't see any reason to perform a query to the node for this.

pinosu: @webmaster128 should we remove this as part of next release? 
I can create a separate pr for it

webmaster128: Let's just create a ticket as this might open a can of worms looking at the code. IMO this is not relevant enough to delay the release. We should get CosmWasm 2.1 and other fixes in wasmd out as soon as possible.

webmaster128: Would't it be easy to charge a small amount of gas here for the sake of consistency?

pinosu: yes, I think it makes sense to charge some gas for this. I will update the pr

webmaster128 (approved):

x/wasm/keeper/querier.go
```diff
@@ -428,6 +431,11 @@ func (q GrpcQuerier) BuildAddress(c context.Context, req *types.QueryBuildAddres
 	if len(salt) == 0 {
 		return nil, status.Error(codes.InvalidArgument, "empty salt")
 	}
+
+	ctx := sdk.UnwrapSDKContext(c)
+	costBuildAddress := DefaultGasCostBuildAddress * types.DefaultGasMultiplier
```
chipshort, line 436: Is this multiplied by the gas multiplier on purpose? To me the code looks like it's just using the cosmos sdk gas meter, so that expects sdk gas as input, not cosmwasm gas.

pinosu: Yes, it makes sense to remove the Multiplier because it is related to CW gas. 
I used it to make the cost in line with humanizeAddress operation. 
I will update the pr and remove it. @chipshort do you think something like 500_000 is a good value for `DefaultGasCostBuildAddress` ?

x/wasm/keeper/querier.go
```diff
@@ -25,7 +25,7 @@ import (
 )
 
 // DefaultGasCostBuildAddress is the SDK gas cost to build a contract address
-const DefaultGasCostBuildAddress = 5
+const DefaultGasCostBuildAddress = 500_000
```
webmaster128, line 28: This is a Cosmos SDK gas value where 100_000-200_00 is the gas range of a full transaction with verification, state changes, events etc. I think we need some much smaller value here.


What about 10?

pinosu: Sounds good! I will update the pr 👍

chipshort (approved):

PR #1919, opened 2024-07-02
Improve async ack coverage

chipshort:
followup of #1876


x/wasm/keeper/handler_plugin_test.go
```diff
@@ -270,19 +285,39 @@ func TestIBCRawPacketHandler(t *testing.T) {
...
+	// also store a packet to be acked
+	ackPacket := channeltypes.Packet{
+		Sequence:           1,
+		SourcePort:         "src-port",
+		SourceChannel:      "channel-0",
+		DestinationPort:    ibcPort,
+		DestinationChannel: "channel-1",
+		Data:               []byte{},
+		TimeoutHeight:      clienttypes.Height{},
+		TimeoutTimestamp:   0,
```
webmaster128, line 297: I think one of them must be set to a non-empty value, but not sure

x/wasm/keeper/handler_plugin_test.go
```diff
@@ -323,16 +384,16 @@ func TestIBCRawPacketHandler(t *testing.T) {
 			assert.Nil(t, evts)
 			require.NotNil(t, data)
 			assert.Len(t, msgResponses, 1)
-			assert.Equal(t, "/cosmwasm.wasm.v1.MsgIBCSendResponse", msgResponses[0][0].TypeUrl)
-
-			expMsg := types.MsgIBCSendResponse{Sequence: 1}
+			// assert.Equal(t, "/cosmwasm.wasm.v1.MsgIBCSendResponse", msgResponses[0][0].TypeUrl)
```
webmaster128, line 387: There should be a way to get the typeURL of `spec.expResp` via



```
// MessageName returns the fully-qualified proto name for the given message type.
func MessageName(x Message) string {
```

which is in `/Users/me/go/pkg/mod/github.com/cosmos/gogoproto@v1.5.0/proto/properties.go`

x/wasm/keeper/handler_plugin_test.go
chipshort: For actual data, one of them should be set, but here we just test that WriteAcknowledgement is called with the same packet that was saved earlier, so what that packet looks like is not that relevant, except for the parts we care about (destination port, channel and sequence)

webmaster128 (approved):
🔥

PR #1920, opened 2024-07-02
Add go mod tidy check to CI

chipshort:


.circleci/config.yml
```diff
@@ -61,6 +61,23 @@ jobs:
           paths:
             - ".git"
 
+  tidy-go:
+    executor: golang
+    steps:
+      - checkout
+      - run:
+          name: Check go mod tidy
+          # Use --check or --exit-code when available (Go 1.22?)
```
webmaster128, line 70: This is implemented now

```suggestion
          # Use --diff when available (Go 1.23)
```

webmaster128 (approved):

Issue #1912, opened 2024-06-27
Update wasmvm to 2.1


webmaster128:
- [x] Update library to RC (#1913)
- [x] expose pinned contract metrics (https://github.com/CosmWasm/wasmvm/pull/526)
- [x] IBC callbacks (#1817)
- [x] Async acks (#1876)
- [x] Add cosmwasm_2_1 to built in capabilities
- [ ] Update library to 2.1.0

PR #1922, opened 2024-07-03
Pinned Metrics

chipshort:
related to #1912

### DIFF
--- a/x/wasm/keeper/metrics.go
+++ b/x/wasm/keeper/metrics.go
@@ -62,27 +62,25 @@ func (p *WasmVMMetricsCollector) Describe(descs chan<- *prometheus.Desc) {
 // Collect is called by the Prometheus registry when collecting metrics.
 func (p *WasmVMMetricsCollector) Collect(c chan<- prometheus.Metric) {
 	m, err := p.source.GetMetrics()
-	if err != nil {
-		return
+	if err == nil {
+		c <- prometheus.MustNewConstMetric(p.CacheHitsDescr, prometheus.CounterValue, float64(m.HitsPinnedMemoryCache), labelPinned)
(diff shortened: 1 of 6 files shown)


x/wasm/keeper/metrics.go
```diff
@@ -68,6 +73,16 @@ func (p *WasmVMMetricsCollector) Collect(c chan<- prometheus.Metric) {
 	c <- prometheus.MustNewConstMetric(p.CacheElementsDescr, prometheus.GaugeValue, float64(m.ElementsMemoryCache), labelMemory)
 	c <- prometheus.MustNewConstMetric(p.CacheSizeDescr, prometheus.GaugeValue, float64(m.SizeMemoryCache), labelMemory)
 	c <- prometheus.MustNewConstMetric(p.CacheSizeDescr, prometheus.GaugeValue, float64(m.SizePinnedMemoryCache), labelPinned)
+
+	pm, err := p.source.GetPinnedMetrics()
+	if err != nil {
+		return
+	}
```
webmaster128, line 80: If errors lead to not updating the values and otherwise ignore them (which is fine), can we write the code in a way like:

1. If GetMetrics() succeeds, perform updates. Otherwise skip updates.
1. If GetPinnedMetrics() succeeds, perform updates. Otherwise skip updates.

This extends the old behaviour to more blocks

chipshort: you mean like this? (see latest commit)
Or do you mean: if one of them errors, update nothing

webmaster128: Exactly, I like. If no updates are coming to prometheus you can do the debugging. No need for an hard exit here.

webmaster128 (approved):

PR #1924, opened 2024-07-04
Migrate Version checks

chipshort:
related to #1912 and #1813


x/wasm/keeper/keeper.go
```diff
@@ -483,26 +484,57 @@ func (k Keeper) migrate(
...
+		// no need to call the migrate entrypoint if the versions are the same
+	case report.ContractMigrateVersion != nil:
+		// migrating from contract without migrate version to one with migrate version
+		// in the future, we will provide MigrateInfo to the contract in this case
+		response, err = k.callMigrateEntrypoint(sdkCtx, contractAddress, wasmvmtypes.Checksum(newCodeInfo.CodeHash), msg, setupCost)
+		if err != nil {
+			return nil, err
+		}
+	case oldReport.ContractMigrateVersion != nil:
+		// prevent migrating from contract with migrate version to one without migrate version
```
chipshort, line 524: This one might be debatable. Should we allow it?

webmaster128: What about this: the default case is that migrate is called. Only if both migrate versions are set and they are equal, migrate is skipped. That pretty much keeps the current behaviour except for one simple optimization.

I would not differentiate between upgrading and downgrading. The contract developer can handle this themselves.

x/wasm/keeper/keeper.go
```diff
@@ -526,15 +558,59 @@ func (k Keeper) migrate(
 		sdk.NewAttribute(types.AttributeKeyContractAddr, contractAddress.String()),
 	))
 
-	sdkCtx = types.WithSubMsgAuthzPolicy(sdkCtx, authZ.SubMessageAuthorizationPolicy(types.AuthZActionMigrateContract))
-	data, err := k.handleContractResponse(sdkCtx, contractAddress, contractInfo.IBCPortID, res.Ok.Messages, res.Ok.Attributes, res.Ok.Data, res.Ok.Events)
-	if err != nil {
-		return nil, errorsmod.Wrap(err, "dispatch")
+	var data []byte
+	if response != nil {
```
webmaster128, line 562:
```suggestion

	// if migrate entry point was called
	if response != nil {
```

Is this comment correct?

Dockerfile
```diff
@@ -17,10 +17,10 @@ RUN apk add git
...
 COPY . /code/
 # See https://github.com/CosmWasm/wasmvm/releases
-ADD https://github.com/CosmWasm/wasmvm/releases/download/v2.1.0-rc.1/libwasmvm_muslc.aarch64.a /lib/libwasmvm_muslc.aarch64.a
-ADD https://github.com/CosmWasm/wasmvm/releases/download/v2.1.0-rc.1/libwasmvm_muslc.x86_64.a /lib/libwasmvm_muslc.x86_64.a
-RUN sha256sum /lib/libwasmvm_muslc.aarch64.a | grep 293a7b1d8468374fc5d079f7082e83e2866729d2d80f8a2d4b0a69e06113c2c7
-RUN sha256sum /lib/libwasmvm_muslc.x86_64.a | grep 1f0d477ea71e2d0062e1091ef08ea791908fc4c73d0ef95b56326c96df07db8f
+ADD https://github.com/CosmWasm/wasmvm/releases/download/v2.1.0-rc.2/libwasmvm_muslc.aarch64.a /lib/libwasmvm_muslc.aarch64.a
+ADD https://github.com/CosmWasm/wasmvm/releases/download/v2.1.0-rc.2/libwasmvm_muslc.x86_64.a /lib/libwasmvm_muslc.x86_64.a
+RUN sha256sum /lib/libwasmvm_muslc.aarch64.a | grep c22273727e2a1b7af1968162e30087c948dbaf4f53a2dbdece1de089cba3872d
+RUN sha256sum /lib/libwasmvm_muslc.x86_64.a | grep 3ef25e229658ca3e44c0dd55426006f590d88cebb6e566c52f35cf9b4a203345
```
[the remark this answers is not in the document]
chipshort: Accidentally commited that on this branch first and forgot to remove it after I cherry-picked it for the other one.

x/wasm/keeper/keeper.go
chipshort: Yes, the comment is correct. response is only set if we called the migrate entrypoint above.

x/wasm/keeper/keeper.go
chipshort: Ok, that simplifies it a lot.

webmaster128 (approved):

PR #1935, opened 2024-07-17
Fix release step

webmaster128:
I just configured `WASMD_GITHUB_RELEASES_TOKEN` in CircleCI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -244,29 +244,14 @@ jobs:
 
   release-tagged:
     executor: golang
-    environment:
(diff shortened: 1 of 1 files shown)


chipshort (approved):
LGTM
Can we trigger the job manually somehow to get the 0.52 release or should I just create that manually?

webmaster128:
> Can we trigger the job manually somehow to get the 0.52 release or should I just create that manually?

Since I renamed the env variable, I assume the old CircieCI config will not work anymore. Would be great if you can do it manually. No binary needed.

PR #1943, opened 2024-08-01
Add CodeInfo rpc query

pinosu:


proto/cosmwasm/wasm/v1/query.proto
```diff
@@ -65,6 +65,12 @@ service Query {
     option (google.api.http).get = "/cosmwasm/wasm/v1/code";
   }
 
+  // CodeInfo gets the metadata for a single wasm code
+  rpc CodeInfo(QueryCodeInfoRequest) returns (CodeInfoResponse) {
```
webmaster128, line 69: Should we have a dedicated type here callen `QueryCodeInfoResponse` for consistency? Then we could also use the opportunity to clean it up a little bit

x/wasm/client/cli/query.go
```diff
@@ -252,20 +252,17 @@ func GetCmdQueryCodeInfo() *cobra.Command {
 			}
 
 			queryClient := types.NewQueryClient(clientCtx)
-			res, err := queryClient.Code(
+			res, err := queryClient.CodeInfo(
```
webmaster128, line 255: Oh nice, this makes this query much more efficient. No more downloading and discarding the Wasm blob

webmaster128 (commented):
Nice

cc @srdtrk

proto/cosmwasm/wasm/v1/query.proto
```diff
@@ -222,6 +222,13 @@ message QueryCodeInfoRequest {
   uint64 code_id = 1; // grpc-gateway_out does not support Go style CodeID
 }
 
+// QueryCodeInfoResponse is the response type for the Query/Code RPC method
+message QueryCodeInfoResponse {
+  option (gogoproto.equal) = true;
+  CodeInfoResponse code_info = 1
+      [ (gogoproto.embed) = true, (gogoproto.jsontag) = "" ];
```
webmaster128, line 229: Can we inline all the fields and rename `DataHash` -> `Checksum`?

proto/cosmwasm/wasm/v1/query.proto
```diff
@@ -208,7 +214,24 @@ message QuerySmartContractStateResponse {
...
-  uint64 code_id = 1; // grpc-gateway_out does not support Go style CodID
+  uint64 code_id = 1; // grpc-gateway_out does not support Go style CodeID
+}
+
+// QueryCodeInfoRequest is the request type for the Query/Code RPC method
+message QueryCodeInfoRequest {
+  uint64 code_id = 1; // grpc-gateway_out does not support Go style CodeID
+}
+
+// QueryCodeInfoResponse is the response type for the Query/Code RPC method
```
chipshort, line 225: I guess the RPC method in the doc comment should be adjusted

chipshort (commented):
Nice!

chipshort (approved):
LGTM

PR #1957, opened 2024-08-12
Fix typo in "predictable".

xavierlepretre:
Presumably this typo fix does not impact tests.

Unsure whether it impacts other repos.

### DIFF
--- a/x/wasm/keeper/addresses.go
+++ b/x/wasm/keeper/addresses.go
@@ -22,8 +22,8 @@ func (k Keeper) ClassicAddressGenerator() AddressGenerator {
 	}
 }
 
-// PredicableAddressGenerator generates a predictable contract address
-func PredicableAddressGenerator(creator sdk.AccAddress, salt, msg []byte, fixMsg bool) AddressGenerator {
+// PredictableAddressGenerator generates a predictable contract address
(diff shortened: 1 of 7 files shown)


x/wasm/keeper/addresses.go
```diff
@@ -22,8 +22,8 @@ func (k Keeper) ClassicAddressGenerator() AddressGenerator {
 	}
 }
 
-// PredicableAddressGenerator generates a predictable contract address
-func PredicableAddressGenerator(creator sdk.AccAddress, salt, msg []byte, fixMsg bool) AddressGenerator {
+// PredictableAddressGenerator generates a predictable contract address
+func PredictableAddressGenerator(creator sdk.AccAddress, salt, msg []byte, fixMsg bool) AddressGenerator {
```
webmaster128, line 26: Is this a breaking change @pinosu?

webmaster128 (commented):
Thanks!

x/wasm/keeper/addresses.go
pinosu: No, it's not a breaking change. Looks good! 💯

pinosu (approved):

PR #1965, opened 2024-08-20
Fix buf.lock commit for cosmos-sdk

SpekalsG3:
proto is dependant on `buf.build/cosmos/cosmos-sdk:v0.50.0` but `buf.lock` was never updated and was pointing to old `0.47.0`


pinosu (approved):
Thank you! 👍

PR #1966, opened 2024-08-21
Conditional migrate calling

kulikthebird:


x/wasm/keeper/keeper.go
```diff
@@ -565,7 +567,30 @@ func (k Keeper) callMigrateEntrypoint(
...
+		Sender:          senderAddress.String(),
+		OldStateVersion: OldStateVersion,
+	}
+	var res *wasmvmtypes.ContractResult
+	var gasUsed uint64
+	var err error
+	if OldStateVersion == nil {
+		res, gasUsed, err = k.wasmVM.Migrate(newChecksum, env, msg, vmStore, cosmwasmAPI, &querier, k.gasMeter(sdkCtx), gasLeft, costJSONDeserialization)
+		if err != nil {
+			if strings.Contains(err.Error(), "The called function args arity mismatch, expected 3 args") {
```
chipshort, line 581: I'm pretty sure this gas handling together with the `Contains` check for the error can be exploited.
Setup: contract has old `migrate` entrypoint and does a lot of work there, but panics with "The called function args arity mismatch, expected 3 args" at the end. The VM behavior would look like this:
1. `Migrate` is called and calls into the entrypoint. After that `gasUsed` is set to however much the contract used and `err` is something like "... panicked at 'The called function args arity mismatch, expected 3 args', src/contract.rs:169:5"
2. The error message passes both `err != nil` and the `Contains` check, so VM tries to call `Migrate2`, which fails at the arity check with very little `gasUsed`, completely discarding the previous `gasUsed`.

Solution: We should only add to `gasUsed` and subtract from `gasLeft` in between calls, or handle this by calling `k.consumeRuntimeGas(sdkCtx, gasUsed)` and `gasLeft = k.runtimeGasForContract(sdkCtx)`.

x/wasm/keeper/keeper.go
```diff
@@ -565,7 +567,30 @@ func (k Keeper) callMigrateEntrypoint(
...
+
+	migrateInfo := wasmvmtypes.MigrateInfo{
+		Sender:          senderAddress.String(),
+		OldStateVersion: OldStateVersion,
+	}
+	var res *wasmvmtypes.ContractResult
+	var gasUsed uint64
+	var err error
+	if OldStateVersion == nil {
+		res, gasUsed, err = k.wasmVM.Migrate(newChecksum, env, msg, vmStore, cosmwasmAPI, &querier, k.gasMeter(sdkCtx), gasLeft, costJSONDeserialization)
```
chipshort, line 579: I would also say we should add a comment explaining why it is okay to do the second call if the first one errors. When you don't know anything about how `Migrate` and `Migrate2` work under the hood you might mistakenly think that this would migrate twice in some circumstances.
Something like:
> Migrate and Migrate2 call the same contract entrypoint, but with different parameters, so one of them will always fail without actually calling into the contract. Because of that, it is okay to try the other one in case of an error.

x/wasm/keeper/keeper.go
```diff
@@ -565,7 +567,30 @@ func (k Keeper) callMigrateEntrypoint(
...
+			}
+		}
+	} else {
+		res, gasUsed, err = k.wasmVM.MigrateWithInfo(newChecksum, env, msg, migrateInfo, vmStore, cosmwasmAPI, &querier, k.gasMeter(sdkCtx), gasLeft, costJSONDeserialization)
+		if err != nil {
+			if strings.Contains(err.Error(), "The called function args arity does not match") {
+				res, gasUsed, err = k.wasmVM.Migrate(newChecksum, env, msg, vmStore, cosmwasmAPI, &querier, k.gasMeter(sdkCtx), gasLeft, costJSONDeserialization)
+			}
+		}
+	}
```
webmaster128, line 592: As mentioned in the call today, I suggest here to always call `k.wasmVM.MigrateWithInfo(` with the additional information and then do the error handling in the Rust land (probably libwasmvm)

x/wasm/keeper/keeper.go
```diff
@@ -565,7 +567,30 @@ func (k Keeper) callMigrateEntrypoint(
...
+			}
+		}
+	} else {
+		res, gasUsed, err = k.wasmVM.Migrate2(newChecksum, env, msg, migrateInfo, vmStore, cosmwasmAPI, &querier, k.gasMeter(sdkCtx), gasLeft, costJSONDeserialization)
+		if err != nil {
+			if strings.Contains(err.Error(), "The called function args arity mismatch, expected 2 args") {
+				res, gasUsed, err = k.wasmVM.Migrate(newChecksum, env, msg, vmStore, cosmwasmAPI, &querier, k.gasMeter(sdkCtx), gasLeft, costJSONDeserialization)
+			}
+		}
+	}
```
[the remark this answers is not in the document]
kulikthebird: Thank you! Though, I've moved the check for arity from wasmd to cosmwasm-vm, according to our yesterday's talk.

x/wasm/keeper/keeper.go
kulikthebird: This is already solved by using only the new MigrateWithInfo call

x/wasm/keeper/keeper.go
kulikthebird: Yes, that's a good solution. I've added the migrate function arity check in cosmwasm-vm, so it is handled purely on the Rust side now.

x/wasm/keeper/keeper.go
kulikthebird: I think that's solved by moving the arity check to the cosmwasm-vm. Now the type system of rust prevents such a trick - cosmwasm-vm won't call migrate second time if the contract panics with any msg string.

x/wasm/keeper/keeper.go
```diff
@@ -553,6 +553,8 @@ func (k Keeper) callMigrateEntrypoint(
 	newChecksum wasmvmtypes.Checksum,
 	msg []byte,
 	newCodeID uint64,
+	senderAddress sdk.AccAddress,
+	OldStateVersion *uint64,
```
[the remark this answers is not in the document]
kulikthebird: Done

kulikthebird:
I need to recompile the hackatom binaries to reduce the size

kulikthebird:
> Did you use the optimizer?

I updated the binaries and used the optimizer, thanks!

pinosu (approved):
Nice work! LGTM 👍

x/wasm/keeper/submsg_test.go
```diff
@@ -652,6 +652,29 @@ func TestMigrateGovSubMsgAuthzPropagated(t *testing.T) {
...
+							Wasm: &wasmvmtypes.WasmMsg{Migrate: &wasmvmtypes.MigrateMsg{
+								ContractAddr: example1.Contract.String(),
+								NewCodeID:    example2.CodeID,
+								Msg:          []byte(submsgPayload),
+							}},
+						},
+					},
+				},
+			},
+		}, 0, nil
```
chipshort, line 676: nit: Could you deduplicate this? This code and the one for MigrateFn are identical.

chipshort (commented):
LGTM

x/wasm/keeper/submsg_test.go
kulikthebird: https://github.com/CosmWasm/wasmd/pull/2022

PR #1973, opened 2024-08-28
Fix error wrapping on recursive calls

pinosu:
Now the error looks like: `dispatch: max call depth exceeded`


chipshort (approved):
LGTM, tested it locally

webmaster128:
Thank you. I'll inform the reporter

PR #1979, opened 2024-09-03
chore: wasm go.mod

tac0turtle:


tac0turtle:
there is not much value in this until the dependency graph is cleaned up. Tests are a bit mangled with e2e/integration/unit in the module so the module depends on many sdk modules.

PR #1987, opened 2024-09-11
Handle discount in setup costs for in memory cache contracts

### LINKED ISSUE #1798: Improve setup cost calculation

pinosu:
Resolves #1798

### DIFF
--- a/x/wasm/keeper/ante.go
+++ b/x/wasm/keeper/ante.go
@@ -121,6 +121,7 @@ func (g GasRegisterDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bo
 	return next(types.WithGasRegister(ctx, g.gasRegister), tx, simulate)
 }
 
+// TxContractsDecorator implements an AnteHandler that keeps track of which contracts were already accessed during the current transaction. This allows discounting further calls to those contracts, as they are likely to be in the memory cache of the VM already.
 type TxContractsDecorator struct{}
 
 // NewTxContractsDecorator constructor.
--- a/x/wasm/keeper/ante.go
+++ b/x/wasm/keeper/ante.go
@@ -128,7 +128,7 @@ func NewTxContractsDecorator() *TxContractsDecorator {
 	return &TxContractsDecorator{}
 }
 
-// AnteHandle adds the gas register to the context.
+// AnteHandle initializes a new TxContracts object to the context.
 func (d TxContractsDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (sdk.Context, error) {
 	txContracts := types.NewTxContracts()
 	return next(types.WithTxContracts(ctx, txContracts), tx, simulate)
--- a/x/wasm/keeper/ante.go
+++ b/x/wasm/keeper/ante.go
@@ -120,3 +120,16 @@ func NewGasRegisterDecorator(gr types.GasRegister) *GasRegisterDecorator {
 func (g GasRegisterDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (sdk.Context, error) {
 	return next(types.WithGasRegister(ctx, g.gasRegister), tx, simulate)
 }
+
+type TxContractsDecorator struct{}
+
+// NewTxContractsDecorator constructor.
+func NewTxContractsDecorator() *TxContractsDecorator {
+	return &TxContractsDecorator{}
+}
+
+// AnteHandle adds the gas register to the context.
+func (d TxContractsDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (sdk.Context, error) {
+	txContracts := types.NewTxContracts()
+	return next(types.WithTxContracts(ctx, txContracts), tx, simulate)
+}
--- a/x/wasm/keeper/keeper.go
+++ b/x/wasm/keeper/keeper.go
@@ -1167,20 +1167,20 @@ func (k Keeper) IsPinnedCode(ctx context.Context, codeID uint64) bool {
 }
 
 func (k Keeper) checkDiscountEligibility(ctx sdk.Context, checksum []byte, isPinned bool) (sdk.Context, bool) {
-	sdkCtx := sdk.UnwrapSDKContext(ctx)
 	if isPinned {
-		return sdkCtx, true
+		return ctx, true
 	}
 
 	txContracts, ok := types.TxContractsFromContext(ctx)
-	if !ok {
-		txContracts = types.NewTxContracts()
+	if !ok || txContracts.GetContracts() == nil {
+		// this should never happen because tx contracts are initialized in ante handler
+		panic("tx contracts must not be nil")
 	} else if txContracts.Exists(checksum) {
-		return sdkCtx, true
+		return ctx, true
 	}
 
 	txContracts.AddContract(checksum)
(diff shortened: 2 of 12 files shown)


x/wasm/keeper/keeper.go
```diff
@@ -640,8 +648,8 @@ func (k Keeper) reply(ctx sdk.Context, contractAddress sdk.AccAddress, reply was
 		return nil, err
 	}
 
-	// always consider this pinned
-	replyCosts := k.gasRegister.ReplyCosts(true, reply)
+	ctx, discount := k.checkDiscountEligibility(ctx, string(codeInfo.CodeHash), k.IsPinnedCode(ctx, contractInfo.CodeID))
+	replyCosts := k.gasRegister.ReplyCosts(discount, reply)
```
chipshort, line 652: discount should always be `true` here, right? I think the new implementation consumes more gas because the `IsPinnedCode` call checks the store.

chipshort (commented):
Nice work! Can we test somehow that the TxContracts are reset correctly?

x/wasm/keeper/keeper.go
webmaster128: agreed, here we can just say `discount := true`

x/wasm/types/types.go
```diff
@@ -429,3 +429,18 @@ func (a AccessConfig) AllAuthorizedAddresses() []string {
 	}
 	return []string{}
 }
+
+type TxContracts map[string]struct{}
```
webmaster128, line 433: Can we make this map private such that we can change it any time later? Maybe as a private field of `TxContracts`

x/wasm/types/types.go
```diff
@@ -429,3 +429,18 @@ func (a AccessConfig) AllAuthorizedAddresses() []string {
...
+
+func NewTxContracts() TxContracts {
+	return make(TxContracts, 0)
+}
+
+func (a TxContracts) AddContract(checksum string) {
+	a[checksum] = struct{}{}
+}
+
+func (a TxContracts) Exists(checksum string) bool {
```
webmaster128, line 443: I would make those arguments of type `[]byte` and do the conversion to string internally. This can then be one of:
- string() cast
- hex encoding
- base64 encoding
- Copy to array `[32]byte`

No need to change the encoding right now but good to have options later on

webmaster128 (commented):
🙌

x/wasm/types/types.go
```diff
@@ -429,3 +430,29 @@ func (a AccessConfig) AllAuthorizedAddresses() []string {
 	}
 	return []string{}
 }
+
+type txContracts map[string]struct{}
+
+type TxContracts struct {
+	contracts txContracts
+}
```
webmaster128, line 438: Nice! Could you add unit tests for `TxContracts` to ensure the existence checks are working

x/wasm/types/types.go
```diff
@@ -429,3 +430,29 @@ func (a AccessConfig) AllAuthorizedAddresses() []string {
 	}
 	return []string{}
 }
+
+type txContracts map[string]struct{}
+
+type TxContracts struct {
+	contracts txContracts
```
webmaster128, line 437:
```suggestion
type TxContracts struct {
    // contracts contains the contracts (identified by checksum) which have already been executed in a transaction
	contracts txContracts
```

x/wasm/keeper/keeper.go
```diff
@@ -1158,6 +1166,23 @@ func (k Keeper) IsPinnedCode(ctx context.Context, codeID uint64) bool {
 	return ok
 }
 
+func (k Keeper) checkDiscountEligibility(ctx sdk.Context, checksum []byte, isPinned bool) (sdk.Context, bool) {
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
```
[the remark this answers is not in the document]
pinosu: No we don't need this. I will remove 👍

x/wasm/keeper/ante.go
```diff
@@ -120,3 +120,16 @@ func NewGasRegisterDecorator(gr types.GasRegister) *GasRegisterDecorator {
...
+type TxContractsDecorator struct{}
+
+// NewTxContractsDecorator constructor.
+func NewTxContractsDecorator() *TxContractsDecorator {
+	return &TxContractsDecorator{}
+}
+
+// AnteHandle initializes a new TxContracts object to the context.
+func (d TxContractsDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (sdk.Context, error) {
+	txContracts := types.NewTxContracts()
```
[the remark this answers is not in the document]
pinosu: yes, AnteHandlers are executed before each transaction.

x/wasm/keeper/keeper.go
```diff
@@ -1158,6 +1166,23 @@ func (k Keeper) IsPinnedCode(ctx context.Context, codeID uint64) bool {
...
+	if isPinned {
+		return sdkCtx, true
+	}
+
+	txContracts, ok := types.TxContractsFromContext(ctx)
+	if !ok {
+		txContracts = types.NewTxContracts()
```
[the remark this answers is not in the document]
pinosu: yeah, this should never happen basically. I think you are right, we can just ignore it and print a warning.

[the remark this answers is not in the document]
webmaster128: okay, cool. In this case I would avoid the `txContracts = types.NewTxContracts()` here as it can obfuscate some buggy setup

chipshort (commented):
Nice work! Just one doc comment then it's good to merge.

chipshort (approved):
LGTM 💐

PR #1989, opened 2024-09-13
Add customizable WasmLimits

### LINKED ISSUE #1986: Pass `Config` to wasmvm and make `WasmLimits` available to contract dev
requires #1981

* provide a way to x/wasm users to pass in the `Config` implemented in https://github.com/CosmWasm/wasmvm/issues/559
* provide a `WasmLimitsConfig` query for the `WasmLimits` inside the config that returns it as a json string such that users have a way to get the config from the chain. This is needed such that contract developers can [provide it to cosmwasm-check
](https://github.com/CosmWasm/cosmwasm/blob/3b15bfba6406cac2f807b9af93077c04d27e79c5/packages/check/src/main.rs#L46C84-L46C100).
  * Note: If there is a better way to do this than a query or we rename this then that's fine, but we need to adjust the cosmwasm-check help text.

chipshort:
closes #1986


chipshort:
obviously not mergeable yet, but would love some feedback from you @pinosu

proto/cosmwasm/wasm/v1/query.proto
```diff
@@ -317,6 +324,15 @@ message QueryContractsByCreatorResponse {
...
 }
 
+// QueryWasmLimitsConfigRequest is the request type for the
+// Query/WasmLimitsConfig RPC method.
+message QueryWasmLimitsConfigRequest {}
+
+// QueryWasmLimitsConfigResponse is the response type for the
+// Query/WasmLimitsConfig RPC method. It contains the JSON encoded limits for
+// static validation of Wasm files.
+message QueryWasmLimitsConfigResponse { string config = 1; }
```
pinosu, line 334: Do we have a specific struct that we can define for config or we want to keep it flexible?
In that case, we could evaluate using bytes instead of string.  WDYT?

x/wasm/keeper/test_common.go
```diff
@@ -811,3 +811,8 @@ func keyPubAddr() (crypto.PrivKey, sdk.AccAddress) {
 	addr := sdk.AccAddress(pub.Address())
 	return key, addr
 }
+
+// ptr returns a pointer to the value passed in
```
pinosu, line 815: To my knowledge this convention is not so used in Go.
Maybe we could remove it, if it is not used in many places. WDYT?

pinosu (commented):
Nice work!
I'm still thinking if it's best to have the wasmLimits in wasmConfig or add a new paramenter to the NewKeeper func . I guess we need to think about the scope of the two things and if they can match or just have similar names.

x/wasm/keeper/test_common.go
chipshort: We can remove it, but then the test gets much more verbose because I have to create a variable first for every value. Well, I guess it's not as bad for my specific test, since I use the same value for all of the fields in my test, so I'll remove it.

pinosu: yeah, I thought so as well.. but since all the values are the same we can just create one variable.
If needed we can introduce this piece of code in the future.

proto/cosmwasm/wasm/v1/query.proto
chipshort: As mentioned in our chat, I want to keep it flexible. The reason for string instead of bytes is that I want to make it easy for contract devs to copy and paste the limits. They need pass them into cosmwasm-check. See here: https://github.com/CosmWasm/cosmwasm/blob/cf413c5ad6a58a87e0be894584f01506f3b2e0af/packages/check/src/main.rs#L41-L50

x/wasm/types/types.go
```diff
@@ -315,8 +315,14 @@ func NewWasmCoins(cosmosCoins sdk.Coins) (wasmCoins []wasmvmtypes.Coin) {
 	return wasmCoins
 }
 
-// WasmConfig is the extra config required for wasm
-type WasmConfig struct {
+// VMConfig contains configurations that are passed on to CosmWasm VM.
```
pinosu, line 318: Nice!

pinosu (commented):
Great work!
Should we also define `DefaultVMConfig()`, with the limits set o some reasonable default value?

chipshort:
> Should we also define `DefaultVMConfig()`, with the limits set o some reasonable default value?

This is handled in cosmwasm-vm because I want to avoid chain developers forgetting to set one of the values to the default. That's why the values all use pointers. If you pass a `nil` ptr, cosmwasm-vm uses the default value.

chipshort:
I rebased this on main, removing the commit with the `replace` for the local wasmvm dependency. Only changes other than that are in the latest commit (just some changes to adjust to the renamings in wasmvm)

pinosu (approved):
Great work! LGTM 👍

PR #1999, opened 2024-09-29
Custom AuthorizationPolicy with code & contract info

codehans:
CosmWasm integration with THORChain requires a different model for permissioning than standard `AccessConfig` via `x/gov`:

- Code upload & instantiate permissions will be configured via "Mimir" (node-set configuration parameters), and will be restricted based on code checksum. This is in order to be able to provide clear accountability from the raw wasm code, through git commit, audit, compilation checksum, through to on-chain permissions. 
- Current design does not allow the `create` function to discriminate permissions based on the actual code being uploaded, instantiated (or contracts being modified).
- This PR extends the `AuthorizationPolicy interface` loaded in `msg_server.go` to have the `CodeInfo` passed during `CanCreateCode` and `CanInstantiateContract` checks. 
- It also adds `ContractInfo` where appropriate, for a more uniform and flexible interface
- All existing behaviour is retained
- As `selectAuthorizationPolicy` is currently fixed, this PR also adds a `customAuthPolicy func(ctx context.Context, actor string) (types.AuthorizationPolicy, bool)` value to the `Keeper` struct, which can be set in `NewKeeper` with an `Option`
- This allows chains to write custom `AuthorizationPolicy` implementations, accessing state outside x/wasm etc

### DIFF
--- a/x/wasm/keeper/authz_policy.go
+++ b/x/wasm/keeper/authz_policy.go
@@ -10,20 +10,28 @@ var _ types.AuthorizationPolicy = DefaultAuthorizationPolicy{}
 
 type DefaultAuthorizationPolicy struct{}
 
-func (p DefaultAuthorizationPolicy) CanCreateCode(chainConfigs types.ChainAccessConfigs, actor sdk.AccAddress, contractConfig types.AccessConfig) bool {
+func (p DefaultAuthorizationPolicy) CanCreateCode(checksum []byte, chainConfigs types.ChainAccessConfigs, actor sdk.AccAddress, contractConfig types.AccessConfig) bool {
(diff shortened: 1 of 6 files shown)


pinosu:
Thanks for your PR.
The use case seems interesting. I just want to show another possibility we have at the moment implemented in wasmd for similar use case:
You can use the command:
```
wasmd tx wasm grant store-code \
    wasm1427kpq9mmnfp0madk5axh2ukmjgpfh6skzdxku \
    3F4CD47C39C57FE1733FB41ED176EEBD9D5C67BAF5DF8A1EEDA1455E758F8514:everybody
```
to grant the permission to an address to store a code with specific checksum and specific instantiate permission.
Could this kind of mechanism work for you as well?

PR #2030, opened 2024-11-07
Layer Build

ethanfrey:
This is currently based on v0.53.0 from Cosmwasm/wasmd.
It should occasionally be rebased on latest official release, and only maintain minimal changes for our build environment.

You should be able to build a layer-flavored image to run on testnets from here.
Docker images should be auto-built from CI.


ethanfrey:
oops. meant to make this on the fork

PR #2031, opened 2024-11-08
Add autofix protobuf job

### LINKED ISSUE #1967: [CI] Add job to check for mismatches between protobuf and generated code
It's easy to forget updating the protobuf generated code when one does not know about that.
A CI job that generates everything and checks for no diff will prevent that

Even better: In the cosmwasm repo we now use autofix.ci to automatically update all generated files. Maybe that's an option here too.

chipshort:
closes #1967

Steps for the job are taken out of the Makefile's `proto-all` job.
Should we also add a `proto-lint` check to CI or does that already exist somewhere?

### DIFF
--- a/.github/workflows/autofix.yml
+++ b/.github/workflows/autofix.yml
@@ -11,10 +11,10 @@ jobs:
     runs-on: ubuntu-latest
     container:
       image: ghcr.io/cosmos/proto-builder:0.14.0
-      options: --user root
+      options: --user root # workaround for a problem during actions/checkout, see https://github.com/actions/checkout/issues/1014
     steps:
       - uses: actions/checkout@v4
-      - name: Fix permissions
+      - name: Fix permissions # needed because of root user fix above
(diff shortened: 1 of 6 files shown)


pinosu (changes requested):
Nice work!
I think we don't have proto linting covered anywhere else, so it would be great if you could add it in here.

Also I noticed we can update the `Makefile` in such a way that we use the same protoImage also for `proto-lint` and `proto-check-breaking` by replacing `@$(DOCKER_BUF)` with `@$(protoImage) buf` .
This will make it straightforward also for the CI

chipshort:
Ok, it seems the problem in 42e330575fcabdd21ae3eb81863eea2e4fd144ee and 1e50ff7b46a465377cb2faaa5e96bed6333850b3 was caused by the fact that the CI doesn't run on this branch, but the rebased-on-main version of it. Rebasing on main fixed it.

pinosu (approved):
LGTM 👍

PR #2032, opened 2024-11-08
Add typo check to CI

chipshort:
CI config is copied from https://github.com/CosmWasm/cosmwasm/blob/main/.github/workflows/typo-check.yml


_typos.toml
```diff
@@ -0,0 +1,8 @@
+[default]
+extend-ignore-re = ["(?Rm)^.*(#|//)\\s*spellchecker:disable-line$"]
+
+[default.extend-identifiers]
+iNdEx = "iNdEx"
+
+[files]
+extend-exclude = ["**/go.mod", "**/go.sum", "**/*.pb.go"]
```
chipshort, line 8: I added some excludes here to avoid checking package hashes and generated files. There was already one false-positive in `go.sum`.

pinosu (approved):
Nice to have this in wasmd! LGTM 👍

PR #2035, opened 2024-11-11
Alex/upgrade sdk 0.52 updates

alpe:
It builds now with most tests passing.

Please note that the dependencies are not setup proper yet. The SDK is not tagged and ibc-go is not upgraded.


pinosu (approved):
LGTM 👍 
Thank you Alex!

PR #2048, opened 2024-12-03
Create test.yml

faddat:
I noticed that the pointer issue is not resolved, so I absolutely must, in the strongest possible terms, urge you to merge this pull request because the issue is still there.

For these tests, passing 95% of the time should be treated as passing 0% of the time.

# note

I noticed that the references came from wasmvm.  So I am making this PR be the actual test, and it will use the latest wasmvm from `main`.

If this runs the test incorrectly, please let me know.  

Frankly, I don't trust wasmvm, and never have, due to its use of cgo, which is nigh on impossible to debug.


faddat:
Indeed.  Sorry I forgot to close this :)

It totally does work and that's super

gday

PR #2052, opened 2024-12-05
chore: Add tests for messages execution order

kulikthebird:


kulikthebird:
Related PR with implementation of the replier contract: https://github.com/CosmWasm/cosmwasm/pull/2306

chipshort (commented):
Really nice! The test cases all look correct to me.
If it's easily possible to make this a table test, that would be great.

chipshort (approved):
Nice! Looks good to me

PR #2053, opened 2024-12-06
Update 0.3x branch to wasmvm 1.5.5

chipshort:
That should allow chains stuck on this version to at least receive security updates for wasmvm.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,10 +15,10 @@ RUN apk add git
 WORKDIR /code
 COPY . /code/
 # See https://github.com/CosmWasm/wasmvm/releases
-ADD https://github.com/CosmWasm/wasmvm/releases/download/v1.3.0/libwasmvm_muslc.aarch64.a /lib/libwasmvm_muslc.aarch64.a
-ADD https://github.com/CosmWasm/wasmvm/releases/download/v1.3.0/libwasmvm_muslc.x86_64.a /lib/libwasmvm_muslc.x86_64.a
-RUN sha256sum /lib/libwasmvm_muslc.aarch64.a | grep b1610f9c8ad8bdebf5b8f819f71d238466f83521c74a2deb799078932e862722
-RUN sha256sum /lib/libwasmvm_muslc.x86_64.a | grep b4aad4480f9b4c46635b4943beedbb72c929eab1d1b9467fe3b43e6dbf617e32
+ADD https://github.com/CosmWasm/wasmvm/releases/download/v1.5.5/libwasmvm_muslc.aarch64.a /lib/libwasmvm_muslc.aarch64.a
+ADD https://github.com/CosmWasm/wasmvm/releases/download/v1.5.5/libwasmvm_muslc.x86_64.a /lib/libwasmvm_muslc.x86_64.a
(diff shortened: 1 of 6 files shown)


webmaster128:
Nice!

Could you add a comment to https://github.com/CosmWasm/wasmd/blob/releases/v0.3x/app/wasm.go that the `cosmwasm_1_4` capability is not implemented on this branch?

--------------

> I see some changes there were never released.

Very well possible. It was unclear how much Cosmos SDK 0.45 was needed.

> Is the `releases/v0.3x` branch correct?

I think so but let us double check with Pino.

pinosu (approved):
LGTM 👍 

yes this is the correct branch. Last release was v0.33.0 and then we stopped backporting fixes. There are few commits after that but never released a new version

app/wasm.go
```diff
@@ -4,6 +4,10 @@ package app
 // See https://github.com/CosmWasm/cosmwasm/blob/main/docs/CAPABILITIES-BUILT-IN.md
 // This functionality is going to be moved upstream: https://github.com/CosmWasm/wasmvm/issues/425
 func AllCapabilities() []string {
+	// This branch contains the 0.33.x version line. It was updated from wasmvm v1.3.0 to v1.5.x
+	// to be able to provide security updates.
+	// Therefore this deliberately does not include "cosmwasm_1_4" to not enable the features
+	// guarded by that capability, even though it is theoretically available.
```
webmaster128, line 10: The features are not available in wasmd since we need to add code to the codebase to make them work. This is the part we don't plan to do for resource constraints.

chipshort: True, I adjusted the comment

webmaster128 (approved):
Nice, thank you!

Dockerfile
```diff
@@ -15,10 +15,10 @@ RUN apk add git
 WORKDIR /code
 COPY . /code/
 # See https://github.com/CosmWasm/wasmvm/releases
-ADD https://github.com/CosmWasm/wasmvm/releases/download/v1.3.0/libwasmvm_muslc.aarch64.a /lib/libwasmvm_muslc.aarch64.a
-ADD https://github.com/CosmWasm/wasmvm/releases/download/v1.3.0/libwasmvm_muslc.x86_64.a /lib/libwasmvm_muslc.x86_64.a
-RUN sha256sum /lib/libwasmvm_muslc.aarch64.a | grep b1610f9c8ad8bdebf5b8f819f71d238466f83521c74a2deb799078932e862722
-RUN sha256sum /lib/libwasmvm_muslc.x86_64.a | grep b4aad4480f9b4c46635b4943beedbb72c929eab1d1b9467fe3b43e6dbf617e32
+ADD https://github.com/CosmWasm/wasmvm/releases/download/v1.5.5/libwasmvm_muslc.aarch64.a /lib/libwasmvm_muslc.aarch64.a
```
webmaster128, line 18: Oh, we should make that [1.5.6](https://github.com/CosmWasm/wasmvm/releases/tag/v1.5.6) now

webmaster128 (changes requested):

Dockerfile
chipshort: Good catch!

webmaster128 (approved):

PR #2057, opened 2024-12-12
Run tests 20 times

pinosu:


pinosu:
This is not meant to be merged, but just to show that current main contain the fix to the pointer problem raised in #2011 .
Closing as all tests are passing.

PR #2065, opened 2024-12-18
Fix AcceptListGrpcQuerier concurrency issues

### LINKED ISSUE #2051: AcceptListGrpcQuerier is not concurrent-safe
# AcceptListGrpcQuerier is not concurrent-safe

## Issue description

The [AcceptListGrpcQuerier](https://github.com/CosmWasm/wasmd/blob/0ef57e797c8ebe5ad140f32d80ae277bef9ea971/x/wasm/keeper/query_plugins.go#L347) is not concurrent-safe. This leads to the state when one query changes the result of another. Or even worse, when a query changes the state of the transaction that uses the query  (leads to consensus failure).

The `AcceptListGrpcQuerier` accepts the map of the `[request.Path]Response` that is a singleton for the app. Hence there is a chance, when you query a smart contract (which uses the GRPC query to query the chain) to get the result of the query which is processing in parallel.

Example with step-by-step comments:

```go
// 2 goroutines read bank balance of acc1 and acc2 concurrently, so they both get BalanceResponse{}
protoResponse, accepted := acceptList[request.Path]
if !accepted {
return nil, wasmvmtypes.UnsupportedRequest{Kind: fmt.Sprintf("'%s' path is not allowed from the contract", request.Path)}
}

// the rote is also the same for both goroutines
route := queryRouter.Route(request.Path)
if route == nil {
return nil, wasmvmtypes.UnsupportedRequest{Kind: fmt.Sprintf("No route to query '%s'", request.Path)}
}

// the requests are different, so the res is different for each goroutine
res, err := route(ctx, &abci.RequestQuery{
Data: request.Data,
Path: request.Path,
})
if err != nil {
return nil, err
}

// but here, becasue of the race condition, one goroutine can get the result of another goroutine
return ConvertProtoToJSONMarshal(codec, protoResponse, res.Value)
```

Test that proves the statement:

```go
package keeper_test

import (
  "fmt"
  "sync/atomic"
  "testing"

  abci "github.com/cometbft/cometbft/abci/types"
  "github.com/cosmos/cosmos-sdk/baseapp"
  "github.com/cosmos/cosmos-sdk/codec"
  banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
  "github.com/stretchr/testify/require"
  "golang.org/x/sync/errgroup"

  "github.com/CosmWasm/wasmd/app"
  "github.com/CosmWasm/wasmd/x/wasm/keeper"
  wasmvmtypes "github.com/CosmWasm/wasmvm/v2/types"
)

const (
  denom1 = "denom1"
  denom2 = "denom2"
)

var _ keeper.GRPCQueryRouter = mockedQueryRouter{}

type mockedQueryRouter struct {
  codec codec.Codec
}

func (m mockedQueryRouter) Route(_ string) baseapp.GRPCQueryHandler {
  return func(ctx sdk.Context, req *abci.RequestQuery) (*abci.ResponseQuery, error) {
    balanceReq := &banktypes.QueryBalanceRequest{}
    if err := m.codec.Unmarshal(req.Data, balanceReq); err != nil {
      return nil, err
    }

    coin := sdk.NewInt64Coin(balanceReq.Denom, 1)
    balanceRes := &banktypes.QueryBalanceResponse{
      Balance: &coin,
    }

    resValue, err := m.codec.Marshal(balanceRes)
    if err != nil {
      return nil, err
    }

    return &abci.ResponseQuery{
      Value: resValue,
    }, nil
  }
}

func TestGRPCQuerier(t *testing.T) {
  acceptedQueries := keeper.AcceptedQueries{
    "/bank.Balance": &banktypes.QueryBalanceResponse{},
  }
  cdc := app.MakeEncodingConfig(t).Codec
  router := mockedQueryRouter{
    codec: cdc,
  }
  querier := keeper.AcceptListStargateQuerier(acceptedQueries, router, cdc)

  eg := errgroup.Group{}
  //eg.SetLimit(1) // uncomment to check that the test passes with a single goroutine
  errorsCount := atomic.Uint64{}
  for range 50 {
    for _, denom := range []string{denom1, denom2} {
      denom := denom // copy
      eg.Go(func() error {
        queryReq := &banktypes.QueryBalanceRequest{
          Address: keeper.RandomBech32AccountAddress(t),
          Denom:   denom,
        }
        grpcData, err := cdc.Marshal(queryReq)
        if err != nil {
          return err
        }

        wasmGrpcReq := &wasmvmtypes.StargateQuery{
          Data: grpcData,
          Path: "/bank.Balance",
        }

        wasmGrpcRes, err := querier(sdk.Context{}, wasmGrpcReq)
        if err != nil {
          return err
        }

        queryRes := &banktypes.QueryBalanceResponse{}
        if err := cdc.UnmarshalJSON(wasmGrpcRes, queryRes); err != nil {
          return err
        }

        expectedCoin := sdk.NewInt64Coin(denom, 1)
        if queryRes.Balance == nil {
          fmt.Printf(
            "Error: expected %s, got nil\n",
            expectedCoin.String(),
          )
          errorsCount.Add(1)
          return nil
        }
        if queryRes.Balance.String() != expectedCoin.String() {
          fmt.Printf(
            "Error: expected %s, got %s\n",
            expectedCoin.String(),
            queryRes.Balance.String(),
          )
          errorsCount.Add(1)
          return nil
        }

        return nil
      })
    }
  }

  require.NoError(t, eg.Wait())
  require.Zero(t, errorsCount.Load())
}

```

The output is:

```log
=== RUN   TestGRPCQuerier
Error: expected 1denom2, got nil
Error: expected 1denom1, got 1denom2
Error: expected 1denom2, got nil
Error: expected 1denom1, got nil
Error: expected 1denom1, got 1denom2
Error: expected 1denom2, got 1denom1
Error: expected 1denom2, got nil
Error: expected 1denom1, got nil
Error: expected 1denom1, got 1denom2
Error: expected 1denom1, got nil
Error: expected 1denom1, got nil
Error: expected 1denom1, got nil
Error: expected 1denom2, got nil
Error: expected 1denom2, got 1denom1
Error: expected 1denom1, got 1denom2
Error: expected 1denom2, got nil
Error: expected 1denom1, got nil
Error: expected 1denom1, got nil
Error: expected 1denom2, got nil
Error: expected 1denom2, got nil
Error: expected 1denom1, got 1denom2
Error: expected 1denom1, got nil
Error: expected 1denom2, got 1denom1
Error: expected 1denom2, got nil
Error: expected 1denom2, got 1denom1
Error: expected 1denom2, got nil
Error: expected 1denom1, got 1denom2
Error: expected 1denom1, got nil
Error: expected 1denom2, got 1denom1
Error: expected 1denom1, got 1denom2
Error: expected 1denom2, got nil
Error: expected 1denom1, got nil
Error: expected 1denom1, got 1denom2
Error: expected 1denom1, got nil
Error: expected 1denom2, got 1denom1
Error: expected 1denom1, got 1denom2
Error: expected 1denom1, got nil
Error: expected 1denom1, got 1denom2
Error: expected 1denom1, got 1denom2
    query_plugins_test.go:121:
        	Error Trace:	/go/src/github.com/CosmWasm/wasmd/x/wasm/keeper/query_plugins_test.go:121
        	Error:      	Should be zero, but was 39
        	Test:       	TestGRPCQuerier
```

## Solutions

* The best solution for the issue is to change the signature of the `AcceptedQueries`, and instead of the `proto.Message`, use the `func()proto.Message` to use a unique proto response per-goroutine/request.

```json
AcceptedQueries map[string]func()proto.Message
```

* Or, add a synchronization (locks), which is not the best solution, because it can be a bottleneck for the app.

* Or, add a deep copy of the proto object before using it, which is not the best solution, because it might not work with all possible implementations of the `proto.Message` interface.

pinosu:
Resolves #2051

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -51,6 +51,7 @@ require (
 	github.com/distribution/reference v0.5.0
 	github.com/rs/zerolog v1.33.0
 	github.com/spf13/viper v1.19.0
+	golang.org/x/sync v0.10.0
 	google.golang.org/genproto/googleapis/api v0.0.0-20240814211410-ddb44dafa142
 )
 
@@ -201,7 +202,6 @@ require (
 	golang.org/x/exp v0.0.0-20240404231335-c0f41cb1a7a0 // indirect
 	golang.org/x/net v0.30.0 // indirect
 	golang.org/x/oauth2 v0.23.0 // indirect
-	golang.org/x/sync v0.10.0 // indirect
 	golang.org/x/sys v0.28.0 // indirect
 	golang.org/x/term v0.27.0 // indirect
 	golang.org/x/text v0.21.0 // indirect
--- a/tests/integration/query_plugin_integration_test.go
+++ b/tests/integration/query_plugin_integration_test.go
@@ -952,8 +952,8 @@ func TestAcceptListStargateQuerier(t *testing.T) {
 
 	addrs := app.AddTestAddrsIncremental(wasmApp, ctx, 2, sdkmath.NewInt(1_000_000))
 	accepted := wasmKeeper.AcceptedQueries{
-		"/cosmos.auth.v1beta1.Query/Account": &authtypes.QueryAccountResponse{},
-		"/no/route/to/this":                  &authtypes.QueryAccountResponse{},
+		"/cosmos.auth.v1beta1.Query/Account": func() proto.Message { return &authtypes.QueryAccountResponse{} },
+		"/no/route/to/this":                  func() proto.Message { return &authtypes.QueryAccountResponse{} },
 	}
 
 	marshal := func(pb proto.Message) []byte {
--- a/x/wasm/keeper/query_plugins.go
+++ b/x/wasm/keeper/query_plugins.go
@@ -346,7 +346,7 @@ func RejectGrpcQuerier(ctx sdk.Context, request *wasmvmtypes.GrpcQuery) (proto.M
 // WithQueryPlugins(&QueryPlugins{Grpc: AcceptListGrpcQuerier(acceptList, queryRouter, codec)})
 func AcceptListGrpcQuerier(acceptList AcceptedQueries, queryRouter GRPCQueryRouter, codec codec.Codec) func(ctx sdk.Context, request *wasmvmtypes.GrpcQuery) (proto.Message, error) {
 	return func(ctx sdk.Context, request *wasmvmtypes.GrpcQuery) (proto.Message, error) {
-		protoResponse, accepted := acceptList[request.Path]
+		protoResponseFn, accepted := acceptList[request.Path]
 		if !accepted {
 			return nil, wasmvmtypes.UnsupportedRequest{Kind: fmt.Sprintf("'%s' path is not allowed from the contract", request.Path)}
 		}
@@ -364,6 +364,7 @@ func AcceptListGrpcQuerier(acceptList AcceptedQueries, queryRouter GRPCQueryRout
 			return nil, err
 		}
 
+		protoResponse := protoResponseFn()
 		// decode the query response into the expected protobuf message
 		err = codec.Unmarshal(res.Value, protoResponse)
 		if err != nil {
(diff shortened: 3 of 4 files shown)


chipshort (approved):
LGTM, nice work

PR #2075, opened 2024-12-26
Backport SimulationStoreCode fix

pinosu:
Blocked by https://github.com/CosmWasm/wasmvm/issues/599

### DIFF
--- a/x/wasm/keeper/keeper.go
+++ b/x/wasm/keeper/keeper.go
@@ -154,16 +154,29 @@ func (k Keeper) create(ctx sdk.Context, creator sdk.AccAddress, wasmCode []byte,
 	}
 
 	ctx.GasMeter().ConsumeGas(k.gasRegister.CompileCosts(len(wasmCode)), "Compiling wasm bytecode")
-	checksum, err = k.wasmVM.StoreCode(wasmCode)
+	isSimulation, ok := types.ExecModeSimulation(ctx)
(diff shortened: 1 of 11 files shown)


x/wasm/keeper/keeper.go
```diff
@@ -154,16 +154,30 @@ func (k Keeper) create(ctx sdk.Context, creator sdk.AccAddress, wasmCode []byte,
 	}
 
 	ctx.GasMeter().ConsumeGas(k.gasRegister.CompileCosts(len(wasmCode)), "Compiling wasm bytecode")
-	checksum, err = k.wasmVM.StoreCode(wasmCode)
-	if err != nil {
-		return 0, checksum, sdkerrors.Wrap(types.ErrCreateFailed, err.Error())
+	isSimulation, ok := types.ExecModeSimulation(ctx)
+	if !ok {
+		k.Logger(ctx).Info("cannot getexec mode from context")
```
chipshort, line 159: Should this be a warning instead? If this case happens, your nodes can still be attacked.

chipshort (commented):
Great work!

x/wasm/keeper/keeper.go
chipshort: Oh, I see. Alright, in that case info probably fits the best.

chipshort (approved):

PR #2077, opened 2024-12-30
Keep compatibility for context key

pinosu:


chipshort:
I didn't know this is something we should keep stable across versions. Does anyone depend on the exact values?

pinosu:
> I didn't know this is something we should keep stable across versions. Does anyone depend on the exact values?

Context keys are set as an enumeration so I think it is good to not mix keys that have different scopes but rather create a new one for new scope.
This topic anyway will be irrelevant in future releases because from cosmos-sdk v0.52 onwards the sdk context is deprecated and  also this storage.

chipshort (approved):
Alright, then let's keep them stable

PR #2090, opened 2025-01-12
refactor: using maps.Clone to simplify the code

zhuhaicity:
This is a [new function](https://pkg.go.dev/maps@go1.21.1#Clone) added in the go1.21 standard library, which can make the code more concise and easy to read.


pinosu (approved):
LGTM 👍

PR #2097, opened 2025-01-17
chore: Cleanup in IBC tests

kulikthebird:
Start using ibc-go ibctesting module instead of our custom one. Wasm related part were copy-pasted from the previous solution along with some minor updates


tests/wasmibctesting/utils.go
```diff
@@ -0,0 +1,574 @@
...
+	return fmt.Errorf("packet commitment does not exist on either endpoint for provided packet")
+}
+
+// RelayAndAckPendingPackets sends pending packages from path.EndpointA to the counterparty chain and acks
+func RelayAndAckPendingPackets(chainA *WasmTestChain, chainB *WasmTestChain, path *ibctesting.Path) error {
+	// get all the packet to relay src->dest
+	src := path.EndpointA
+	require.NoError(chainA, src.UpdateClient())
+	chainA.Logf("Relay: %d Packets A->B, %d Packets B->A\n", len(*chainA.PendingSendPackets), len(*chainB.PendingSendPackets))
+	for _, v := range *chainA.PendingSendPackets {
```
chipshort, line 312: I think this function now implicitly assumes that `chainA` and `chainB` are `path.EndpointA` and `path.EndpointB` (or the other way around). Can we validate that somehow?

chipshort (commented):
Nice, just wondering about the todo comments and if we can make sure the changed test functions cannot be used incorrectly accidentally.

tests/wasmibctesting/utils.go
```diff
@@ -0,0 +1,574 @@
...
+	toSend, _ := ibctesting.ParsePacketsFromEvents(channeltypes.EventTypeSendPacket, result.Events)
+	// require.NoError(chain, err)
+	if len(toSend) > 0 {
+		// Keep a queue on the chain that we can relay in tests
+		*chain.PendingSendPackets = append(*chain.PendingSendPackets, toSend...)
+	}
+}
+
+// TODO tkulik: Przy CreateChannels trzeba przechwycić pakiety w SendMesgs
+func (chain *WasmTestChain) OverrideSendMsgs(msgs ...sdk.Msg) (*abci.ExecTxResult, error) {
```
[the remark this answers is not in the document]
kulikthebird: As above - already implemented

tests/wasmibctesting/utils.go
kulikthebird: It was implemented that way before (unless I made a mistake). I tried not to change the current behavior of the test env

tests/wasmibctesting/utils.go
```diff
@@ -0,0 +1,574 @@
...
+	return nil
+}
+
+// TimeoutPendingPackets returns the package to source chain to let the IBC app revert any operation.
+// from A to B
+func TimeoutPendingPackets(coord *ibctesting.Coordinator, chainSrc *WasmTestChain, path *ibctesting.Path) error {
+	src := path.EndpointA
+	dest := path.EndpointB
+
+	toSend := chainSrc.PendingSendPackets
```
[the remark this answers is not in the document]
kulikthebird: This was also implemented the way it was done before this change. One thing that need to be clarified here:
`chainSrc` is of type `wasmibctesting.WasmTestChain` and `src.Chain` is of type `ibctesting.TestChain`. In the proposed solution the `WasmTestChain` contains a ref to `ibctesting.TestChain`, so in this context they points to the same test chain. `WasmTestChain` is a wrapper around `ibctesting.TestChain` that contains an extra field - `PendingSendPackets`. This is why we need the `chainSrc` in this method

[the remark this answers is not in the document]
chipshort: Before this change, the function signature looked like this:
https://github.com/CosmWasm/wasmd/blob/04cb6e5408cc54c27247b0b327dfa99769d5103c/tests/ibctesting/coordinator.go#L280
The new signature takes an additional parameter `chainSrc *WasmTestChain`. I get that this is necessary because of the need to access the pending packets in the new `WasmTestChain` type. My point is that it was previously not possible to misspecify the source chain (because the function didn't allow you to pass it in), but now it's possible because you can pass in a `chainSrc` such that `chainSrc.TestChain != src.Chain`.
I guess it's a bit of a nit, since all of this is testing utils anyways and all the existing tests are written correctly, but I think it would be nice to add something along the lines of:
`require.Equal(chainA, chainSrc.TestChain, src.Chain, "chainSrc needs to be endpoint A of the path given to TimeoutPendingPackets")`

tests/wasmibctesting/utils.go
chipshort: This is analogous to the case I commented above. It used to take only a `path *ibctesting.Path` parameter (and the `Coordinator` ofc), but now the test writer passes in `chainA *WasmTestChain, chainB *WasmTestChain`, so you can pass in chains that have nothing to do with the path you are passing in.
I guess it's less likely that someone would do that though, compared to the case above.
The behavior itself stayed the same (which is good). Just that the new signature allows someone writing a test to shoot themselves into the foot in a new way.

tests/wasmibctesting/utils.go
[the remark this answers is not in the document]
kulikthebird: I've added a new type `WasmPath`, that wraps the `ibctesting.Path`. Now the method does not need the test chain as an argument

tests/wasmibctesting/utils.go
kulikthebird: As above - solved by introducing a new wrapper type

tests/integration/relay_pingpong_test.go
```diff
@@ -252,7 +251,7 @@ func (p player) IBCPacketReceive(_ wasmvm.Checksum, _ wasmvmtypes.Env, msg wasmv
 	nextValue := p.incrementCounter(lastBallSentKey, store)
 	newHit := NewHit(p.actor, nextValue)
 	respHit := &wasmvmtypes.IBCMsg{SendPacket: &wasmvmtypes.SendPacketMsg{
-		ChannelID: packet.Src.ChannelID,
+		ChannelID: packet.Dest.ChannelID,
```
chipshort, line 254: Wait a minute. Was this wrong all along? Nice catch! Why did it even work in the first place?

tests/wasmibctesting/utils.go
```diff
@@ -0,0 +1,588 @@
...
+func NewWasmPath(chainA, chainB *WasmTestChain) *WasmPath {
+	return &WasmPath{
+		Path:   *ibctesting.NewPath(chainA.TestChain, chainB.TestChain),
+		chainA: chainA,
+		chainB: chainB,
+	}
+}
+
+// RelayAndAckPendingPackets sends pending packages from path.EndpointA to the counterparty chain and acks
+func RelayAndAckPendingPackets(path *WasmPath) error {
```
chipshort, line 321: Nice solution

chipshort (approved):
Looks good to me. Nice work!

tests/integration/relay_pingpong_test.go
chipshort: Ah, that makes sense. Our ibctesting module was probably using a counter per chain while the ibc-go one uses a global one.
Anyway, this change is the correct one, since the pingpong mock is receiving the packet here, so the channel id on its side is the destination channel of the packet.

PR #2101, opened 2025-01-21
Refactor build address cmd

pinosu:


chipshort (approved):
Nice 👍

webmaster128 (approved):
For the record: this change was needed to avoid an unused query connection to a node

PR #2102, opened 2025-01-23
feat: ICS20 Transfer v2 message

kulikthebird:


chipshort (commented):
Nice work! Could you add a new test case to `TestEncodeIbcMsg`?
Also, of course we need to wait for the wasmvm release to merge to main.
Maybe we should setup a development branch here and in wasmvm for this and eureka?

kulikthebird:
Support for ICS-20 v2 has been aborted and is not currently planned by the IBC team.

Issue #2001, opened 2024-10-03
Deprecate `SmartQueryGasLimit`


chipshort:
We should slowly phase it out because it makes the `SmartContractState` query non-deterministic.
However, RPC nodes currently rely on this to prevent people from querying contracts with infinite loops.
We should redirect them to the [Cosmos SDK query gas limit](https://github.com/cosmos/cosmos-sdk/pull/16239) instead.

To summarize:
- Deprecate `SmartQueryGasLimit`
- Log a warning when it is provided (either in CLI or in config)

PR #2103, opened 2025-01-27
Adjust `SmartContractState` query gas limit

chipshort:
Currently this always puts the limit to `queryGasLimit`, but we should not do that if the remaining gas in the context is less than that.
This is a first step towards #2001


pinosu (approved):
Nice! LGTM 👍

PR #2111, opened 2025-01-31
Make `ContractAddrLen` variable

chipshort:
This should allow chains to customize the address length for contract addresses (both classic and instantiate2).
It's related to this issue: https://github.com/CosmWasm/cosmwasm/issues/2155#issuecomment-2134034689
Injective implemented this for their fork (they use 20).


pinosu (approved):
LGTM 👍

PR #2115, opened 2025-02-03
Add gas cost for `UnpackAny`

chipshort:
breaking for people using a custom Any encoder

Benchmark results:
```
BenchmarkUnpackAny/garbage_any-8          2235267       521.9 ns/op     136 B/op       4 allocs/op
BenchmarkUnpackAny/single_MsgExec-8       1736608       686.1 ns/op       24 B/op       2 allocs/op
BenchmarkUnpackAny/10000_MsgExec-8        1724133       689.4 ns/op       24 B/op       2 allocs/op
BenchmarkUnpackAny/100000_MsgExec-8       1723904       690.6 ns/op       24 B/op       2 allocs/op
```

So, it looks like it takes roughly 700ns
Our gas target in CosmWasm uses a target of 1e12 Gas/s = 1e3 Gas/ns. That means:
1e3 Gas/ns * 700ns = 700000 Gas (in CW Gas units)

With the default gas multiplier, it's 5 SDK gas

### DIFF
--- a/x/wasm/keeper/handler_plugin_encoders.go
+++ b/x/wasm/keeper/handler_plugin_encoders.go
@@ -213,7 +213,7 @@ func EncodeAnyMsg(unpacker codectypes.AnyUnpacker) AnyEncoder {
 		}
 		var sdkMsg sdk.Msg
 
-		ctx.GasMeter().ConsumeGas((700000*uint64(len(msg.Value)))/types.DefaultGasMultiplier, "unpacking AnyMsg")
+		ctx.GasMeter().ConsumeGas(700000/types.DefaultGasMultiplier, "unpacking AnyMsg")
 		if err := unpacker.UnpackAny(&codecAny, &sdkMsg); err != nil {
 			return nil, errorsmod.Wrap(types.ErrInvalidMsg, fmt.Sprintf("Cannot unpack proto message with type URL: %s", msg.TypeURL))
 		}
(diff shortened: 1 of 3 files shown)


x/wasm/keeper/handler_plugin_encoders.go
```diff
@@ -206,12 +206,14 @@ func EncodeStakingMsg(sender sdk.AccAddress, msg *wasmvmtypes.StakingMsg) ([]sdk
...
 func EncodeAnyMsg(unpacker codectypes.AnyUnpacker) AnyEncoder {
-	return func(sender sdk.AccAddress, msg *wasmvmtypes.AnyMsg) ([]sdk.Msg, error) {
+	return func(ctx sdk.Context, sender sdk.AccAddress, msg *wasmvmtypes.AnyMsg) ([]sdk.Msg, error) {
 		codecAny := codectypes.Any{
 			TypeUrl: msg.TypeURL,
 			Value:   msg.Value,
 		}
 		var sdkMsg sdk.Msg
+
+		ctx.GasMeter().ConsumeGas((700000*uint64(len(msg.Value)))/types.DefaultGasMultiplier, "unpacking AnyMsg")
```
pinosu, line 216: what about extracting 700000 as a const and adding a short comment about what it represents (similar to what you did in the pr description)

pinosu (approved):
Great work! LGTM 👍

x/wasm/keeper/handler_plugin_encoders.go
chipshort: Nice idea! Will do that

chipshort:
Ok, I added the constant, but also fixed a bug that I forgot to push the fix for: The gas cost (at least according to my benchmark) seems pretty constant and not related to the message size (see numbers above), so it should not be multiplied with the length

PR #2120, opened 2025-02-06
running tests

josefkedwards:
Getting Juliens Suggestion implemented in .go


chipshort:
This just seems to be AI garbage

PR #2131, opened 2025-02-25
chore: edited the license badge

Olexandr88:


webmaster128:
Thank you for bringing this up. After https://github.com/CosmWasm/wasmd/pull/2145 the auto-detection now works and the change here is not needed anymore. See the README file.

PR #2139, opened 2025-02-26
chore: update ibc-go v10 to beta.1

gjermundgaraba:


app/app.go
```diff
@@ -567,7 +557,7 @@ func NewWasmApp(
 		appCodec,
 		runtime.NewKVStoreService(keys[icahosttypes.StoreKey]),
 		app.GetSubspace(icahosttypes.SubModuleName),
-		app.IBCFeeKeeper, // use ics29 fee as ics4Wrapper in middleware stack
+		app.IBCKeeper.ChannelKeeper,
 		app.IBCKeeper.ChannelKeeper,
```
webmaster128, line 561: This `app.IBCKeeper.ChannelKeeper, app.IBCKeeper.ChannelKeeper,` looks fishy

x/wasm/keeper/handler_plugin_encoders.go
```diff
@@ -327,33 +326,6 @@ func EncodeIBCMsg(portSource types.ICS20TransferPortSource) func(ctx sdk.Context
...
-			}
-			msg := &ibcfeetypes.MsgPayPacketFeeAsync{
-				PacketId: channeltypes.PacketId{
-					PortId:    msg.PayPacketFeeAsync.PortID,
-					ChannelId: msg.PayPacketFeeAsync.ChannelID,
-					Sequence:  msg.PayPacketFeeAsync.Sequence,
-				},
-				PacketFee: ibcfeetypes.NewPacketFee(fee, sender.String(), msg.PayPacketFeeAsync.Relayers),
-			}
-			return []sdk.Msg{msg}, nil
```
webmaster128, line 356: I would suggest keeping the cases here and add more specific error messages than the fallback

webmaster128 (commented):
Two first thoughts, incomplete review

app/app.go
gjermundgaraba: It is correct here though, as there is no stack (porttypes.ICS4Wrapper), so we can just use the channelkeeper as the stack.

x/wasm/keeper/handler_plugin_encoders.go
gjermundgaraba: Good point. Added that and tests for it too.

kulikthebird (commented):
I've had to change the branch base from `main` to `feat/ibc_eureka` - some extra commits from main are included now that shouldn't be the part of this PR. Other than that looks good to me. Thank you!

pinosu (approved):
Nice work! LGTM 👍

gjermundgaraba:
FYI: I just updated to the latest RC, in which there were a couple of minor changes. The "biggest" is probably callbacks getting embedded into ibc-go, so it's not a separate module in go.mod.

Take a look and let me know.

pinosu:
> FYI: I just updated to the latest RC, in which there were a couple of minor changes. The "biggest" is probably callbacks getting embedded into ibc-go, so it's not a separate module in go.mod.
> 
> Take a look and let me know.

LGTM 👍

PR #2145, opened 2025-03-04
Update LICENSE to clean copy of Apache 2.0

webmaster128:
Copied from https://www.apache.org/licenses/LICENSE-2.0.txt

Trying to address this broken auto-detection:
Bildschirmfoto 2025-03-04 um 21 37 44

Let's see if GitHub understands our license then

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,3 @@
-License: Apache2.0
 
                                  Apache License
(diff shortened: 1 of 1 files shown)


pinosu (approved):
LGTM 👍

PR #2151, opened 2025-03-06
chore: fix struct name in comment

tiaoxizhan:
fix struct name in comment

### DIFF
--- a/x/wasm/types/wasmer_engine.go
+++ b/x/wasm/types/wasmer_engine.go
@@ -277,7 +277,7 @@ type WasmEngine interface {
 		deserCost wasmvmtypes.UFraction,
 	) (*wasmvmtypes.IBCBasicResult, uint64, error)
 
-	// IBCSourceCallback is available on IBC-callbacks-enabled contracts and is called when an
+	// IBCDestinationCallback is available on IBC-callbacks-enabled contracts and is called when an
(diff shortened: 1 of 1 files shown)


x/wasm/types/wasmer_engine.go
```diff
@@ -277,7 +277,7 @@ type WasmEngine interface {
 		deserCost wasmvmtypes.UFraction,
 	) (*wasmvmtypes.IBCBasicResult, uint64, error)
 
-	// IBCSourceCallback is available on IBC-callbacks-enabled contracts and is called when an
+	// IBCDestinationCallback is available on IBC-callbacks-enabled contracts and is called when an
 	// IBC-callbacks-enabled IBC message previously sent by this contract is either acknowledged or
 	// times out.
```
chipshort, line 282: It's not just the function name. The whole comment is incorrect.

chipshort:
```suggestion
	// IBCDestinationCallback is available on IBC-callbacks-enabled contracts and is called when an
	// IBC-callbacks-enabled IBC message previously sent from a different chain is being
	// acknowledged on this chain.
```

chipshort (approved):
Thank you

x/wasm/types/wasmer_engine.go
tiaoxizhan: Thanks. Modified.

Please review again.

PR #2154, opened 2025-03-08
chore: fix the wrong field name

laterlaugh:


pinosu:
Thanks for your contribution.
This needs to be fixed first in the proto file, and then regenerate the code using `make proto-all`

pinosu (changes requested):
See the comment I left

laterlaugh:
> See the comment I left

Thanks. Modified.

Please review again.

pinosu (approved):
LGTM

PR #2172, opened 2025-03-18
Refactor: Improve error messages and type annotations in wasm types

Pronoss:
- Fixed typo in `ErrMigrationFailed` description (errors.go): updated wording from "execution" to "migration" failure
- Explicitly defined the type for `AllAccessTypes` slice (params.go)
- Replaced `pkg/errors.Wrap()` with `errorsmod.Wrap()` for consistency with Cosmos SDK error handling (params.go)

### DIFF
--- a/x/wasm/types/params.go
+++ b/x/wasm/types/params.go
@@ -12,7 +12,7 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
-var AllAccessTypes = []AccessType{
+var AllAccessTypes []AccessType = []AccessType{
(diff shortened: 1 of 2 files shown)


x/wasm/types/params.go
```diff
@@ -12,7 +12,7 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
-var AllAccessTypes = []AccessType{
+var AllAccessTypes []AccessType = []AccessType{
```
chipshort, line 15: Why? Doesn't this get inferred?

Issue #2165, opened 2025-03-13
Add e2e test for IBC2SendMsg


jawoznia:
The implementation looks good to me. On top of that we should have an e2e test that uses this utility. Let's modify the `ibc2` contract code so that it can return IBCv2 messages that will be sent from one instance to another instance of `ibc2` contract. The `ibc2_test.go` is a good starting point

_Originally posted by @kulikthebird in https://github.com/CosmWasm/wasmd/pull/2127#pullrequestreview-2678703925_

PR #2174, opened 2025-03-18
Add E2E test for IBC2SendMsg (#2165)

jawoznia:


tests/e2e/ibc2_test.go
```diff
@@ -51,15 +55,52 @@ func TestIBC2ReceiveEntrypoint(t *testing.T) {
 
 	path.Path.SetupV2()
 
-	var err error
-	timeoutTimestamp := chainA.GetTimeoutTimestampSecs()
-	packet, err := path.EndpointB.MsgSendPacket(timeoutTimestamp, mockv2.NewMockPayload(contractPortB, contractPortA))
+	// IBC v2 Payload from contract on Chain B to contract on Chain A
+	payload := mockv2.NewMockPayload(contractPortB, contractPortA)
+	type PacketMsg struct {
```
kulikthebird, line 60: I don't see where this structure is used in the contract. If it's not used, we can use the default value from the `NewMockPayload` method

kulikthebird (commented):
Great job! Some comments to check, but the whole procedure looks good

kulikthebird (approved):
:+1:

PR #2187, opened 2025-03-25
chore: fix 404 status URL

threehonor:
The existing link is no longer valid.   I replace it with the latest one.


chipshort (approved):
Thanks

PR #2188, opened 2025-03-26
refactor: remove redundant import aliases

todaymoon:
I found a lot of redundant aliases through this lint tool https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#redundant-import-alias. If possible, I would be very keen to integrate it into the CI.


todaymoon:
friendly ping~

PR #2190, opened 2025-03-26
chore: Remove unused keepers

kulikthebird:


chipshort (approved):
LGTM.
Interesting that we don't even use it.

PR #2193, opened 2025-03-27
Make `validateAddress` cheaper if canonicalization fails

chipshort:
We effectively only do the canonicalization work in this case, so the user should only have to pay for that part.


pinosu (approved):
LGTM 👍

chipshort:
@MergifyIO backport releases/v0.6x

Issue #2147, opened 2025-03-05
/cosmwasm/wasm/v1/contract/build_address query does not work


vladjdk:
The `cosmwasm/wasm/v1/contract/build_address` endpoint is broken.
Most paths have `/contract/{address}/query` so `build_address` is colliding and the router is interpreting it as an address that can't be parsed.

Example query:
https://cosmos-api.stakeandrelax.net/cosmwasm/wasm/v1/contract/build_address?code_hash=4ee07a1474cb1429cfbdba98fb52ca2efc2fe8602f8e1978dbc3f45b71511ca9&creator_address=cosmos1pgerac2dwufp7mzs4403h3je9ld4vkqngjtgdx&salt=74657374696e67

Response:
`{"code":2,"message":"decoding bech32 failed: invalid separator index -1","details":[]}`

NotJeremyLiu:
valid issue

PR #2211, opened 2025-04-09
fix(api)!: fix clashing build address query

vladjdk:
Closes: #2147

Changes the REST URL path for `build_address` and runs `make proto-gen`.


pinosu:
Closing in favor of #2373

PR #2214, opened 2025-04-14
fix: Use seconds instead of nanoseconds for IBCv2 msg

kulikthebird:


x/wasm/keeper/handler_plugin_encoders.go
```diff
@@ -360,7 +361,7 @@ func EncodeIBCv2Msg(sender sdk.AccAddress, msg *wasmvmtypes.IBC2Msg) ([]sdk.Msg,
 		}
 		msg := &channeltypesv2.MsgSendPacket{
 			SourceClient:     msg.SendPacket.ChannelID,
-			TimeoutTimestamp: msg.SendPacket.Timeout,
+			TimeoutTimestamp: uint64(time.Unix(0, int64(msg.SendPacket.Timeout)).Unix()),
```
kulikthebird, line 364: IBC-go module uses seconds as a timeout timestamp

jawoznia (approved):
Nice. LGTM!

PR #2229, opened 2025-04-28
chore: Change port prefix for IBCv2 messages to "wasm2"

kulikthebird:


x/wasm/keeper/ibc.go
```diff
@@ -24,7 +24,7 @@ func ContractFromPortID(portID string) (sdk.AccAddress, error) {
 	return sdk.AccAddressFromBech32(portID[len(portIDPrefix):])
 }
 
-const portIDPrefixV2 = "wasmV2"
+const portIDPrefixV2 = "wasm2"
```
webmaster128, line 27:
```suggestion
// The port prefix refers to "CosmWasm over IBC v2" and ensures packets are routed to the right entry points
const portIDPrefixV2 = "wasm2"
```

webmaster128 (approved):

PR #2231, opened 2025-04-28
feat: Implement IBCv2 packet sending and two entry points: Receive & Timeout 

kulikthebird:
This PR introduces significant updates to support IBCv2 Send, Receive and  Timeout entry points, including new features, refactoring, and bug fixes. The primary goal is to enable and stabilize IBCv2 message handling, encoding, and timeout management, along with supporting infrastructure and tests.

🧪 Testing
Added end-to-end test coverage for the three entry points.

🔧 Maintenance
Bumped wasmvm dependency from 2.2.3 to 3.0.0-ibc2.0.


x/wasm/keeper/ibc2.go
```diff
@@ -0,0 +1,226 @@
...
+	sourceClient string,
+	destinationClient string,
+	sequence uint64,
+	payload channeltypesv2.Payload,
+	signer sdk.AccAddress,
+) error {
+	return nil
+}
+
+func (module IBC2Handler) OnRecvPacket(
```
chipshort, line 41: Oh nice, so we have all the entrypoints already set up here?

x/wasm/keeper/handler_plugin_encoders.go
```diff
@@ -336,6 +346,31 @@ func EncodeIBCMsg(portSource types.ICS20TransferPortSource) func(ctx sdk.Context
...
+			payloads = append(payloads, channeltypesv2.Payload{
+				SourcePort:      payload.SourcePort,
+				DestinationPort: payload.DestinationPort,
+				Version:         payload.Version,
+				Encoding:        payload.Encoding,
+				Value:           payload.Value,
+			})
+		}
+		msg := &channeltypesv2.MsgSendPacket{
+			SourceClient:     msg.SendPacket.SourceClient,
```
chipshort, line 363: Do we need to validate something here or is it fine that this can be set freely by the contract?

chipshort (approved):
Didn't review all of it, but what I looked at looks fine to me.

x/wasm/keeper/handler_plugin_encoders.go
kulikthebird: ibc-go check for presence of a given source client: [packet.go => sendPacket](https://github.com/cosmos/ibc-go/blob/aef0a7a09c0b704be7f518099542b5dae6bf087d/modules/core/04-channel/v2/keeper/packet.go#L28). If there is no such client, we get an error. Also it doesn't seem that we  should prevent contracts to call any clients they need to call.

chipshort: Ah, so this determines which chain the packet will go to. Then it makes sense.

x/wasm/keeper/ibc2.go
kulikthebird: Two of them are implemented at the moment - receive and timeout. Rest returns `nil`

PR #2237, opened 2025-04-30
Fix and simplify recurse gas tests

chipshort:
These tests have been a constant annoyance for me when upgrading wasmvm. Because some of the changes in here, it was not obvious anymore how to split up the changes in gas cost, so the numbers slowly drifted away from what they should be.

This should make it a bit more clear what the expected numbers are and make it easier to update them in the future.


kulikthebird (approved):
:+1:

PR #2241, opened 2025-05-06
IBCv2 Acknowledge Receive

aumetra:


x/wasm/keeper/keeper.go
```diff
@@ -378,12 +378,6 @@ func (k Keeper) instantiate(
 		ibcPort := PortIDForContract(contractAddress)
 		contractInfo.IBCPortID = ibcPort
 	}
-	if report.HasIBC2EntryPoints {
```
kulikthebird, line 381: Have a look at [my send entry-point PR here](https://github.com/CosmWasm/wasmd/pull/2252/files#diff-4e16010ab332946722b789b22962fb5a0814d529cd8de443b925d20ccd13705f). It's important to register the contract during intantiate procedure. It'll be fixed when we update ibc-go to the latest version

kulikthebird:
> @kulikthebird Any idea how I can make the CI pass here? Technically the release binaries have been built, no?

Yep - I've just pushed a fix related to HasIBC2EntryPoints removal. There are still two things to cover:
  * gas usage has changed probably because of wasmer update
  * Docker image - we have to release a new tag of wasmvm

aumetra:
Well, I get the gas usage part.. we just need to push a new prerelease tag for wasmvm, I guess

aumetra:
Okay I fucked the commit history. Just uhh.. give me a sec

PR #2250, opened 2025-05-09
Bump cosmos-sdk v0.53.0

pinosu:


chipshort (commented):
Changes look alright to me so far, but what's happening with the CI?

kulikthebird:
I have the same problem with benchmarks tests as CI on my machine. Btw. we need that for IBCv2 implementation on 0.55.x - I guess we'd need to backport this.

app/app.go
```diff
@@ -847,19 +900,6 @@ func NewWasmApp(
...
-	if err != nil {
-		panic(err)
-	}
-	err = msgservice.ValidateProtoAnnotations(protoFiles)
-	if err != nil {
-		// Once we switch to using protoreflect-based antehandlers, we might
-		// want to panic here instead of logging a warning.
-		_, _ = fmt.Fprintln(os.Stderr, err.Error())
-	}
-
```
chipshort, line 862: Just curious: Why is this removed?

pinosu: Because it was removed from `simapp/app.go` and moved to `baseapp/baseapp.go` in cosmos-sdk . See https://github.com/cosmos/cosmos-sdk/pull/24450

pinosu:
> We should also backport this to the main branch

sure! should we get also the linter issue fixed in this PR so that when backported has already everything in place?

PR #2253, opened 2025-05-12
Fixed adding new accounts

DariuszDepta:
Fixed `setup_wasmd.sh` script, so adding new accounts is now possible like this (inside Docker container):
```shell
/opt # ./setup_and_run.sh alice bob
```


pinosu (approved):
LGTM 👍

PR #2259, opened 2025-05-14
feat: Add a transaction hash to TransactionInfo

kulikthebird:


x/wasm/types/types.go
```diff
@@ -291,7 +292,8 @@ func NewEnv(ctx sdk.Context, contractAddr sdk.AccAddress) wasmvmtypes.Env {
 		},
 	}
 	if txCounter, ok := TXCounter(ctx); ok {
-		env.Transaction = &wasmvmtypes.TransactionInfo{Index: txCounter}
+		hash := sha256.Sum256(ctx.TxBytes())
```
chipshort, line 295: Is this the same hash that is used in Cosmos SDK / CometBFT? A quick scan through the CometBFT code looks to me like they use something called [tmhash](https://github.com/cometbft/cometbft/blob/858345263c49ef2738deebcbe9bf12382571e6f0/types/tx.go#L29). Not sure if that's the correct one though. We should test that.
We want to have the same hash here that is also shown on the cli when you submit the transaction (e.g. when you run `wasmd tx wasm instantiate`) and that is used to query information about the transaction (e.g. using `wasmd q tx ...`).

chipshort: Interesting. But looks like they [truncate it](https://github.com/cometbft/cometbft/blob/809582f4f269090e51c30bd4e586a91eafea8522/crypto/tmhash/hash.go#L79).

For cross-checking locally, we can just add a line here that logs the hash, start a wasmd node and cause any transaction with an env (e.g. wasm instantiate) using cli. Then we'll see how the hash looks in the cli output and how it looks when we log it.
As for writing a real test: We can probably use cyberpunk contract's `MirrorEnv` to get back the hash in test code and do something similar to what I described above, so it would probably have to be a new system test.

webmaster128: Tendermint change the tranbsaction hast size from 20 bytes to 32 bytes at some point in 2018/2019. So question is if and where the `type sha256trunc struct` is even used. The [top level Sum function](https://github.com/cometbft/cometbft/blob/v0.38.17/crypto/tmhash/hash.go#L18-L22) does not truncate.

PR #2260, opened 2025-05-15
Corrected typos

pinkflower32:
Changes made in:

- /x/wasm/keeper/query_plugins_test.go ("expRes" → "express")

- /x/wasm/keeper/query_plugins_test.go ("expRes" → "express")

- /x/wasm/keeper/query_plugins_test.go ("expRes" → "express")


kulikthebird:
It's called that way on purpose

PR #2261, opened 2025-05-16
IBCv2 Acknowledge Receive 

aumetra:
Same as #2241 but without the fucked up commit history


aumetra:
This coverage complaint makes no sense. If it wasn't covered by tests it wouldn't make the tests fail. Fuckin' hell..

aumetra:
It's even inside the `test-cover` stage where the contract execution failed when the faulty contract was still in the test data! Ugh.. whatever.

x/wasm/keeper/ibc2.go
```diff
@@ -117,9 +117,56 @@ func (module IBC2Handler) OnAcknowledgementPacket(
...
+	}
+	if res == nil {
+		// If this gets executed, that's a bug in wasmvm
+		return errorsmod.Wrap(types.ErrVMError, "internal wasmvm error")
+	}
+	if res.Err != "" {
+		return types.MarkErrorDeterministic(errorsmod.Wrap(types.ErrExecuteFailed, res.Err))
+	}
+
+	return k.handleIBCBasicContractResponse(ctx, contractAddr, contractInfo.IBCPortID, res.Ok)
```
kulikthebird, line 167: `IBC2PortID`

tests/wasmibctesting/utils.go
```diff
@@ -679,6 +815,26 @@ func ParseAckFromEvents(events []abci.Event) ([]byte, error) {
...
 }
 
+// ParseAckFromEventsV2 parses events emitted from a MsgRecvPacket and returns the
+// acknowledgement.
+func ParseAckFromEventsV2(events []abci.Event) ([]byte, error) {
+	for _, ev := range events {
+		if ev.Type == channeltypes.EventTypeWriteAck {
+			for _, attr := range ev.Attributes {
+				if attr.Key == channeltypesv2.AttributeKeyEncodedAckHex {
+					// The first two bytes is a noise
```
kulikthebird, line 825: After deeper analysis it seems that it's not a noise. In order to properly decode the Acknowledgement type we need to do it like this:
```go
ack, err := ParseAckFromEventsV2(res.GetEvents())
if err != nil {
	return fmt.Errorf("tried to relay packet without ack but got ack")
}

var msg channeltypesv2.Acknowledgement
err = proto.Unmarshal(ack, &msg)
```

tests/wasmibctesting/utils.go
```diff
@@ -74,6 +81,26 @@ func (chain *WasmTestChain) CaptureIBCEventsV2(result *abci.ExecTxResult) {
 		// Keep a queue on the chain that we can relay in tests
 		*chain.PendingSendPacketsV2 = append(*chain.PendingSendPacketsV2, toSend...)
 	}
+
+	toAck, err := ParsePacketsFromEventsV2(channeltypesv2.EventTypeWriteAck, result.Events)
```
kulikthebird, line 85: It is not used by the sync ack test and can be removed

kulikthebird (commented):
It's good to have this endpoint implemented, well done! I've left some comments inline , otherwise LGTM

tests/wasmibctesting/utils.go
```diff
@@ -469,6 +577,34 @@ func RelayPendingPacketsV2(path *WasmPath) error {
 	return nil
 }
 
+// RelayAndAckPendingPackets sends pending packages from path.EndpointA to the counterparty chain and acks
+func RelayPendingPacketsWithAcksV2(path *WasmPath) error {
+	// get all the packet to relay src->dest
+	src := path.EndpointA
+	require.NoError(path.chainA, src.UpdateClient())
+	path.chainA.Logf("Relay: %d PacketsV2 A->B, %d PacketsV2 B->A\n", len(*path.chainA.PendingSendPacketsV2), len(*path.chainB.PendingSendPacketsV2))
```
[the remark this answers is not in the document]
aumetra: There is. Check the second part of the log statement

kulikthebird (approved):
LGTM!

PR #2267, opened 2025-05-21
Add `funds` to IBC destination callback if it's a transfer message

chipshort:
Original issue: https://github.com/CosmWasm/cosmwasm/issues/2398
CosmWasm PR: https://github.com/CosmWasm/cosmwasm/pull/2484

Marked as draft because of wasmvm commit dependency.


x/wasm/ibc.go
```diff
@@ -412,9 +416,56 @@ func (i IBCHandler) IBCReceivePacketCallback(
...
+	var transfer *wasmvmtypes.IBCTransferCallback
+
+	var parsedAck channeltypes.Acknowledgement_Result
+	err = json.Unmarshal(ack.Acknowledgement(), &parsedAck)
+
+	// detect successful IBC transfer, meaning:
+	// 1. it was sent to the transfer module
+	// 2. the acknowledgement was successful
+	if packet.GetDestPort() == i.transferKeeper.GetPort(cachedCtx) &&
+		ack.Success() && err == nil && slices.Equal(parsedAck.Result, []byte{1}) {
```
kulikthebird, line 428: Why do we need to additionally parse the ack and look for the `0x1` if we already check for `ack.Success()`? The 0x1 is used as a success result, but in general it is an arbitrary array of data.

x/wasm/ibc.go
```diff
@@ -412,9 +416,56 @@ func (i IBCHandler) IBCReceivePacketCallback(
...
+		}
+
+		// just making sure we have a valid address
+		receiverAddr, err := sdk.AccAddressFromBech32(transferData.Receiver)
+		if err != nil {
+			return err
+		}
+
+		// For a more in-depth explanation of the logic here, see the transfer module implementation:
+		// https://github.com/cosmos/ibc-go/blob/a6217ab02a4d57c52a938eeaff8aeb383e523d12/modules/apps/transfer/keeper/relay.go#L147-L175
```
kulikthebird, line 442: Is there any better source of spec/docs for this implementation? Is it ibc-go dependent logic or is it part of the IBC protocol that is officialy documented?

kulikthebird (commented):
Looks good in general. Let's make sure that the two things I've mentioned are checked before merging it.

x/wasm/ibc.go
chipshort: I just wanted to be extra sure that we don't notify a contract of a failed transfer, but you are right, the `ack.Success()` check should be enough. It's just this weird ibc-go specific thing that isn't part of the IBC spec.

x/wasm/ibc.go
chipshort: The ICS20 spec really doesn't do a good job explaining it, but the sequence diagram in there sort of explains it: https://github.com/cosmos/ibc/blob/main/spec/app/ics-020-fungible-token-transfer/README.md#data-structures

pinosu (approved):
Great work! LGTM 👍

PR #2276, opened 2025-05-28
chore: fix the issue that package is being imported more than once

tongjicoder:
We forked this project and added lint checks in the CI, using  `staticcheck` to identify this issue.

```shell
tests/integration/query_plugin_integration_test.go:33:2: package "github.com/CosmWasm/wasmd/x/wasm/keeper" is being imported more than once (ST1019)
```

### DIFF
--- a/tests/integration/query_plugin_integration_test.go
+++ b/tests/integration/query_plugin_integration_test.go
@@ -30,7 +30,6 @@ import (
 	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
 
 	"github.com/CosmWasm/wasmd/app"
-	"github.com/CosmWasm/wasmd/x/wasm/keeper"
 	wasmKeeper "github.com/CosmWasm/wasmd/x/wasm/keeper"
 	"github.com/CosmWasm/wasmd/x/wasm/keeper/testdata"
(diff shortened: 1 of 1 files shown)


tests/integration/query_plugin_integration_test.go
```diff
@@ -61,7 +60,7 @@ func TestMaskReflectCustomQuery(t *testing.T) {
 	}
 	ownerQueryBz, err := json.Marshal(ownerQuery)
 	require.NoError(t, err)
-	ownerRes, err := keeper.QuerySmart(ctx, contractAddr, ownerQueryBz)
+	ownerRes, err := wasmKeeper.QuerySmart(ctx, contractAddr, ownerQueryBz)
```
chipshort, line 63: Most of these calls were not using the package "keeper", they were using the variable "keeper", so that's why CI is failing now. Could you remove the incorrect replacements again and only replace the ones where the actual package was used (in `TestIBCListChannelsQuery` if I'm not mistaken)

tongjicoder: Oh, sorry for my mistake. I have made the changes; please review it again.

chipshort (approved):
Thanks!

PR #2277, opened 2025-05-28
feat: Register all contracts in IBCv2 port router

kulikthebird:


app/app.go
```diff
@@ -875,6 +875,10 @@ func NewWasmApp(
 		if err := app.WasmKeeper.InitializePinnedCodes(ctx); err != nil {
 			panic(fmt.Sprintf("failed initialize pinned codes %s", err))
 		}
+
+		// TODO: Remove the registration of each contract in https://github.com/CosmWasm/wasmd/issues/2278
+		//       and add IBCv2 port ID prefix before calling `SetRouterV2` during WasmKeeper creation.
+		app.WasmKeeper.RegisterContractsInIbc2Router(ctx)
```
kulikthebird, line 881: I couldn't find a way to test the contracts registration during the app initialization procedure. It seems like the contracts list on the app start up is always empty. As for now I found out that the only way to update the application with an up-to-date state is via `InitGenesis` procedure.

PR #2296, opened 2025-06-18
Fix `TestIBCReflectContract`

chipshort:
Looks like it was an incorrectly resolved conflict in #2258. The reflect contract expects an ordered channel and because of #2291 that's being enforced now.


pinosu (approved):
LGTM 👍

PR #2317, opened 2025-07-07
Add an new endpoint for querying contracts by code id and creator

BigtoC:


BigtoC:
Hello @chipshort! Could you help re-run the `Protobuf Linting` job? It's a rate limit error.

BigtoC:
> hey @BigtoC , can you explain what is the use case for this query? you have query contract by codeid and query contract by creator as 2 separate queries, what is the need for this endpoint?

Hi @pinosu, I want to reduce call cost and time by making one RPC call to handle this logic, instead of making 2 separate calls to the RPC node.

pinosu:
> Hi @pinosu, I want to reduce call cost and time by making one RPC call to handle this logic, instead of making 2 separate calls to the RPC node.
Hi, this seems like a niche use case. The query itself is free (no fees), and the time saved wouldn't be significant, especially considering that such queries are generally not meant to be run frequently.
Why do you run this query so often to consider it relevant?

BigtoC:
> > Hi @pinosu, I want to reduce call cost and time by making one RPC call to handle this logic, instead of making 2 separate calls to the RPC node.
> 
> Hi, this seems like a niche use case. The query itself is free (no fees), and the time saved wouldn't be significant, especially considering that such queries are generally not meant to be run frequently. Why do you run this query so often to consider it relevant?

Hi, I'm working with my team on a web app, which serves as an admin portal to manage all deployed contracts of specific code IDs and creators. I think we can improve the network performance by adding this endpoint.
I understand this is a niche use case. I had another merged PR that solved a problem with a niche use case, but it did beneficial for a product I was working on, so I'm thinking that I can make this PR to help me and maybe others in the future.
Thank you for reviewing 🙏

PR #2324, opened 2025-07-10
Add sync.WaitGroup to ensure all validator votes complete in SubmitAndVoteGovProposal

GarmashAlex:
Previously, validator voting in SubmitAndVoteGovProposal was performed in parallel goroutines without synchronization, which could lead to the test finishing before all votes were cast. This commit introduces a sync.WaitGroup to properly wait for all voting goroutines to complete, ensuring test reliability and correctness.

### DIFF
--- a/tests/system/cli.go
+++ b/tests/system/cli.go
@@ -10,6 +10,7 @@ import (
 	"slices"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


pinosu (approved):
LGTM 👍

PR #2380, opened 2025-10-11
refactor: use reflect.TypeFor

rocksload:
Try to use better api reflect.TypeFor instead of reflect.TypeOf when we have known the type.

More info https://github.com/golang/go/issues/60088


pinosu (approved):
LGTM

PR #2494, opened 2026-05-12
Implement IBC Callbacks Plus for v0.61.x

pinosu:
This brings the IBC Hooks developer experience to ibc-go's callbacks middleware: when the memo carries `dest_callback.calldata` (a field ibc-go already parses but wasmd so far ignored), the destination contract is called through its plain `execute` entrypoint with the transferred funds in `info.funds`(exactly what a Hooks `wasm` memo does today). The contract doesn't need a dedicated IBC entrypoint for this, and ibc-go itself is untouched.

The implementation deliberately mirrors Hooks wherever possible, to minimize the migration effort for contracts already working with Hooks:

- A new middleware (`x/wasm/ibc_callbacks_plus_middleware.go`) between ibc-go callbacks and transfer rewrites the packet `Receiver` to an intermediate address on recv when `calldata` is present, so transfer credits the intermediate and the funds can be re-attached to the contract call afterwards. This is the same trick Hooks plays in `OnRecvPacketOverride`. No-op otherwise.
- The intermediate address derivation is copied verbatim from Hooks (`ibc-wasm-hook-intermediary`, hash of `destChannel/originalSender`), so `info.sender` is identical under Hooks and Callbacks+ and sender-keyed contract state survives a migration from one to the other.
- `IBCReceivePacketCallback` (`x/wasm/ibc.go`) executes `dest_callback.address` with calldata as msg, the intermediate as sender and the transferred amount as funds, like the Hooks `MsgExecuteContract` dispatch. If the execute fails, the callback errors and the callbacks middleware writes an error ack, so the transfer is reverted and refunded on the source chain.

What differs from Hooks:

- Without `calldata`, behavior is unchanged: vanilla `ibc_destination_callback` (ADR-008). The memo field is the only switch, existing contracts are unaffected.
- Source side stays on callbacks semantics: ack/timeout go to `ibc_source_callback`, no `sudo` lifecycle messages. `src_callback.calldata` is rejected at send time (same rule as cosmos/evm's contract keeper).
- Works on IBC v2 too (v2 rewriter variant + callbacks v2 middleware wired in app.go), where Hooks doesn't exist. On v2, `calldata` requires JSON encoding, otherwise failure ack.

## Coexistence with Hooks

Chains that already run Hooks can keep it and add Callbacks+ next to it, so existing Hooks contracts keep working while new ones use callbacks memos. The only problem is a memo carrying both families on the same side; for that there's a new dedup middleware (`x/wasm/ibc_dedup_middleware.go`) to put on top of the stack:

- rejects `wasm` + `dest_callback` on recv, `ibc_callback` + `src_callback` on send
- cross-side combos (e.g. `wasm` + `src_callback`) pass through untouched
- not wired in wasmd's own app since wasmd doesn't run Hooks; a commented line in app.go shows where it goes

## Notes for reviewers

- Bumps ibc-go v10.5.0 → v10.6.0.
- The intermediate is re-derived in `IBCReceivePacketCallback` instead of read from the packet: ibc-go passes the packet by value, so the rewriter's change isn't visible there.
- Memo now parsed in three places (ibc-go callbacks, rewriter, receive callback). Unavoidable without changing ibc-go's `ContractKeeper` interface; the shared helper reuses ibc-go's `GetCallbackData`, so no second schema.
- The intermediate is deterministic and pre-fundable. Not a new risk (same as Hooks), harmless since the exact amount travels via `Funds`.
- A `calldata` memo reaches old ADR-008 contracts through `execute` instead of `ibc_destination_callback` — whether a contract tolerates that is a contract-level concern.

## Testing

- Unit: rewriter (v1 + v2), keeper callbacks (calldata dispatch, derivation, `src_callback.calldata` rejection, ack forwarding), dedup middleware.
- e2e (`tests/e2e/ibc_callbacks_plus_test.go`): successful calldata execute, failing execute with refund, fallback without calldata, send-time rejection, cross-side memo combos.


dianab-cl:
Please add a PR description with the intentions of this PR and expected outcome.

pinosu:
> Please add a PR description with the intentions of this PR and expected outcome.

Done! Let me know if you would like to have more context as we can share a notion doc on slack for example.

PR #2537, opened 2026-08-19
Wasm: compare allowlist entries by decoded address

### LINKED ISSUE #2538: Wasm access allowlists should match Bech32 addresses case-insensitively
## Summary

`AccessConfig.Allowed` performs a case-sensitive string comparison for `AccessTypeAnyOfAddresses` entries:

```go
return slices.Contains(a.Addresses, actor.String())
```

However, `validateBech32Addresses` explicitly treats Bech32 addresses as case-insensitive and accepts all-uppercase representations.

As a result, a valid uppercase Bech32 address can pass `ValidateBasic()` but fail authorization when compared with the lowercase representation returned by `actor.String()`.

For example:

- configured address: `COSMOS1ABC...`
- actor address: `cosmos1abc...`

These addresses represent the same account, but `Allowed()` returns `false`.

## Expected behavior

Equivalent Bech32 representations should authorize the same account.

## Proposed fix

Decode configured addresses with `sdk.AccAddressFromBech32` and compare the decoded address bytes using `AccAddress.Equals`.

A regression test should cover:

- lowercase configured addresses;
- uppercase configured addresses;
- different addresses;
- invalid addresses rejected by `ValidateBasic()`.

zloglevel:
## Summary

- Compare `AccessTypeAnyOfAddresses` entries by decoded `sdk.AccAddress` values instead of raw Bech32 strings.
- Honor valid uppercase Bech32 allowlist entries.
- Add regression coverage for lowercase, uppercase, different, and invalid addresses.
- Add an Unreleased changelog entry.

## Testing

- `go test ./x/wasm/types`
- `go test ./x/wasm/keeper`

Fixes https://github.com/CosmWasm/wasmd/issues/2538

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 [Full Changelog](https://github.com/CosmWasm/wasmd/compare/v0.70.2...HEAD)
 
+- Fix case-insensitive Bech32 address matching in wasm access allowlists
+
 ## [v0.70.2](https://github.com/CosmWasm/wasmd/tree/v0.70.2) (2026-05-22)
 
 [Full Changelog](https://github.com/CosmWasm/wasmd/compare/v0.70.1...v0.70.2)
--- a/x/wasm/types/params.go
+++ b/x/wasm/types/params.go
@@ -142,7 +142,13 @@ func (a AccessConfig) Allowed(actor sdk.AccAddress) bool {
 	case AccessTypeEverybody:
 		return true
 	case AccessTypeAnyOfAddresses:
-		return slices.Contains(a.Addresses, actor.String())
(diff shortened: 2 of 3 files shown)


zloglevel:
@pinosu Could you please take a look when you have a chance? Thanks!

zloglevel:
@pinosu Hi — a quick follow-up on this one. The change compares allowlist entries using decoded sdk.AccAddress values so valid uppercase Bech32 addresses are handled correctly, with regression cases for lowercase, uppercase, different, and invalid addresses.

The change is limited to x/wasm/types plus tests. Would you mind taking a look when you have bandwidth? Thanks!

PR #2540, opened 2026-08-24
docs: align proposal simulator comments with function names

wangjingshuiku:
Update exported proposal simulator comments to begin with the function names they document and clarify the corresponding proposal operation.

No simulation behavior is changed and `go test ./x/wasm/simulation` passes.


wangjingshuiku:
@pinosu Could you please take a look when you have a chance? Thanks!

PR #2542, opened 2026-09-01
fix(wasm): close iterator in IterateContractsByCreator

colinaumaty:
IterateContractsByCreator was the only production prefix-store iteration found here that did not close its iterator, while the neighboring IterateContractsByCode does.


colinaumaty:
@pinosu Hi, this is a small resource-management fix in IterateContractsByCreator that ensures the store iterator is closed with defer iter.Close(). Would you mind taking a look when you have a chance? Thanks!

